PR #3, opened 2022-01-06
remove custom wrapping of sjcl.js output

shakyShane:
WIP


jonathanKingston:
LGTM thanks for making this change :)

jonathanKingston:
@shakyShane I think we should close this unless you're intending to fix it up. I don't know if there's a relevant Asana task even.

PR #4, opened 2022-01-06
Implement navigator interface

shakyShane:
Task: https://app.asana.com/0/0/1201630474140682/f
Original Tech Design: https://app.asana.com/0/481882893211075/1201508656742121/f


src/features/navigator-interface.js
```diff
@@ -0,0 +1,20 @@
...
+    try {
+        if (navigator.duckduckgo) return
+        defineProperty(Navigator.prototype, 'duckduckgo', {
+            value: {
+                platform: args.platform.name,
+                async isDuckDuckGo () {
+                    return true
+                }
+            },
+            enumerable: true,
```
jonathanKingston, line 13: I verified this appears to behave the same as Brave where navigator.brave isn't writable but navigator.brave.isBrave is.

package.json
```diff
@@ -22,6 +22,7 @@
     "sjcl": "^1.0.8"
   },
   "devDependencies": {
+    "@types/chrome": "^0.0.174",
```
jonathanKingston, line 25: Can we remove this until the fast follow issue, I can't see this being used.

shakyShane: done

integration-test/config.js
```diff
@@ -0,0 +1,10 @@
+export default {
+    spec_dir: 'integration-test',
```
jonathanKingston, line 2: Is there a need for this to be JS, curious mainly.

integration-test/helpers/harness.js
```diff
@@ -72,5 +84,41 @@ export async function setup (ops = {}) {
...
+        const url = new URL(urlString)
+
+        // Append the flag so that the script knows to wait for incoming args.
+        url.searchParams.append('wait-for-init-args', 'true')
+
+        await page.goto(url.href)
+
+        // wait until contentScopeFeatures.load() has completed
+        await page.evaluate(() => {
+            return window.__content_scope_status === 'loaded'
```
jonathanKingston, line 107: Shouldn't these evaluates be waitForFunction?

jonathanKingston:
Looking good, left some questions.

Where did you get that icon from, is it open source, do we need it?

integration-test/helpers/harness.js
shakyShane: Nope we don't need it, removed

shakyShane:
icon removed

integration-test/config.js
shakyShane: I'd ask the opposite - why keep it as JSON which is incredibly restrictive if the `.js` version is valid for Jasmine - opens the door for logic/comments/eslint etc.

Basically I don't have a great reason, other that, "why not"?

integration-test/helpers/harness.js
shakyShane: yep! I had moved to `evaluate` because you need to if your return value is a promise (which mine was, but isn't any more) - pushing the change :)

shakyShane:
@jonathanKingston those bits are addressed :)

integration-test/config.js
jonathanKingston: > why keep it as JSON which is incredibly restrictive

This is the feature, don't open yourself up to flexibility you don't need yet.

Either way it's fine, I'd be pushing back further for .eslintrc or similar files.

jonathanKingston (approved):

PR #14, opened 2022-01-24
Make protections testable

kdzwinel:
This PR modifies some of the content scope scripts to allow passing custom `window` object. This is to support testing - I want to control the whole "browser" environment so scripts have to use `globalThis`, which can be overwritten, instead of `window`.

https://app.asana.com/0/0/1201702136255075/f


src/cookie.js
```diff
@@ -51,5 +51,5 @@ export class Cookie {
 
 // support node-requires (for test import)
 if (typeof module !== 'undefined' && module.exports) {
-    module.exports = Cookie
```
kdzwinel, line 54: this is to match how this class is exported at the top:

> export class Cookie

makes sure that export is the same in browser vs node

jonathanKingston: Can we just make the test es6 modules compliant so we can remove this instead?

shakyShane:
@kdzwinel 👋🏻 

I very much appreciate the effort to increase testing here :)

I'd like to suggest that we don't alter any method signatures though, at least not for the purpose of testing - for the following 2 reasons.

1 - the content-script features have a very small API surface currently, and I believe we should strive to keep that as small as possible. Choosing to take up a second-position parameter, adding `globalThis`, limits our choices for API changes in the future. An object would be better here, but honestly I'd rather that not contain environmental things.

2 - if we create a codebase where browser globals are accessed locally in individual function signatures, it creates a situation where engineers on new part of the codebase may not be sure whether or not to access things globally, or to plumb them in as you had to do in `src/features/fingerprinting-screen-size.js` for example.

**Proposal:**

Although I'm not usually a fan of global mocks, I believe the scope of this project warrants a singular way for accessing/testing `globalThis` - without changing any defined APIs.

Could we just install JSDOM's globals on an as-needed basis, like this:

https://github.com/duckduckgo/content-scope-scripts/commit/4d5d1952220ee8e24393146d5d72ef6adf889989

As @jonathanKingston has mentioned in the Asana task, if we were to pass `globalThis` around, there would be other places where we'd need to consider global access too, which this approach would solve too.

@kdzwinel what do you think?

shakyShane:
Also worth noting that I took some time to ensure features can be tested within an extension - even with the ability to provide config within the test, please see https://github.com/duckduckgo/content-scope-scripts/blob/main/integration-test/test-navigator-interface.js#L24

src/cookie.js
kdzwinel: Yeah, good point. I ended up removing and… everything still seems to work? Both here and in extension tests pass.

src/features/tracking-cookies-1p.js
```diff
@@ -116,6 +118,8 @@ function applyCookieExpiryPolicy () {
 
 // Set up 1st party cookie blocker
 export function load (args) {
+    trackerHosts.clear()
```
kdzwinel, line 121: this is not related to globals, but related to testing. This file has trackerHosts variable which was retaining state between tests, there was no way to clear it.

kdzwinel:
Thanks both for comments, I updated the PR as discussed. PTAL.

sammacbeth (dismissed):
This looks good to me. I think you should commit the updated built files though too.

sammacbeth (approved):
👍

PR #17, opened 2022-01-27
Replace FLOC with browsingTopics

SlayterDev:
https://app.asana.com/0/1163321984198618/1201719300245228/f

`browsingTopics` is effectively replacing `floc`. This PR turns our `floc` protection into a `browsingTopics` protection.

New to this build process. Hopefully this looks right.


src/features/google-rejected.js
```diff
@@ -0,0 +1,25 @@
+export function init () {
+    if ('browsingTopics' in Document.prototype) {
```
jonathanKingston, line 2: If we make the if statement inside the try, we don't need to do 3 of them aka:

```
try {
    if ('browsingTopics' in Document.prototype) {
        ....
    }
    if (...) {
        ...
    }
} catch {
    // Throw away this exception, it's likely a confict with another extension
}
```

src/features/google-rejected.js
```diff
@@ -0,0 +1,25 @@
...
+        try {
+            delete Document.prototype.browsingTopics
+        } catch {
+            // Throw away this exception, it's likely a confict with another extension
+        }
+    }
+
+    if ('joinAdInterestGroup' in Navigator.prototype) {
+        try {
+            delete Navigator.prototype.joinAdInterestGroup
```
jonathanKingston, line 12: I think there's some more here https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/ad_auction/navigator_auction.idl:

```
  [CallWith=ScriptState, Measure, RaisesException]
  void joinAdInterestGroup(AuctionAdInterestGroup group, double durationSeconds);

  [CallWith=ScriptState, Measure, RaisesException]
  void leaveAdInterestGroup(optional AuctionAdInterestGroup group);

  [CallWith=ScriptState, Measure, RaisesException]
  void updateAdInterestGroups();

  [RuntimeEnabled=Fledge, CallWith=ScriptState, Measure, RaisesException]
  Promise<USVString?> runAdAuction(AuctionAdConfig config);

  [RuntimeEnabled=Fledge, CallWith=ScriptState, Measure, RaisesException]
  sequence<USVString> adAuctionComponents([Clamp] unsigned short numComponents);

```

jonathanKingston (changes requested):
I think we're missing a few APIs here.

jonathanKingston (approved):

Pitpul1 (commented):
Wel whel

PR #37, opened 2022-03-01
Making a test package

jonathanKingston:
One potential direction to reducing dependency hell.

- Moving CSS from BSK to it's own dependency that holds a weak dependency on BSK.
- Moving UserScripts from BSK to be owned here.


jonathanKingston:
Also resolved in https://github.com/duckduckgo/content-scope-scripts/pull/65

PR #64, opened 2022-05-06
Moving to use Swift module

jonathanKingston:
https://app.asana.com/0/0/1202234437236404/f


swift-package/Resources/ContentScopeScripts.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+public struct ContentScopeScripts {
+    public static var bundle: Bundle = .module
```
miasma13, line 4: I would suggest adding a helper here for getting the content scope script's URL within its bundle. This would contain the details of how to access the resource (e.g. subfolders and filenames) within the package.

miasma13 (changes requested):

jonathanKingston:
This was resolved in https://github.com/duckduckgo/content-scope-scripts/pull/65

PR #65, opened 2022-05-06
Remote value support

jonathanKingston:


src/utils.js
```diff
@@ -242,9 +242,22 @@ function camelcase (dashCaseText) {
...
  */
-export function getFeatureSetting (featureName, args, prop) {
+export function getFeatureSetting (featureName, args, prop, defaultValue) {
     const camelFeatureName = camelcase(featureName)
-    return args.featureSettings?.[camelFeatureName]?.[prop]
+    const configSetting = args.featureSettings?.[camelFeatureName]?.[prop]
+
+    // Config may pass the string 'undefined' which should be interpreted as undefined
+    // Otherwise return the config value
+    if (configSetting !== undefined) {
```
jonathanKingston, line 251: Can we place this logic change in a function that wraps _getFeatureSetting_ aka _getAttributeValue_ and then express the setting value as an object in the config?

The reason being is you may in future wish to express other types not expressible in JSON like ArrayBuffer or NaN, Symbol or something like that.

In the config this could look something like:

```
"settings": {
    "keyboard": {type: 'undefined'},
    "availLeft": {type: 'number', value: 12},
 }
 ```
 

I don't think the rest of the code will want to support the 'undefined' string value.

File: Package.swift [hunk 2]
```diff
@@ -0,0 +1,27 @@
+// swift-tools-version:5.3
```
jonathanKingston, line 1: Have you tested all this in a MacOS build?

src/utils.js
SlayterDev: Fair enough 👍

Package.swift
SlayterDev: I have indeed. It does work. But I'll verify this value is correct as well.

SlayterDev: This swift tools version matches BSK. Not sure if you were commenting about the value in particular or the swift package as a whole but both are functional.

jonathanKingston: No just I meant the whole concept, I did some testing just wasn't sure if it made sense to couple both things together. :)

src/utils.js
```diff
@@ -236,15 +236,50 @@ function camelcase (dashCaseText) {
...
  * @param {object} args
  * @param {string} prop
  * @returns {any}
  */
-export function getFeatureSetting (featureName, args, prop) {
+export function getFeatureSetting (featureName, args, prop, defaultValue) {
     const camelFeatureName = camelcase(featureName)
-    return args.featureSettings?.[camelFeatureName]?.[prop]
+    const configSetting = args.featureSettings?.[camelFeatureName]?.[prop]
+    return getFeatureAttr(configSetting, defaultValue)
```
jonathanKingston, line 282: This is wrapped the wrong way around to what I meant (sorry if this wasn't clear).

_getFeatureAttr_ should call into _getFeatureSetting_ which is already in use, the return value of the _getFeatureSetting_ function should just return the raw config setting value and shouldn't be modified in this change.

src/features/fingerprinting-screen-size.js and src/features/fingerprinting-hardware.js should call into _getFeatureAttr_  instead of _getFeatureSetting_ 

_getFeatureAttr_  should handle unwrapping the config value after calling _getFeatureSetting_ 

Additionally  _getFeatureAttr_  probably should handle the AppleSilicon forking logic rather than having a special case for concurrency specifically. for example hardwareConcurrency currently doesn't support _{type: 'undefined'}_ and I think this could be handled by expressing this in the config return value.

For example the config could have:
```
"settings": {
    "keyboard": {type: 'undefined'},
    "availLeft": {type: 'number', value: 12},
    "hardwareConcurrency": [
        {type: 'number', value: 12},
        {type: 'number', value: 12, criteria: {arch: 'AppleSilicon'}},
     ]
 }
 ```
This should allow for different criteria later and also allow us to change for example 'keyboard' as a remote config update.

Happy to discuss over MM or Asana if that's more helpful than GH comments :)

File: src/utils.js [hunk 4]
```diff
@@ -236,6 +236,91 @@ function camelcase (dashCaseText) {
...
+            if (item.criteria.arch === 'AppleSilicon' && isAppleSilicon()) {
+                bestOption = item
+                break
+            }
+        } else {
+            bestOption = item
+        }
+    }
+
+    if (bestOption && bestOption.type) {
```
jonathanKingston, line 273: I think this can be restructured to return just the best option and _getFeatureAttr_ handle the undefined and value code as it already does.

SlayterDev: Yeah not very DRY. Fixed.

src/utils.js
```diff
@@ -236,6 +236,84 @@ function camelcase (dashCaseText) {
...
+/**
+ * Take configSeting which should be an array of possible values.
+ * If a value contains a criteria that is a match for this environment then return that value.
+ * Otherwise return the first value that doesn't have a criteria.
+ *
+ * @param {array} configSetting - Config setting which should contain a list of possible values
+ * @returns - The value from the list that best matches the criteria in the config
+ */
+function processAttrByCriteria (configSetting) {
+    if (!configSetting.some(item => item.criteria !== undefined)) {
```
jonathanKingston, line 258: I think this will give out some unexpected behaviour.

You've verified that the argument to this method is an array in the _getFeatureAttr_ callsite, so this code if the items didn't have a _criteria_  property will return _[{...}, {...}]_

I'd delete this statement.

src/utils.js
```diff
@@ -236,6 +236,84 @@ function camelcase (dashCaseText) {
...
+    return gl.getSupportedExtensions().indexOf('WEBGL_compressed_texture_etc') !== -1
+}
+
+/**
+ * Take configSeting which should be an array of possible values.
+ * If a value contains a criteria that is a match for this environment then return that value.
+ * Otherwise return the first value that doesn't have a criteria.
+ *
+ * @param {array} configSetting - Config setting which should contain a list of possible values
+ * @returns - The value from the list that best matches the criteria in the config
```
jonathanKingston, line 255:
```suggestion
 * @param {*[]} configSetting list of possible config options
 * @returns {*|undefined} The value from the list that best matches the criteria in the config
```

Optionally you could setup a type for this something like:
```
@typedef ConfigOption
@property {Record<string, *>} critera
@property {string} type
@property {*} value
```

And then instead:
```suggestion
 * @param {ConfigOption[]} configSetting list of possible config options
 * @returns {ConfigOption|undefined} The value from the list that best matches the criteria in the config
```

(See other comments, I think this function should return _undefined_ when the match isn't found as an error state)

Another note is that tsconfig.json should include this file as we're adding some typings (I checked locally and nothing goes wrong by adding it so it's worth including so that automation catches issues when someone edits the file)
 - FYI: We're slowly rolling out typing, so it's an opt in for extension and content-scope-scripts currently.

src/utils.js
```diff
@@ -236,6 +236,84 @@ function camelcase (dashCaseText) {
...
+
+    if (configSetting === undefined) {
+        return defaultValue
+    }
+
+    const configSettingType = typeof configSetting
+    switch (configSettingType) {
+    case 'object':
+        if (Array.isArray(configSetting)) {
+            configSetting = processAttrByCriteria(configSetting)
```
jonathanKingston, line 300: This should cater for returning undefined and return _defaultValue_ if it is.

src/utils.js
```diff
@@ -236,6 +236,84 @@ function camelcase (dashCaseText) {
...
+        }
+
+        if (!configSetting.type) {
+            return configSetting
```
jonathanKingston, line 304: Should this be defaultValue, this seems like an error state right or are we expecting to support things that don't match the structure we penned out in Asana/this PR?

src/utils.js
```diff
@@ -236,6 +236,84 @@ function camelcase (dashCaseText) {
...
+        }
+
+        if (configSetting.type === 'undefined') {
+            return undefined
+        }
+
+        return configSetting.value
+    default:
+        return configSetting
```
jonathanKingston, line 313: Should this be _defaultValue_?

src/utils.js
```diff
@@ -236,6 +236,84 @@ function camelcase (dashCaseText) {
...
+ *
+ * @param {array} configSetting - Config setting which should contain a list of possible values
+ * @returns - The value from the list that best matches the criteria in the config
+ */
+function processAttrByCriteria (configSetting) {
+    if (!configSetting.some(item => item.criteria !== undefined)) {
+        return configSetting
+    }
+
+    let bestOption = configSetting[0]
```
jonathanKingston, line 262: This will fail if we have an erroneous _configSetting_ of an empty array. I think we should declare this as _undefined_ and the existing code will return it.
```suggestion
    let bestOption
```

In the calling code I think we should handle the _undefined_ case and return _defaultValue_

src/utils.js
SlayterDev: I think I was trying to add insurance here for other formats. But upon further thought this is the only thing using this format so we'll be able to control new settings through the config. `defaultValue` seems right.

src/utils.js
SlayterDev: Same thinking as above. Will change to `defaultValue`

jonathanKingston:
Squash merged as LGTM, I'm not able to r+ my own PR (we should avoid doing this again)

Thanks

PR #68, opened 2022-05-12
fix error cookie override when `path=/`

tachibana-shin:


sammacbeth:
Thanks for this patch!

Have you tested this change with the site you saw the original issue on (https://github.com/duckduckgo/duckduckgo-privacy-extension/issues/1161)? If so, could you also share the `stack` value that caused the original error? Think might help us create a test-case for this.

jonathanKingston:
So I included this change into https://github.com/duckduckgo/content-scope-scripts/pull/78/ thank you for bringing this to our attention!

PR #72, opened 2022-05-17
Add nonTracking3pCookies script

SlayterDev:
https://app.asana.com/0/72649045549333/1202156648532924/f

Adds blocking for non-tracking 3rd party cookies.


src/features/non-tracking-3p-cookies.js
```diff
@@ -0,0 +1,39 @@
...
+    })
+}
+
+export function init (args) {
+    args.cookie.debug = args.debug
+    if (globalThis.top !== globalThis && !args.cookie.isTrackerFrame && args.cookie.shouldBlock && args.cookie.isThirdParty) {
+        // overrides expiry policy with blocking - only in subframes
+        blockCookies(args.debug)
+    }
+}
```
SlayterDev, line 39: This feels weird to me as the script is exactly the same as our `tracking-cookies-3p` script with the exception of the `isTrackingFrame` check. Although with the architecture of the content-scope-scripts I'm not sure if there is a more DRY approach.

sammacbeth (approved):

PR #78, opened 2022-05-26
Simplify cookie logic

jonathanKingston:


src/features/cookie.js
```diff
@@ -0,0 +1,215 @@
...
+        const scriptOrigins = getStackTraceOrigins(stack)
+        const getCookieContext = {
+            stack,
+            scriptOrigins,
+            value: 'getter'
+        }
+
+        if (shouldBlockTrackingCookie() || shouldBlockNonTrackingCookie()) {
+            if (cookiePolicy.debug) {
+                debugHelper('block', 'frame', getCookieContext)
```
SlayterDev, line 118: Should this say 3p frame as well like on line 135?

jonathanKingston: Cleared this up anyway also add an ignore case :)

SlayterDev (approved):
🚀

PR #84, opened 2022-06-01
Update hostname matching to match background behavior

SlayterDev:
This PR matches the behavior in the background page for partyness checks. Adds tldts as a dependency.


src/utils.js
```diff
@@ -101,14 +103,49 @@ export function getTabOrigin () {
     return framingOrigin
 }
 
+function findWebsiteOwner (requestData) {
+    // find the site owner
+    const siteUrlList = Array.from(requestData.siteUrlSplit)
```
jonathanKingston, line 108: Can we use the existing function rather than the while loop?

src/utils.js
```diff
@@ -101,14 +103,49 @@ export function getTabOrigin () {
...
+            return entities[siteToCheck]
+        }
+    }
+}
+
+/**
+ * Tests whether the two URL's belong to the same
+ * top level domain.
+ */
+function isSameTopLevelDomain (url1, url2) {
```
jonathanKingston, line 124: We should probably file a follow up to share as much of this code with the extension.

shared/entities.js
```diff
@@ -0,0 +1,5758 @@
+
+/* eslint-disable quote-props */
```
jonathanKingston, line 2: Let's analyse the perf cost to injecting this into the extension (along with the tldts)? This is why is why it was avoided before.

PR #109, opened 2022-07-06
Add windows browser specific content scripts

q71114:
https://app.asana.com/0/72649045549333/1201631403895850/f

Related PR in config: https://github.com/duckduckgo/privacy-configuration/pull/305


build/windows/contentScope.js
```diff
@@ -0,0 +1,4424 @@
...
+        const Status = {
+            Inactive: 'inactive',
+            Accessed: 'accessed',
+            Active: 'active',
+            Paused: 'paused'
+        };
+
+        const isFrameInsideFrame = window.self !== window.top && window.parent !== window.top;
+
+        function windowsPostMessage (name, data) {
```
szanto90balazs, line 852: I believe this should also be using the secure messaging methods: https://github.com/more-duckduckgo-org/windows-browser/blob/develop/WindowsBrowser/Scripts/SecureContentScriptSandboxTemplate.js

szanto90balazs: I see what you're saying, still a little bit misleading, but I can't think of a better way than having those things at the same location, which seems like a no-go.

scripts/generateSJCL.js
```diff
@@ -10,7 +10,11 @@ async function init () {
         return
     }
 
-    await exec('cd node_modules/sjcl/ && ./configure --no-export --compress=none --without-all --with-hmac --with-codecHex && make')
```
jonathanKingston, line 13: I'm going to guess the perl side works for linux/mac too maybe they can be completely shared.

src/apple-utils.js
```diff
@@ -1,32 +1,4 @@
...
-        unprotectedDomain = featureList.filter(domain => domain.domain === partialDomain).length > 0
-
-        domainParts.shift()
-    }
-
-    return unprotectedDomain
-}
+import { getTopLevelURL, isUnprotectedDomain } from './utils'
 
 export function processConfig (data, userList, preferences) {
```
jonathanKingston, line 3: Thanks for sharing the common methods above :)

Does it make sense to share this with windows also; it seems like there's very little difference. Or break it into even smaller functions perhaps?

src/features/windows-permission-usage.js
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    const Status = {
+        Inactive: 'inactive',
+        Accessed: 'accessed',
+        Active: 'active',
+        Paused: 'paused'
+    }
+
+    const isFrameInsideFrame = window.self !== window.top && window.parent !== window.top
```
jonathanKingston, line 20: I think this can be swapped with: https://github.com/duckduckgo/content-scope-scripts/blob/50df895796a246a0cceb2e352b25e299d4361b19/src/utils.js#L59-L65

q71114: If my understanding is correct `isBeingFramed` return `true` when the caller is an iframe, but in the permission script we need to determine if it's an iframe inside an iframe. Can we do that with the existing methods?

src/apple-utils.js
q71114: Yeah, you're right I moved that one to the common utils as well.

src/features/windows-permission-usage.js
jonathanKingston: Oh sorry you're right!
ancestorOrigins will give a better response to detect this though as it will work cross origin. I'm happy to ignore for now though :)

PR #123, opened 2022-08-15
Windows native-provided Geolocation 

RendijsSmukulis:
Replaces the original `watchPosition` and `getCurrentPosition`  geolocation methods with ones that interact with the native code through messaging.


build/apple/contentScope.js
```diff
@@ -3994,22 +4038,19 @@
...
 
-              const successHandler = args[0];
-              args[0] = function (position) {
-                  signalPermissionStatus(Permission.Geolocation, Status.Active);
-                  successHandler?.(position);
-              };
-              const id = DDGReflect.apply(target, thisArg, args);
-              watchedPositions.add(id);
-              return id
+              const messageId = Math.random().toString();
```
q71114, line 4041: Use a uuid instead of a random number: `const messageId =crypto.randomUUID()`

q71114 (dismissed):
LGTM apart from the random number as id.

build/apple/contentScope.js
RendijsSmukulis: Nice, thank you! Was looking for an existing alternative, didn't find one.

q71114 (approved):

PR #131, opened 2022-08-19
Add license to package.json

jonathanKingston:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@duckduckgo/content-scope-scripts",
-  "version": "1.1.2",
```
kdzwinel, line 3: We do versioned releases on this repo - https://github.com/duckduckgo/content-scope-scripts/releases , is this `version` field not useful at all?

jonathanKingston: It's useful if we automate it and keep it up to date.
But also I think it only becomes useful if we publish to npm; we can add back in as/when/if we do that.

kdzwinel (approved):
LGTM

PR #137, opened 2022-09-05
Add Chrome MV3 build target

kzar:
Add a Chrome MV3 build target, that produces a main-world content
script ready for injection. Since there is not yet a way to pass
arguments to chrome.scripting.registerContentScript[1] registered
content scripts, generate a secret in the main-world (before the
website's own scripts have a chance to run) and emit that as an event
for our content script. That way, the main-world script still has a
way to share a secret with the content script for messaging.

Note: The registerContentScript does not yet allow about:blank frames
      to be targeted[2].

1 - https://developer.chrome.com/docs/extensions/reference/scripting/#method-registerContentScripts
2 - https://crbug.com/1360392


inject/chrome-mv3.js
```diff
@@ -0,0 +1,36 @@
...
+    case 'register':
+        // Background has disabled features
+        if (!message?.argumentsObject) {
+            return
+        }
+
+        if (message?.argumentsObject.debug) {
+            window.addEventListener('message', (m) => {
+                if (m.data.action && m.data.message) {
+                    chrome.runtime.sendMessage({
```
jonathanKingston, line 21: This won't work will it?
We're in the page world now so we'd have to pass this back to the isolated world first.

inject/chrome-mv3.js
```diff
@@ -0,0 +1,36 @@
+/* global contentScopeFeatures */
+
+const secret = window.crypto.randomUUID()
+
+contentScopeFeatures.load()
+
+window.addEventListener(secret, ({ detail: message }) => {
+    switch (message?.type) {
```
jonathanKingston, line 8: Nit: Rather than all the ? checking for message an early return would make the code simpler.
```suggestion
    if (message) {
        return
    }
    switch (message.type) {
```

Also typescript is a little dumb sometimes I'm guessing `contentScopeFeatures.init(message?.argumentsObject)` was added because TS was complaining; but you're already returning early in that switch 😢

inject/chrome-mv3.js
kzar: Doh, good point. Done.

inject/chrome-mv3.js
kzar: Done.

jonathanKingston (approved):

PR #146, opened 2022-09-09
Add Android integration

joshliebe:
https://app.asana.com/0/0/1202939051866215/f

### DIFF
--- a/scripts/inject.mjs
+++ b/scripts/inject.mjs
@@ -50,6 +50,8 @@ async function init() {
         initOther('inject/mozilla.js', process.argv[2]);
     } else if (process.argv[2] == "apple") {
         initOther('inject/apple.js', process.argv[2]);
+    } else if (process.argv[2] == "android") {
+        initOther('inject/android.js', process.argv[2]);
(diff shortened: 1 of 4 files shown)


scripts/inject.mjs
```diff
@@ -50,6 +50,8 @@ async function init() {
         initOther('inject/mozilla.js', process.argv[2]);
     } else if (process.argv[2] == "apple") {
         initOther('inject/apple.js', process.argv[2]);
+    } else if (process.argv[2] == "android") {
+        initOther('inject/android.js', process.argv[2]);
```
jonathanKingston, line 54: You'll need another change in https://github.com/duckduckgo/content-scope-scripts/blob/main/package.json#L9 to output this file and then check in the resulting built file also :)

joshliebe: Yeah, I just realized this. Will add 👍

joshliebe:
@jonathanKingston I've updated the build script and added the built JS file. Please re-check!

jonathanKingston (approved):
LGTM thanks!

PR #148, opened 2022-09-15
Click to Load support

ladamski:
Add CTL content script (vs embedding it in each product)

https://app.asana.com/0/481882893211075/1201720254973502/f


ladamski:
@jonathanKingston would love your initial take on this PR.  AFAICT CTL script doesn't use defineProperty, proxies or reflection but not sure what else I might be missing.  A few things to note:
- unit tests pass on my local but for some reason I'm getting an error about something needing to be updated in the PR?
- pretty sure sendMessage needs a proxy to work across platforms, so that's pending

jonathanKingston:
Hey @ladamski,

Travelling at the moment but my initial view is this looks great.

I agree with making sendMessage cross platform we could handle that in a utility probably like we have for internal differences for DDGProxy.

Let's try and make as minimal changes to this PR as possible prior to landing to what we currently have; we can always fast follow with nits etc.

The framework should do a good enough job to prevent scope leaking so there's no real need to indent all the code into init etc.

Weird about the unit test perhaps there's a code output difference based on npm or node versions?

jonathanKingston:
The chrome.runtime.getURL should probably abstracted into a function we can change for each platform as well.

I'm wondering if it makes sense to have a way to disable platform specific features for a build (outside of the config);  that way features like this wouldn't appear in the bundle until the native integration is complete. I'll raise a task in Asana so we don't forget; it doesn't need to block here.

ladamski:
Appreciate the quick feedback, thank you!   No rush on the full review. 

I can take a shot at proxying sendMessage and getURL, and good point on the per-platform feature settings.  For the settings though, do we want to manage those settings within this repo or as part of https://github.com/duckduckgo/privacy-configuration/tree/main/features?

Also agree on getting this landed sooner than later, happy to do follow-ups to clean it up.

ladamski:
Actually, looking at getURL the trick is that on Mac (and I'm guessing iOS) I couldn't find an equivalent to load resources dynamically from a content script, instead I ended up passing those fonts in as mime-encoded params: https://github.com/more-duckduckgo-org/macos-browser/blob/develop/DuckDuckGo/Content%20Blocker/clickToLoad.js#L16

So perhaps the approach for getURL is to avoid it entirely?

ladamski:
Additional wrinkle: for Mac, sendMessage isn't supported so you can to use this funky postMessage syntax: https://github.com/more-duckduckgo-org/macos-browser/blob/develop/DuckDuckGo/Content%20Blocker/clickToLoad.js#L731

Not sure how to make a generic proxy for this though, as the calling syntax is `window.webkit.messageHandlers.<msgID>.postMessage(<msgArgs>)`.  Could do a static mapping I guess but that might as well live in the CTL script itself?

ladamski:
@jonathanKingston  I updated the PR with the latest CTL script, also now includes UI strings (removed from the config).  These will later be localized but for now this should work.  I did a smoke test it in the extension and this script worked.

Given the immediate focus is on getting this landed in MV3, I think we should defer the conversations around getURL and sendMessage proxies until we cross that bridge for Mac (or Win)?

ladamski:
Alright I fixed the tests, not sure where this was introduced but the apple dist directory seemed broken (https://github.com/duckduckgo/content-scope-scripts/pull/148/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L12) so hopefully this is good to go now.

jonathanKingston:
> Alright I fixed the tests, not sure where this was introduced but the apple dist directory seemed broken (https://github.com/duckduckgo/content-scope-scripts/pull/148/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L12) so hopefully this is good to go now.

The former directory is correct it's a weird path name as that's the expectation of a Swift module.

ladamski:
> The former directory is correct it's a weird path name as that's the expectation of a Swift module.

Ok so I can revert that but then tests fail at:

```
> lint-no-output-globals
> eslint build/apple/contentScope.js --rule '{"no-implicit-globals": ["error"]}' --no-ignore --no-eslintrc --parser-options=ecmaVersion:11


Oops! Something went wrong! :(

ESLint: 7.32.0

No files matching the pattern "build/apple/contentScope.js" were found.
Please check for typing mistakes in the pattern.
```

This step explicitly looks for that path?

jonathanKingston:
> This step explicitly looks for that path?

Ah yeah that should change.
I wonder if that test isn't running in ci perhaps; or we just found the output and deleted it perhaps?

Either way do you mind changing that in package.json?

ladamski:
@jonathanKingston ok reverting the last change and updated the eslint path for Apple instead, seems to work now!

jonathanKingston:
Shame GH doesn't allow diffing across repos via the web interface!

Comparing to the extension:
```diff
< (function clickToLoad () {
---
> export function init (args) {
363a364,730
>      *  Strings (temporarily here, until localized)
>      *********************************************************/
>     const strings = {
>         Facebook: {
>             informationalModal: {
>                 icon: 'blocked_facebook_logo.svg',
>                 messageTitle: 'Logging in with Facebook lets them track you',
>                 messageBody: "Once you're logged in, DuckDuckGo can't block Facebook content from tracking you on this site.",
>                 confirmButtonText: 'Log In',
>                 rejectButtonText: 'Go back'
>             },
>             elementData: {
>                 'FB Like Button': {
>                     selectors: [
>                         '.fb-like'
>                     ],
>                     replaceSettings: {
>                         type: 'blank'
>                     }
>                 },
>                 'FB Button iFrames': {
>                     selectors: [
>                         "iframe[src*='://www.facebook.com/plugins/like.php']",
>                         "iframe[src*='://www.facebook.com/v2.0/plugins/like.php']",
>                         "iframe[src*='://www.facebook.com/plugins/share_button.php']",
>                         "iframe[src*='://www.facebook.com/v2.0/plugins/share_button.php']"
>                     ],
>                     replaceSettings: {
>                         type: 'blank'
>                     }
>                 },
>                 'FB Save Button': {
>                     selectors: [
>                         '.fb-save'
>                     ],
>                     replaceSettings: {
>                         type: 'blank'
>                     }
>                 },
>                 'FB Share Button': {
>                     selectors: [
>                         '.fb-share-button'
>                     ],
>                     replaceSettings: {
>                         type: 'blank'
>                     }
>                 },
>                 'FB Page iFrames': {
>                     selectors: [
>                         "iframe[src*='://www.facebook.com/plugins/page.php']",
>                         "iframe[src*='://www.facebook.com/v2.0/plugins/page.php']"
>                     ],
>                     replaceSettings: {
>                         type: 'dialog',
>                         buttonText: 'Unblock Content',
>                         infoTitle: 'DuckDuckGo blocked this content to prevent Facebook from tracking you',
>                         infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                         simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                     },
>                     clickAction: {
>                         type: 'originalElement'
>                     }
>                 },
>                 'FB Page Div': {
>                     selectors: [
>                         '.fb-page'
>                     ],
>                     replaceSettings: {
>                         type: 'dialog',
>                         buttonText: 'Unblock Content',
>                         infoTitle: 'DuckDuckGo blocked this content to prevent Facebook from tracking you',
>                         infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                         simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                     },
>                     clickAction: {
>                         type: 'iFrame',
>                         targetURL: 'https://www.facebook.com/plugins/page.php?href=data-href&tabs=data-tabs&width=data-width&height=data-height',
>                         urlDataAttributesToPreserve: {
>                             'data-href': {
>                                 default: '',
>                                 required: true
>                             },
>                             'data-tabs': {
>                                 default: 'timeline'
>                             },
>                             'data-height': {
>                                 default: '500'
>                             },
>                             'data-width': {
>                                 default: '500'
>                             }
>                         },
>                         styleDataAttributes: {
>                             width: {
>                                 name: 'data-width',
>                                 unit: 'px'
>                             },
>                             height: {
>                                 name: 'data-height',
>                                 unit: 'px'
>                             }
>                         }
>                     }
>                 },
>                 'FB Comment iFrames': {
>                     selectors: [
>                         "iframe[src*='://www.facebook.com/plugins/comment_embed.php']",
>                         "iframe[src*='://www.facebook.com/v2.0/plugins/comment_embed.php']"
>                     ],
>                     replaceSettings: {
>                         type: 'dialog',
>                         buttonText: 'Unblock Comment',
>                         infoTitle: 'DuckDuckGo blocked this comment to prevent Facebook from tracking you',
>                         infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                         simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                     },
>                     clickAction: {
>                         type: 'originalElement'
>                     }
>                 },
>                 'FB Comments': {
>                     selectors: [
>                         '.fb-comments',
>                         'fb\\:comments'
>                     ],
>                     replaceSettings: {
>                         type: 'dialog',
>                         buttonText: 'Unblock Comments',
>                         infoTitle: 'DuckDuckGo blocked these comments to prevent Facebook from tracking you',
>                         infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                         simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                     },
>                     clickAction: {
>                         type: 'allowFull',
>                         targetURL: 'https://www.facebook.com/v9.0/plugins/comments.php?href=data-href&numposts=data-numposts&sdk=joey&version=v9.0&width=data-width',
>                         urlDataAttributesToPreserve: {
>                             'data-href': {
>                                 default: '',
>                                 required: true
>                             },
>                             'data-numposts': {
>                                 default: 10
>                             },
>                             'data-width': {
>                                 default: '500'
>                             }
>                         }
>                     }
>                 },
>                 'FB Embedded Comment Div': {
>                     selectors: [
>                         '.fb-comment-embed'
>                     ],
>                     replaceSettings: {
>                         type: 'dialog',
>                         buttonText: 'Unblock Comment',
>                         infoTitle: 'DuckDuckGo blocked this comment to prevent Facebook from tracking you',
>                         infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                         simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                     },
>                     clickAction: {
>                         type: 'iFrame',
>                         targetURL: 'https://www.facebook.com/v9.0/plugins/comment_embed.php?href=data-href&sdk=joey&width=data-width&include_parent=data-include-parent',
>                         urlDataAttributesToPreserve: {
>                             'data-href': {
>                                 default: '',
>                                 required: true
>                             },
>                             'data-width': {
>                                 default: '500'
>                             },
>                             'data-include-parent': {
>                                 default: 'false'
>                             }
>                         },
>                         styleDataAttributes: {
>                             width: {
>                                 name: 'data-width',
>                                 unit: 'px'
>                             }
>                         }
>                     }
>                 },
>                 'FB Post iFrames': {
>                     selectors: [
>                         "iframe[src*='://www.facebook.com/plugins/post.php']",
>                         "iframe[src*='://www.facebook.com/v2.0/plugins/post.php']"
>                     ],
>                     replaceSettings: {
>                         type: 'dialog',
>                         buttonText: 'Unblock Post',
>                         infoTitle: 'DuckDuckGo blocked this post to prevent Facebook from tracking you',
>                         infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                         simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                     },
>                     clickAction: {
>                         type: 'originalElement'
>                     }
>                 },
>                 'FB Posts Div': {
>                     selectors: [
>                         '.fb-post'
>                     ],
>                     replaceSettings: {
>                         type: 'dialog',
>                         buttonText: 'Unblock Post',
>                         infoTitle: 'DuckDuckGo blocked this post to prevent Facebook from tracking you',
>                         infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                         simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                     },
>                     clickAction: {
>                         type: 'allowFull',
>                         targetURL: 'https://www.facebook.com/v9.0/plugins/post.php?href=data-href&sdk=joey&show_text=true&width=data-width',
>                         urlDataAttributesToPreserve: {
>                             'data-href': {
>                                 default: '',
>                                 required: true
>                             },
>                             'data-width': {
>                                 default: '500'
>                             }
>                         },
>                         styleDataAttributes: {
>                             width: {
>                                 name: 'data-width',
>                                 unit: 'px'
>                             },
>                             height: {
>                                 name: 'data-height',
>                                 unit: 'px',
>                                 fallbackAttribute: 'data-width'
>                             }
>                         }
>                     }
>                 },
>                 'FB Video iFrames': {
>                     selectors: [
>                         "iframe[src*='://www.facebook.com/plugins/video.php']",
>                         "iframe[src*='://www.facebook.com/v2.0/plugins/video.php']"
>                     ],
>                     replaceSettings: {
>                         type: 'dialog',
>                         buttonText: 'Unblock Video',
>                         infoTitle: 'DuckDuckGo blocked this video to prevent Facebook from tracking you',
>                         infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                         simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                     },
>                     clickAction: {
>                         type: 'originalElement'
>                     }
>                 },
>                 'FB Video': {
>                     selectors: [
>                         '.fb-video'
>                     ],
>                     replaceSettings: {
>                         type: 'dialog',
>                         buttonText: 'Unblock Video',
>                         infoTitle: 'DuckDuckGo blocked this video to prevent Facebook from tracking you',
>                         infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                         simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                     },
>                     clickAction: {
>                         type: 'iFrame',
>                         targetURL: 'https://www.facebook.com/plugins/video.php?href=data-href&show_text=true&width=data-width',
>                         urlDataAttributesToPreserve: {
>                             'data-href': {
>                                 default: '',
>                                 required: true
>                             },
>                             'data-width': {
>                                 default: '500'
>                             }
>                         },
>                         styleDataAttributes: {
>                             width: {
>                                 name: 'data-width',
>                                 unit: 'px'
>                             },
>                             height: {
>                                 name: 'data-height',
>                                 unit: 'px',
>                                 fallbackAttribute: 'data-width'
>                             }
>                         }
>                     }
>                 },
>                 'FB Group iFrames': {
>                     selectors: [
>                         "iframe[src*='://www.facebook.com/plugins/group.php']",
>                         "iframe[src*='://www.facebook.com/v2.0/plugins/group.php']"
>                     ],
>                     replaceSettings: {
>                         type: 'dialog',
>                         buttonText: 'Unblock Content',
>                         infoTitle: 'DuckDuckGo blocked this content to prevent Facebook from tracking you',
>                         infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                         simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                     },
>                     clickAction: {
>                         type: 'originalElement'
>                     }
>                 },
>                 'FB Group': {
>                     selectors: [
>                         '.fb-group'
>                     ],
>                     replaceSettings: {
>                         type: 'dialog',
>                         buttonText: 'Unblock Content',
>                         infoTitle: 'DuckDuckGo blocked this content to prevent Facebook from tracking you',
>                         infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                         simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                     },
>                     clickAction: {
>                         type: 'iFrame',
>                         targetURL: 'https://www.facebook.com/plugins/group.php?href=data-href&width=data-width',
>                         urlDataAttributesToPreserve: {
>                             'data-href': {
>                                 default: '',
>                                 required: true
>                             },
>                             'data-width': {
>                                 default: '500'
>                             }
>                         },
>                         styleDataAttributes: {
>                             width: {
>                                 name: 'data-width',
>                                 unit: 'px'
>                             }
>                         }
>                     }
>                 },
>                 'FB Login Button': {
>                     selectors: [
>                         '.fb-login-button'
>                     ],
>                     replaceSettings: {
>                         type: 'loginButton',
>                         icon: 'blocked_facebook_logo.svg',
>                         buttonText: 'Log in with Facebook',
>                         popupTitleText: 'DuckDuckGo blocked this Facebook login',
>                         popupBodyText: 'Facebook tracks your activity on a site when you use them to login.'
>                     },
>                     clickAction: {
>                         type: 'allowFull',
>                         targetURL: 'https://www.facebook.com/v9.0/plugins/login_button.php?app_id=app_id_replace&auto_logout_link=false&button_type=continue_with&sdk=joey&size=large&use_continue_as=false&width=',
>                         urlDataAttributesToPreserve: {
>                             'data-href': {
>                                 default: '',
>                                 required: true
>                             },
>                             'data-width': {
>                                 default: '500'
>                             },
>                             app_id_replace: {
>                                 default: 'null'
>                             }
>                         }
>                     }
>                 }
>             }
>         }
>     }
> 
>     /*********************************************************
693,694c1060,1061
<     async function init (extensionResponseData) {
<         for (const entity of Object.keys(extensionResponseData)) {
---
>     async function init () {
>         for (const entity of Object.keys(strings)) {
696c1063
<             const { informationalModal, simpleVersion } = extensionResponseData[entity]
---
>             const { informationalModal, simpleVersion } = strings[entity]
714c1081
<         await replaceClickToLoadElements(extensionResponseData)
---
>         await replaceClickToLoadElements(strings)
717c1084
<             replaceClickToLoadElements(extensionResponseData, target)
---
>             replaceClickToLoadElements(strings, target)
1229c1596
< })()
---
> }
```

A few things that stand out:

- "strings" also contains code; I would continue to call this something like "config" or similar until the classes etc are separated out.
- We now have nested "init" functions; I think either rename the inner one and/or remove the nesting of being within "init"

> Given the immediate focus is on getting this landed in MV3, I think we should defer the conversations around getURL and sendMessage proxies until we cross that bridge for Mac (or Win)?

Ok but the intent is to keep this working for MV2 right?

ladamski:
> * "strings" also contains code; I would continue to call this something like "config" or similar until the classes etc are separated out.

Good point, fixed.

> * We now have nested "init" functions; I think either rename the inner one and/or remove the nesting of being within "init"

Ditto.

> > Given the immediate focus is on getting this landed in MV3, I think we should defer the conversations around getURL and sendMessage proxies until we cross that bridge for Mac (or Win)?
> 
> Ok but the intent is to keep this working for MV2 right?

Is it though?  If the question is "do we port this to MV2", I think the answer should be "no".  Given the drop dead date at EOY, we'll want to publicly release MV3 by the beginning of Dec (and this would land probably around end of Oct / early Nov).   

So my thinking for this PR is to just to not break the current MV2 version (did some local smoke testing with the extension against this branch, it still works).  

We could even skip this script by default unless MV3 is enabled (https://github.com/duckduckgo/content-scope-scripts/pull/148/files#diff-b34f13d583253af8ab988344e2052310ad4851cf846ab5e6a5be071aaa3cd582R37), since we'll have to do this work for other products too.

jonathanKingston:
This LGTM; we're holding off merging until we have the corresponding PRs :)

ladamski:
@jonathanKingston have gotten a bit stuck on messaging here, TBH.  Tried different permutations of `sendMessage` but couldn't get anything to work with any consistency, and it required a corresponding `externally_connectable` entry in the manifest.json (for the extension) which doesn't allow for domain wildcards anyway.

Switched to looking at postMessage, got close but the response channel gets stripped from the arguments so while I can send a message, apparently I can't reply to it. :/

I put this experiment in a separate branch. You can see where it calls postMessage: https://github.com/duckduckgo/content-scope-scripts/blob/la/ctl-messaging/src/utils.js#L424
and tries to reply here but the `channels` array is empty: https://github.com/duckduckgo/content-scope-scripts/blob/la/ctl-messaging/inject/chrome.js#L117

Any thoughts on how better to do round-trip messaging between the content script and its container?

jonathanKingston:
> Tried different permutations of sendMessage but couldn't get anything to work with any consistency, and it required a corresponding externally_connectable entry in the manifest.json (for the extension) which doesn't allow for domain wildcards anyway.

At a high level sendMessage won't work whilst in page context. We have two execution contexts in C-S-S; at least for the current extension implementations.

Firefox sits inside the content script scope and pikes through xrays to the page to modify prototypes; where as Chrome has a content script that then evaluates a <script> tag into the page so it can access globals and override prototypes.

So when you're calling sendMessage inside the page you're not accessing the API, it won't exist. To work around this; you're going to have to expose something to the page or listen to it in https://github.com/duckduckgo/content-scope-scripts/blob/main/inject/chrome.js (and the mozilla.js) these files can listen to CustomEvents that could then be passed on to the background code.

Something like:

utils.js
```javascript
sendMessage(details) {
  if (isChrome || isFirefox) {
      new CustomEvent('sendMessage', {details})
  }
  ...
}
```

mozilla.js/chrome.js
```
document.addEventListener('sendMessage', (message) => {
  browser.runtime.sendMessage(message.details)
})
```

We also have very similar code for MV3 https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/e921e6a7f8ae9df6f82596026ec9c6a45b1bd6f8/shared/js/content-scripts/content-scope-messaging.js we may wish to consolidate this as part of this work.

ladamski:
@jonathanKingston alright I think this is now functionally complete for both Chrome and Mozilla.  Please take a look and LMK your thoughts!  I'll leave some thoughts in the Asana task on overall landing strategy for this considering the other YT CTL work happening in parallel.

src/utils.js
```diff
@@ -492,3 +494,20 @@ export const windowsSpecificFeatures = ['windowsPermissionUsage']
 export function isWindowsSpecificFeature (featureName) {
     return windowsSpecificFeatures.includes(featureName)
 }
+
+export function createCustomEvent (eventName, eventDetail) {
+    // By default, Firefox protects the event detail Object from the page,
+    // leading to "Permission denied to access property" errors.
+    // See https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/Sharing_objects_with_page_scripts
+    if (typeof cloneInto === 'function') {
```
jonathanKingston, line 502:
```suggestion
    if (hasMozProxies) {
```
If we use this, the rollup code removes this code for other browsers.

src/features/ctl-assets.js
```diff
@@ -0,0 +1,9 @@
...a69GxkZuYaebehk+llgQcRENYKYI719/y8XTmEo8C3d4z9G8nI0nLTFapDsFLnVCnh09S4NLbColszVff41315Y67ke4VuPRkIhrEcD6MC7k+DuNTwbEIEtySJwzSqZqw/he3Z//bGgH1p1+DP+px//IVKUjzxDGbBTnwWLMSHynXdgJ7vcSDJXoiO9gl717efOj59/8HeeBlu1xyIlOebxPGzgg9E0vU0WtnIYQQuno17GXF3BXvZ4/2Phr3d/xC5HbnV/HBghwV7LdZLPgnoKGsMSrXRzhYOgiTuIZth8nUDXDNfe2AzgJO7Ue7QPiObqQ+xl7/ySDGINodlMkwN/2VSWTuU8qP42WtyvHARci1abF0LgegqbvjStyRRzdV8hCEsT3rE4k/UwQxqpiwxFVx8oA2GFfewmqPdalkNv+BrqqohVPBzEFq8gYmB/b7wwr0zN41BF1IvgMacKjJ4W0YftZwCBtGMR7BvMsSnAgI527BKOE8R6RlcXcBD/FP/610drjnqA2FApHIwQND5E4ZGnQuF8RpMV82HAXP0e58SWp092737y6NFTp6EOv4ZwXIO/ROhdLEhQiEF062prhdBH1u6WDN7GQOx++s6WLVv+9uhPt2FJ+Nm1NbwNj3zGsyFX2BQTq8Vw+F+8xYQiWKzqCOd9UmHW2yD6sO8RhP3O9Tl1CixSq1hZCD9+bKvTdTSkCU/nCJQU9OIt1rkQT4DVpogpThzuwx6LENZSmxA32LOF7es6IbecqVIqoWIhvE1sDC7apUU+7GsewpP9bpEW1k5qSxN7aXPE0oFPZyYDxYS8AoXVbZcEH/Zx5PdXLnimIcJOtyUzVzzGmN0shopTVT6JHzJSKeg185dDH3aKeFhS77TLlpJiE2okyBfdzxSVrxKAYteMHLYNyaSdFVJLM3Z0bFzHV6dJCXGiwkYng3R0TBuiGp/MldooC7LYIeRUSXu9fqpxyxwsM5dOHspKfBe2qTchR2cxgvKqtzNfvoncUGSPwwoWMyfu0H2wa3KfV+l6klrUhfgke4lLPXbSxJ4un+7eDGes2VScX5GUmtPqSCzKza9ry1jvXnQLVE24BExWVIbQBqtej/jPZYUwTfU6Zd4srwBmBqvegPgH0QH4hb3J/wYjbMbsQOx/YZgUFByi+r+W/wLr9CPL8dw0PgAAAABJRU5ErkJggg=='
```
jonathanKingston, line 1: Can this live in a directory above or somewhere else?
Currently everything in here has a config feature named after them.

inject/mozilla.js
```diff
@@ -25,15 +25,22 @@ function init () {
             })
         }
         contentScopeFeatures.init(message)
-    }
```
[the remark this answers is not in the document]
ladamski: 👍

src/utils.js
```diff
@@ -176,6 +176,8 @@ export function iterateDataKey (key, callback) {
 }
 
 export function isFeatureBroken (args, feature) {
+    if (feature === 'clickToLoad') feature = 'clickToPlay'
```
jonathanKingston, line 179: Can you just rename the feature? Theres a few places we check feature name that would be broken otherwise.

jonathanKingston:
Updated diff:
```diff
1,13c1,2
< /* global cloneInto */
< (function clickToLoad () {
<     function sendMessage (messageType, options) {
<         return new Promise((resolve, reject) => {
<             chrome.runtime.sendMessage({ messageType, options }, response => {
<                 if (chrome.runtime.lastError) {
<                     reject(new Error(chrome.runtime.lastError.message))
<                 } else {
<                     resolve(response)
<                 }
<             })
<         })
<     }
---
> import { createCustomEvent, sendMessage } from '../utils.js'
> import { logoImg, loadingImages, blockedFBLogo, ddgFont, ddgFontBold } from './ctl-assets.js'
15,26c4
<     function createCustomEvent (eventName, eventDetail) {
<         // By default, Firefox protects the event detail Object from the page,
<         // leading to "Permission denied to access property" errors.
<         // See https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/Sharing_objects_with_page_scripts
<         if (typeof cloneInto === 'function') {
<             eventDetail = cloneInto(eventDetail, window)
<         }
< 
<         return new CustomEvent(eventName, eventDetail)
<     }
< 
<     const devMode = sendMessage('getDevMode')
---
> let devMode = false
28,34c6
<     const loadingImages = {
<         darkMode: '',
<         lightMode: ''
<     }
<     let logoImg
<     let closeIcon
<     let shareFeedbackLink
---
> 
37,38d8
<     const ddgFont = chrome.runtime.getURL('public/font/ProximaNova-Reg-webfont.woff')
<     const ddgFontBold = chrome.runtime.getURL('public/font/ProximaNova-Bold-webfont.woff2')
71,73d40
<             `,
<             toggleButtonText: `
<                 color: #EEEEEE;
91,93d57
<             `,
<             toggleButtonText: `
<                 color: #666666;
117c81
<             margin: 0 auto;
---
>         margin: auto;
197,199d160
<             border: none;
<             padding: 0;
<             margin: 0;
217,219d177
<         dialogBlock: `
<             height: calc(100% - 30px);
<         `,
228,229c186
<             padding: 16px 0;
<             flex: 1 1 1px;
---
>         margin: auto;
231,245d187
<         feedbackLink: `
<             font-family: DuckDuckGoPrivacyEssentials;
<             font-style: normal;
<             font-weight: 400;
<             font-size: 12px;
<             line-height: 12px;
<             color: #ABABAB;
<             text-decoration: none;
<         `,
<         feedbackRow: `
<             height: 30px;
<             display: flex;
<             justify-content: flex-end;
<             align-items: center;
<         `,
252,253d193
<             margin: 0;
<             margin-bottom: 4px;
265,266d204
<             border: none;
<             padding: 0;
273,274c211
<             height: 100%;
<             align-items: flex-start;
---
>         padding-bottom: 36px;
281c218
<             margin: 10px auto;
---
>         margin: 27px auto 10px;
283,284d219
<             border: none;
<             padding: 0;
290c225
<             margin: 0px auto 14px;
---
>         margin: 0px auto 24px;
292,293d226
<             border: none;
<             padding: 0;
295,303d227
<         modalButtonRow: `
<             border: none;
<             padding: 0;
<             margin: auto;
<             width: 100%;
<             display: flex;
<             flex-direction: column;
<             align-items: center;
<         `,
305,308d228
<             width: 100%;
<             display: flex;
<             justify-content: center;
<             align-items: center;
334,348d253
<         closeIcon: `
<             height: 12px;
<             width: 12px;
<             margin: auto;
<         `,
<         closeButton: `
<             display: flex;
<             justify-content: center;
<             align-items: center;
<             min-width: 20px;
<             height: 21px;
<             border: 0;
<             background: transparent;
<             cursor: pointer;
<         `,
353,355d257
<             border: none;
<             padding: 0;
<             margin: 0;
368,369c270
<             width: 340px;
<             padding: 0;
---
>         width: 312px;
373,375c274,275
<             top: 50%;
<             left: 50%;
<             transform: translate(-50%, -50%);
---
>         left: calc(50% - 312px/2);
>         top: calc(50% - 356px/2 + 0.5px);
379d278
<             border: none;
385,386d283
<             border: none;
<             margin: 0;
397,399d293
<             border: none;
<             padding: 0;
<             margin: 0;
402,404c296,297
<             height: 100vh;
<             width: 100vw;
<             box-sizing: border-box;
---
>         height: 100%;
>         width: 100%;
408,410d300
<             border: 0;
<             margin: 0;
<             padding: 0;
442,477d331
<         `,
<         toggleButtonWrapper: `
<             display: flex;
<             align-items: center;
<             cursor: pointer;
<         `,
<         toggleButton: `
<             cursor: pointer;
<             position: relative;
<             width: 30px;
<             height: 16px;
<             margin-top: -3px;
<             margin: 0;
<             padding: 0;
<             border: none;
<             background-color: transparent;
<             text-align: left;
<         `,
<         toggleButtonBg: `
<             right: 0;
<             width: 30px;
<             height: 16px;
<             overflow: visible;
<             border-radius: 10px;
<         `,
<         toggleButtonText: `
<             display: inline-block;
<             margin: 0 0 0 7px;
<             padding: 0;
<         `,
<         toggleButtonBgState: {
<             active: `
<                 background: #3969EF;
<             `,
<             inactive: `
<                 background-color: #666666;
478a333,345
> }
> 
> /*********************************************************
>  *  config (temporarily here, until localized)
>  *********************************************************/
> const config = {
>     Facebook: {
>         informationalModal: {
>             icon: blockedFBLogo,
>             messageTitle: 'Logging in with Facebook lets them track you',
>             messageBody: "Once you're logged in, DuckDuckGo can't block Facebook content from tracking you on this site.",
>             confirmButtonText: 'Log In',
>             rejectButtonText: 'Go back'
480,497c347,354
<         toggleButtonKnob: `
<             position: absolute;
<             display: inline-block;
<             width: 14px;
<             height: 14px;
<             border-radius: 10px;
<             background-color: #ffffff;
<             margin-top: 1px;
<             top: calc(50% - 14px/2 - 1px);
<             box-shadow: 0px 0px 1px rgba(0, 0, 0, 0.05), 0px 1px 1px rgba(0, 0, 0, 0.1);
<         `,
<         toggleButtonKnobState: {
<             active: `
<                 right: 1px;
<             `,
<             inactive: `
<                 left: 1px;
<             `
---
>         elementData: {
>             'FB Like Button': {
>                 selectors: [
>                     '.fb-like'
>                 ],
>                 replaceSettings: {
>                     type: 'blank'
>                 }
499,578c356,364
<         youTubeWrapperDiv: `
<             position: relative;
<             overflow: hidden;
<             border-radius: 12px;
<             height: calc(100% - 30px);
<         `,
<         youTubeDialogButtonRow: `
<             display: flex;
<             flex-direction: column;
<             align-items: center;
<             justify-content: space-between;
<             height: 100%;
<         `,
<         youTubePlaceholder: `
<             display: flex;
<             flex-direction: column;
<             justify-content: flex-start;
<             position: relative;
<             width: 100%;
<             height: 100%;
<             background: rgba(45, 45, 45, 0.8);
<         `,
<         youTubePreviewWrapperImg: `
<             position: absolute;
<             display: flex;
<             justify-content: center;
<             align-items: center;
<             width: 100%;
<             height: 100%;
<         `,
<         youTubePreviewImg: `
<             min-width: 100%;
<             min-height: 100%;
<             height: auto;
<         `,
<         youTubeTitle: `
<             flex: 1;
<             font-family: DuckDuckGoPrivacyEssentialsBold;
<             position: relative;
<             padding: 18px 12px 0;
<             font-size: 14px;
<             font-weight: bold;
<             line-height: 14px;
<             color: #FFFFFF;
<             margin: 0;
<             width: 100%;
<             white-space: nowrap;
<             overflow: hidden;
<             text-overflow: ellipsis;
<             box-sizing: border-box;
<         `,
<         youTubePlayButtonRow: `
<             flex: 2;
<             display: flex;
<             align-items: center;
<             justify-content: center;
<         `,
<         youTubePlayButton: `
<             display: flex;
<             justify-content: center;
<             align-items: center;
<             height: 48px;
<             width: 80px;
<             padding: 0px 24px;
<             border-radius: 8px;
<         `,
<         youTubePreviewToggleRow: `
<             flex: 1;
<             display: flex;
<             flex-direction: column;
<             justify-content: center;
<             align-items: center;
<         `,
<         youTubePreviewToggleText: `
<             color: #EEEEEE;
<             font-weight: 400;
<         `,
<         youTubePreviewInfoText: `
<             color: #ABABAB;
<         `
---
>             'FB Button iFrames': {
>                 selectors: [
>                     "iframe[src*='://www.facebook.com/plugins/like.php']",
>                     "iframe[src*='://www.facebook.com/v2.0/plugins/like.php']",
>                     "iframe[src*='://www.facebook.com/plugins/share_button.php']",
>                     "iframe[src*='://www.facebook.com/v2.0/plugins/share_button.php']"
>                 ],
>                 replaceSettings: {
>                     type: 'blank'
579a366,700
>             },
>             'FB Save Button': {
>                 selectors: [
>                     '.fb-save'
>                 ],
>                 replaceSettings: {
>                     type: 'blank'
>                 }
>             },
>             'FB Share Button': {
>                 selectors: [
>                     '.fb-share-button'
>                 ],
>                 replaceSettings: {
>                     type: 'blank'
>                 }
>             },
>             'FB Page iFrames': {
>                 selectors: [
>                     "iframe[src*='://www.facebook.com/plugins/page.php']",
>                     "iframe[src*='://www.facebook.com/v2.0/plugins/page.php']"
>                 ],
>                 replaceSettings: {
>                     type: 'dialog',
>                     buttonText: 'Unblock Content',
>                     infoTitle: 'DuckDuckGo blocked this content to prevent Facebook from tracking you',
>                     infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                     simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                 },
>                 clickAction: {
>                     type: 'originalElement'
>                 }
>             },
>             'FB Page Div': {
>                 selectors: [
>                     '.fb-page'
>                 ],
>                 replaceSettings: {
>                     type: 'dialog',
>                     buttonText: 'Unblock Content',
>                     infoTitle: 'DuckDuckGo blocked this content to prevent Facebook from tracking you',
>                     infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                     simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                 },
>                 clickAction: {
>                     type: 'iFrame',
>                     targetURL: 'https://www.facebook.com/plugins/page.php?href=data-href&tabs=data-tabs&width=data-width&height=data-height',
>                     urlDataAttributesToPreserve: {
>                         'data-href': {
>                             default: '',
>                             required: true
>                         },
>                         'data-tabs': {
>                             default: 'timeline'
>                         },
>                         'data-height': {
>                             default: '500'
>                         },
>                         'data-width': {
>                             default: '500'
>                         }
>                     },
>                     styleDataAttributes: {
>                         width: {
>                             name: 'data-width',
>                             unit: 'px'
>                         },
>                         height: {
>                             name: 'data-height',
>                             unit: 'px'
>                         }
>                     }
>                 }
>             },
>             'FB Comment iFrames': {
>                 selectors: [
>                     "iframe[src*='://www.facebook.com/plugins/comment_embed.php']",
>                     "iframe[src*='://www.facebook.com/v2.0/plugins/comment_embed.php']"
>                 ],
>                 replaceSettings: {
>                     type: 'dialog',
>                     buttonText: 'Unblock Comment',
>                     infoTitle: 'DuckDuckGo blocked this comment to prevent Facebook from tracking you',
>                     infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                     simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                 },
>                 clickAction: {
>                     type: 'originalElement'
>                 }
>             },
>             'FB Comments': {
>                 selectors: [
>                     '.fb-comments',
>                     'fb\\:comments'
>                 ],
>                 replaceSettings: {
>                     type: 'dialog',
>                     buttonText: 'Unblock Comments',
>                     infoTitle: 'DuckDuckGo blocked these comments to prevent Facebook from tracking you',
>                     infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                     simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                 },
>                 clickAction: {
>                     type: 'allowFull',
>                     targetURL: 'https://www.facebook.com/v9.0/plugins/comments.php?href=data-href&numposts=data-numposts&sdk=joey&version=v9.0&width=data-width',
>                     urlDataAttributesToPreserve: {
>                         'data-href': {
>                             default: '',
>                             required: true
>                         },
>                         'data-numposts': {
>                             default: 10
>                         },
>                         'data-width': {
>                             default: '500'
>                         }
>                     }
>                 }
>             },
>             'FB Embedded Comment Div': {
>                 selectors: [
>                     '.fb-comment-embed'
>                 ],
>                 replaceSettings: {
>                     type: 'dialog',
>                     buttonText: 'Unblock Comment',
>                     infoTitle: 'DuckDuckGo blocked this comment to prevent Facebook from tracking you',
>                     infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                     simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                 },
>                 clickAction: {
>                     type: 'iFrame',
>                     targetURL: 'https://www.facebook.com/v9.0/plugins/comment_embed.php?href=data-href&sdk=joey&width=data-width&include_parent=data-include-parent',
>                     urlDataAttributesToPreserve: {
>                         'data-href': {
>                             default: '',
>                             required: true
>                         },
>                         'data-width': {
>                             default: '500'
>                         },
>                         'data-include-parent': {
>                             default: 'false'
>                         }
>                     },
>                     styleDataAttributes: {
>                         width: {
>                             name: 'data-width',
>                             unit: 'px'
>                         }
>                     }
>                 }
>             },
>             'FB Post iFrames': {
>                 selectors: [
>                     "iframe[src*='://www.facebook.com/plugins/post.php']",
>                     "iframe[src*='://www.facebook.com/v2.0/plugins/post.php']"
>                 ],
>                 replaceSettings: {
>                     type: 'dialog',
>                     buttonText: 'Unblock Post',
>                     infoTitle: 'DuckDuckGo blocked this post to prevent Facebook from tracking you',
>                     infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                     simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                 },
>                 clickAction: {
>                     type: 'originalElement'
>                 }
>             },
>             'FB Posts Div': {
>                 selectors: [
>                     '.fb-post'
>                 ],
>                 replaceSettings: {
>                     type: 'dialog',
>                     buttonText: 'Unblock Post',
>                     infoTitle: 'DuckDuckGo blocked this post to prevent Facebook from tracking you',
>                     infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                     simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                 },
>                 clickAction: {
>                     type: 'allowFull',
>                     targetURL: 'https://www.facebook.com/v9.0/plugins/post.php?href=data-href&sdk=joey&show_text=true&width=data-width',
>                     urlDataAttributesToPreserve: {
>                         'data-href': {
>                             default: '',
>                             required: true
>                         },
>                         'data-width': {
>                             default: '500'
>                         }
>                     },
>                     styleDataAttributes: {
>                         width: {
>                             name: 'data-width',
>                             unit: 'px'
>                         },
>                         height: {
>                             name: 'data-height',
>                             unit: 'px',
>                             fallbackAttribute: 'data-width'
>                         }
>                     }
>                 }
>             },
>             'FB Video iFrames': {
>                 selectors: [
>                     "iframe[src*='://www.facebook.com/plugins/video.php']",
>                     "iframe[src*='://www.facebook.com/v2.0/plugins/video.php']"
>                 ],
>                 replaceSettings: {
>                     type: 'dialog',
>                     buttonText: 'Unblock Video',
>                     infoTitle: 'DuckDuckGo blocked this video to prevent Facebook from tracking you',
>                     infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                     simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                 },
>                 clickAction: {
>                     type: 'originalElement'
>                 }
>             },
>             'FB Video': {
>                 selectors: [
>                     '.fb-video'
>                 ],
>                 replaceSettings: {
>                     type: 'dialog',
>                     buttonText: 'Unblock Video',
>                     infoTitle: 'DuckDuckGo blocked this video to prevent Facebook from tracking you',
>                     infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                     simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                 },
>                 clickAction: {
>                     type: 'iFrame',
>                     targetURL: 'https://www.facebook.com/plugins/video.php?href=data-href&show_text=true&width=data-width',
>                     urlDataAttributesToPreserve: {
>                         'data-href': {
>                             default: '',
>                             required: true
>                         },
>                         'data-width': {
>                             default: '500'
>                         }
>                     },
>                     styleDataAttributes: {
>                         width: {
>                             name: 'data-width',
>                             unit: 'px'
>                         },
>                         height: {
>                             name: 'data-height',
>                             unit: 'px',
>                             fallbackAttribute: 'data-width'
>                         }
>                     }
>                 }
>             },
>             'FB Group iFrames': {
>                 selectors: [
>                     "iframe[src*='://www.facebook.com/plugins/group.php']",
>                     "iframe[src*='://www.facebook.com/v2.0/plugins/group.php']"
>                 ],
>                 replaceSettings: {
>                     type: 'dialog',
>                     buttonText: 'Unblock Content',
>                     infoTitle: 'DuckDuckGo blocked this content to prevent Facebook from tracking you',
>                     infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                     simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                 },
>                 clickAction: {
>                     type: 'originalElement'
>                 }
>             },
>             'FB Group': {
>                 selectors: [
>                     '.fb-group'
>                 ],
>                 replaceSettings: {
>                     type: 'dialog',
>                     buttonText: 'Unblock Content',
>                     infoTitle: 'DuckDuckGo blocked this content to prevent Facebook from tracking you',
>                     infoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.',
>                     simpleInfoText: 'We blocked Facebook from tracking you when the page loaded. If you unblock this content, Facebook will know your activity.'
>                 },
>                 clickAction: {
>                     type: 'iFrame',
>                     targetURL: 'https://www.facebook.com/plugins/group.php?href=data-href&width=data-width',
>                     urlDataAttributesToPreserve: {
>                         'data-href': {
>                             default: '',
>                             required: true
>                         },
>                         'data-width': {
>                             default: '500'
>                         }
>                     },
>                     styleDataAttributes: {
>                         width: {
>                             name: 'data-width',
>                             unit: 'px'
>                         }
>                     }
>                 }
>             },
>             'FB Login Button': {
>                 selectors: [
>                     '.fb-login-button'
>                 ],
>                 replaceSettings: {
>                     type: 'loginButton',
>                     icon: blockedFBLogo,
>                     buttonText: 'Log in with Facebook',
>                     popupTitleText: 'DuckDuckGo blocked this Facebook login',
>                     popupBodyText: 'Facebook tracks your activity on a site when you use them to login.'
>                 },
>                 clickAction: {
>                     type: 'allowFull',
>                     targetURL: 'https://www.facebook.com/v9.0/plugins/login_button.php?app_id=app_id_replace&auto_logout_link=false&button_type=continue_with&sdk=joey&size=large&use_continue_as=false&width=',
>                     urlDataAttributesToPreserve: {
>                         'data-href': {
>                             default: '',
>                             required: true
>                         },
>                         'data-width': {
>                             default: '500'
>                         },
>                         app_id_replace: {
>                             default: 'null'
>                         }
>                     }
>                 }
>             }
>         }
>     }
> }
758,760c879,880
<             // Configure auto-play correctly depending on if the video's preview
<             // loaded.
<             let allowString = videoElement.getAttribute('allow') || ''
---
>         // Ensure the video doesn't auto-play.
>         const allowString = videoElement.getAttribute('allow') || ''
762,765d881
<             if (this.autoplay) {
<                 allowed.add('autoplay')
<                 url.searchParams.set('autoplay', '1')
<             } else {
768,770c884
<             }
<             allowString = Array.from(allowed).join('; ')
<             videoElement.setAttribute('allow', allowString)
---
>         videoElement.setAttribute('allow', Array.from(allowed).join('; '))
918,919c1032,1033
<     async function init (extensionResponseData) {
<         for (const entity of Object.keys(extensionResponseData)) {
---
> async function initCTL (resp) {
>     for (const entity of Object.keys(config)) {
921c1035
<             const { informationalModal, simpleVersion } = extensionResponseData[entity]
---
>         const { informationalModal, simpleVersion } = config[entity]
939c1053
<         await replaceClickToLoadElements(extensionResponseData)
---
>     await replaceClickToLoadElements(config)
942c1056
<             replaceClickToLoadElements(extensionResponseData, target)
---
>         replaceClickToLoadElements(config, target)
944,945d1057
< 
<         window.dispatchEvent(createCustomEvent('ddg-ctp-ready'))
950d1061
< 
976,1058d1086
<     function makeToggleButton (isActive = false, classNames = '', dataKey = '') {
<         const toggleButton = document.createElement('button')
<         toggleButton.className = classNames
<         toggleButton.style.cssText = styles.toggleButton
<         toggleButton.type = 'button'
<         toggleButton.setAttribute('aria-pressed', isActive ? 'true' : 'false')
<         toggleButton.setAttribute('data-key', dataKey)
< 
<         const toggleBg = document.createElement('div')
<         toggleBg.style.cssText = styles.toggleButtonBg + (isActive ? styles.toggleButtonBgState.active : styles.toggleButtonBgState.inactive)
< 
<         const toggleKnob = document.createElement('div')
<         toggleKnob.style.cssText = styles.toggleButtonKnob + (isActive ? styles.toggleButtonKnobState.active : styles.toggleButtonKnobState.inactive)
< 
<         toggleButton.appendChild(toggleBg)
<         toggleButton.appendChild(toggleKnob)
< 
<         return toggleButton
<     }
< 
<     function makeToggleButtonWithText (text, mode, isActive = false, toggleCssStyles = '', textCssStyles = '', dataKey = '') {
<         const wrapper = document.createElement('div')
<         wrapper.style.cssText = styles.toggleButtonWrapper
< 
<         const toggleButton = makeToggleButton(isActive, toggleCssStyles, dataKey)
< 
<         const textDiv = document.createElement('div')
<         textDiv.style.cssText = styles.contentText + styles.toggleButtonText + styles[mode].toggleButtonText + textCssStyles
<         textDiv.textContent = text
< 
<         wrapper.appendChild(toggleButton)
<         wrapper.appendChild(textDiv)
<         return wrapper
<     }
< 
<     async function createYouTubeBlockingDialog (trackingElement, widget) {
<         const bottomRow = document.createElement('div')
<         bottomRow.style.cssText = styles.youTubeDialogButtonRow
< 
<         const previewToggle = makeToggleButtonWithText(
<             widget.replaceSettings.previewToggleText,
<             widget.getMode(),
<             false,
<             '',
<             '',
<             'yt-preview-toggle'
<         )
<         previewToggle.addEventListener(
<             'click',
<             () => makeModal(widget.entity, () => sendMessage('updateSetting', {
<                 name: 'youtubePreviewsEnabled',
<                 value: true
<             }), widget.entity)
<         )
< 
<         const button = makeButton(widget.replaceSettings.buttonText, widget.getMode())
<         bottomRow.appendChild(button)
<         bottomRow.appendChild(previewToggle)
< 
<         const textButton = makeTextButton(widget.replaceSettings.buttonText, widget.getMode())
<         const { contentBlock, shadowRoot } = await createContentBlock(
<             widget, bottomRow, textButton
<         )
<         button.addEventListener('click', widget.clickFunction(trackingElement, contentBlock))
<         textButton.addEventListener('click', widget.clickFunction(trackingElement, contentBlock))
< 
<         // Size the placeholder element to match the original video
<         // element.
<         // Note: If the website later resizes the video element, the
<         //       placeholder will not resize to match.
<         const {
<             width: videoWidth,
<             height: videoHeight
<         } = window.getComputedStyle(trackingElement)
<         contentBlock.style.width = videoWidth
<         contentBlock.style.height = videoHeight
< 
<         return {
<             blockingDialog: contentBlock,
<             shadowRoot
<         }
<     }
< 
1073c1101
<             const icon = await sendMessage('getImage', widget.replaceSettings.icon)
---
>         const icon = widget.replaceSettings.icon
1084,1086c1112,1114
<         /** Facebook CTL */
<         if (widget.replaceSettings.type === 'dialog') {
<             const icon = await sendMessage('getImage', widget.replaceSettings.icon)
---
>     const youTubeVideo = widget.replaceSettings.type === 'youtube-video'
>     if (widget.replaceSettings.type === 'dialog' || youTubeVideo) {
>         const icon = widget.replaceSettings.icon
1095c1123,1125
<             replaceTrackingElement(widget, trackingElement, contentBlock)
---
>         replaceTrackingElement(
>             widget, trackingElement, contentBlock, /* hideTrackingElement= */youTubeVideo
>         )
1096a1127,1139
>         if (youTubeVideo) {
>             // Size the placeholder element to match the original video
>             // element.
>             // Note: If the website later resizes the video element, the
>             //       placeholder will not resize to match.
>             const {
>                 width: videoWidth,
>                 height: videoHeight
>             } = window.getComputedStyle(trackingElement)
>             contentBlock.style.width = videoWidth
>             contentBlock.style.height = videoHeight
>         }
> 
1108,1116d1150
< 
<         /** YouTube CTL */
<         if (widget.replaceSettings.type === 'youtube-video') {
<             await toggleYouTubeCTL(trackingElement, widget)
< 
<             /** TODO
<              *  Listen to `ddg-settings-youtubePreviewsEnabled` message to update YT CTL state
<              * (ie switch between blocked and preview modes)
<              */
1118d1151
<     }
1121,1155d1153
<      * @param {Element} trackingElement
<      *   The original tracking element (YouTube video iframe)
<      * @param {DuckWidget} widget
<      *   The CTP 'widget' associated with the tracking element.
<      */
<     async function toggleYouTubeCTL (trackingElement, widget) {
<         const youtubePreviewsEnabled = await getYouTubePreviewsEnabled()
< 
<         // Show YouTube Preview for embedded video
<         if (youtubePreviewsEnabled === true) {
<             const { placeholder } = await createYouTubePlaceholder(trackingElement, widget)
<             replaceTrackingElement(
<                 widget, trackingElement, placeholder, /* hideTrackingElement= */ true
<             )
< 
<         // Block YouTube embedded video and display blocking dialog
<         } else {
<             const { blockingDialog, shadowRoot } = await createYouTubeBlockingDialog(trackingElement, widget)
<             replaceTrackingElement(
<                 widget, trackingElement, blockingDialog, /* hideTrackingElement= */ true
<             )
<             // Show the extra unblock link in the header if the placeholder or
<             // its parent is too short for the normal unblock button to be visible.
<             // Note: This does not take into account the placeholder's vertical
<             //       position in the parent element.
<             const { height: placeholderHeight } = window.getComputedStyle(blockingDialog)
<             const { height: parentHeight } = window.getComputedStyle(blockingDialog.parentElement)
<             if (parseInt(placeholderHeight, 10) <= 200 || parseInt(parentHeight, 10) <= 200) {
<                 const titleRowTextButton = shadowRoot.querySelector(`#${titleID + 'TextButton'}`)
<                 titleRowTextButton.style.display = 'block'
<             }
<         }
<     }
< 
<     /**
1197,1245d1194
<     sendMessage('initClickToLoad').then(response => {
<         if (document.readyState === 'complete') {
<             init(response)
<         } else {
<             // Content script loaded before page content, so wait for load.
<             window.addEventListener('load', (event) => {
<                 init(response)
<             })
<         }
<     })
< 
<     // Fetch reusable assets
<     sendMessage('getLoadingImage', 'light').then(response => {
<         loadingImages.lightMode = response
<     })
< 
<     sendMessage('getLoadingImage', 'dark').then(response => {
<         loadingImages.darkMode = response
<     })
< 
<     sendMessage('getLogo', '').then(response => {
<         logoImg = response
<     })
< 
<     sendMessage('getCloseIcon', '').then(response => {
<         closeIcon = response
<     })
< 
<     // Listen for events from surrogates
<     addEventListener('ddg-ctp', (event) => {
<         if (!event.detail) return
<         const entity = event.detail.entity
<         if (!entities.includes(entity)) {
<             // Unknown entity, reject
<             return
<         }
<         if (event.detail.appID) {
<             appID = JSON.stringify(event.detail.appID).replace(/"/g, '')
<         }
<         // Handle login call
<         if (event.detail.action === 'login') {
<             if (entityData[entity].shouldShowLoginModal) {
<                 makeModal(entity, runLogin, entity)
<             } else {
<                 runLogin(entity)
<             }
<         }
<     })
< 
1310,1328d1258
<     function makeShareFeedbackLink () {
<         const feedbackLink = document.createElement('a')
<         feedbackLink.style.cssText = styles.feedbackLink
<         feedbackLink.href = shareFeedbackLink
<         feedbackLink.text = 'Share Feedback'
< 
<         return feedbackLink
<     }
< 
<     function makeShareFeedbackRow () {
<         const feedbackRow = document.createElement('div')
<         feedbackRow.style.cssText = styles.feedbackRow
< 
<         const feedbackLink = makeShareFeedbackLink()
<         feedbackRow.appendChild(feedbackLink)
< 
<         return feedbackRow
<     }
< 
1413,1414c1343
<         const icon = await sendMessage('getImage', entityData[entity].modalIcon)
< 
---
>     const icon = entityData[entity].modalIcon
1417,1426d1345
< 
<         const closeModal = () => {
<             document.body.removeChild(modalContainer)
<             cancelModal(entity)
<         }
< 
<         // Protect the contents of our modal inside a shadowRoot, to avoid
<         // it being styled by the website's stylesheets.
<         const shadowRoot = modalContainer.attachShadow({ mode: (await devMode) ? 'open' : 'closed' })
< 
1433c1352
<         const modalTitle = createTitleRow('DuckDuckGo', null, closeModal)
---
>     const modalTitle = createTitleRow('DuckDuckGo')
1460c1379
<         buttonRow.style.cssText = styles.modalButtonRow
---
>     buttonRow.style.cssText = 'margin:auto;'
1462c1381
<         allowButton.style.cssText += styles.modalButton + 'margin-bottom: 8px;'
---
>     allowButton.style.cssText += styles.modalButton + 'margin-right: 15px;'
1468,1469c1387,1391
<         rejectButton.style.cssText += styles.modalButton
<         rejectButton.addEventListener('click', closeModal)
---
>     rejectButton.style.cssText += styles.modalButton + 'float: right;'
>     rejectButton.addEventListener('click', function cancelLogin () {
>         document.body.removeChild(modalContainer)
>         cancelModal(entity)
>     })
1477,1479c1399,1400
<         shadowRoot.appendChild(pageOverlay)
<         shadowRoot.appendChild(modal)
< 
---
>     modalContainer.appendChild(pageOverlay)
>     modalContainer.appendChild(modal)
1483c1404
<     function createTitleRow (message, textButton, closeBtnFn) {
---
> function createTitleRow (message, textButton) {
1505,1517d1425
<         // Close Button
<         if (typeof closeBtnFn === 'function') {
<             const closeButton = document.createElement('button')
<             closeButton.style.cssText = styles.closeButton
<             const closeIconImg = document.createElement('img')
<             closeIconImg.setAttribute('src', closeIcon)
<             closeIconImg.setAttribute('height', '12px')
<             closeIconImg.style.cssText = styles.closeIcon
<             closeButton.appendChild(closeIconImg)
<             closeButton.addEventListener('click', closeBtnFn)
<             row.appendChild(closeButton)
<         }
< 
1542c1450
<         const shadowRootMode = (await devMode) ? 'open' : 'closed'
---
>     const shadowRootMode = devMode ? 'open' : 'closed'
1563,1565d1470
<         if (widget.replaceSettings.type === 'youtube-video') {
<             element.style.cssText += styles.dialogBlock
<         }
1613,1618d1517
<         /** Share Feedback Link */
<         if (widget.replaceSettings.type === 'youtube-video') {
<             const feedbackRow = makeShareFeedbackRow()
<             shadowRoot.appendChild(feedbackRow)
<         }
< 
1622,1623c1521,1531
<     function getYouTubeVideoDetails (videoURL) {
<         return sendMessage('getYouTubeVideoDetails', videoURL)
---
> const updateHandlers = {
>     // Convention is that each function should be named the same as the sendMessage method we are calling into
>     // eg. calling sendMessage('initClickToLoad') will result in a response routed to 'updateHandlers.initClickToLoad()'
>     initClickToLoad: function (resp) {
>         if (document.readyState === 'complete') {
>             initCTL(resp)
>         } else {
>             // Content script loaded before page content, so wait for load.
>             window.addEventListener('load', (event) => {
>                 initCTL(resp)
>             })
1625,1627c1533,1535
< 
<     function getYouTubePreviewsEnabled () {
<         return sendMessage('getSetting', { name: 'youtubePreviewsEnabled' })
---
>     },
>     getDevMode: function (resp) {
>         devMode = resp
1628a1537
> }
1630,1643c1539,1541
<     /**
<      * Creates the placeholder element to replace a YouTube video iframe element
<      * with a preview image. Mutates widget Object to set the autoplay property
<      * as the preview details load.
<      * @param {Element} originalElement
<      *   The YouTube video iframe element.
<      * @param {Object} widget
<      *   The widget Object. We mutate this to set the autoplay property.
<      * @returns {Object}
<      *   Object containing the placeholder element and its shadowRoot.
<      */
<     async function createYouTubePlaceholder (originalElement, widget) {
<         const placeholder = document.createElement('div')
<         placeholder.style.cssText = styles.wrapperDiv + styles.youTubeWrapperDiv
---
> export function init (args) {
>     sendMessage('getDevMode')
>     sendMessage('initClickToLoad')
1645,1746c1543,1549
<         // Put our custom font-faces inside the wrapper element, since
<         // @font-face does not work inside a shadowRoot.
<         // See https://github.com/mdn/interactive-examples/issues/887.
<         const fontFaceStyleElement = document.createElement('style')
<         fontFaceStyleElement.textContent = styles.fontStyle
<         placeholder.appendChild(fontFaceStyleElement)
< 
<         // Size the placeholder element to match the original video element.
<         // Note: The placeholder does later resize, even if the original video
<         //       element would have.
<         const { width, height } = originalElement.getBoundingClientRect()
<         placeholder.style.width = width + 'px'
<         placeholder.style.height = height + 'px'
< 
<         // Protect the contents of our placeholder inside a shadowRoot, to avoid
<         // it being styled by the website's stylesheets.
<         const shadowRoot = placeholder.attachShadow({ mode: (await devMode) ? 'open' : 'closed' })
< 
<         const youTubePreviewDiv = document.createElement('div')
<         youTubePreviewDiv.style.cssText = styles.youTubeWrapperDiv
<         shadowRoot.appendChild(youTubePreviewDiv)
< 
<         /** Preview Image */
<         const previewImageWrapper = document.createElement('div')
<         previewImageWrapper.style.cssText = styles.youTubePreviewWrapperImg
<         youTubePreviewDiv.appendChild(previewImageWrapper)
<         // We use an image element for the preview image so that we can ensure
<         // the referrer isn't passed.
<         const previewImageElement = document.createElement('img')
<         previewImageElement.setAttribute('referrerPolicy', 'no-referrer')
<         previewImageElement.style.cssText = styles.youTubePreviewImg
<         previewImageWrapper.appendChild(previewImageElement)
< 
<         const innerDiv = document.createElement('div')
<         innerDiv.style.cssText = styles.youTubePlaceholder
< 
<         /** Video Title */
<         const titleElement = document.createElement('p')
<         titleElement.style.cssText = styles.youTubeTitle
<         innerDiv.appendChild(titleElement)
< 
<         /** Play Button */
<         const playButtonRow = document.createElement('div')
<         playButtonRow.style.cssText = styles.youTubePlayButtonRow
< 
<         const playButton = document.createElement('button')
<         playButton.style.cssText = styles.button + styles.youTubePlayButton + styles[widget.getMode()].buttonBackground
< 
<         const videoPlayImg = document.createElement('img')
<         const videoPlayIcon = await sendMessage('getImage', widget.replaceSettings.placeholder.videoPlayIcon[widget.getMode()])
<         videoPlayImg.setAttribute('src', videoPlayIcon)
<         playButton.appendChild(videoPlayImg)
< 
<         playButton.addEventListener(
<             'click',
<             widget.clickFunction(originalElement, placeholder)
<         )
<         playButtonRow.appendChild(playButton)
<         innerDiv.appendChild(playButtonRow)
< 
<         /** Preview Toggle */
<         const previewToggleRow = document.createElement('div')
<         previewToggleRow.style.cssText = styles.youTubePreviewToggleRow
< 
<         const previewToggle = makeToggleButtonWithText(
<             widget.replaceSettings.placeholder.previewToggleEnabledText,
<             widget.getMode(),
<             true,
<             '',
<             styles.youTubePreviewToggleText,
<             'yt-preview-toggle'
<         )
<         previewToggle.addEventListener(
<             'click',
<             () => sendMessage('updateSetting', {
<                 name: 'youtubePreviewsEnabled',
<                 value: false
<             })
<         )
< 
<         /** Preview Info Text */
<         const previewText = document.createElement('div')
<         previewText.style.cssText = styles.contentText + styles.toggleButtonText + styles.youTubePreviewInfoText
<         previewText.innerText = widget.replaceSettings.placeholder.previewInfoText + ' '
<         previewText.appendChild(getLearnMoreLink())
< 
<         previewToggleRow.appendChild(previewToggle)
<         previewToggleRow.appendChild(previewText)
<         innerDiv.appendChild(previewToggleRow)
< 
<         youTubePreviewDiv.appendChild(innerDiv)
< 
<         widget.autoplay = false
<         // We use .then() instead of await here to show the placeholder right away
<         // while the YouTube endpoint takes it time to respond.
<         getYouTubeVideoDetails(originalElement.src).then(
<             ({ status, title, previewImage }) => {
<                 if (status === 'success') {
<                     titleElement.innerText = title
<                     titleElement.title = title
<                     if (previewImage) {
<                         previewImageElement.setAttribute('src', previewImage)
---
>     // Listen for events from surrogates
>     addEventListener('ddg-ctp', (event) => {
>         if (!event.detail) return
>         const entity = event.detail.entity
>         if (!entities.includes(entity)) {
>             // Unknown entity, reject
>             return
1748c1551,1552
<                     widget.autoplay = true
---
>         if (event.detail.appID) {
>             appID = JSON.stringify(event.detail.appID).replace(/"/g, '')
1749a1554,1559
>         // Handle login call
>         if (event.detail.action === 'login') {
>             if (entityData[entity].shouldShowLoginModal) {
>                 makeModal(entity, runLogin, entity)
>             } else {
>                 runLogin(entity)
1751c1561,1563
<         )
---
>         }
>     })
> }
1753,1755c1565,1566
<         /** Share Feedback Link */
<         const feedbackRow = makeShareFeedbackRow()
<         shadowRoot.appendChild(feedbackRow)
---
> export function update (args) {
>     if (!(args && args.type)) { return }
1757c1568,1571
<         return { placeholder, shadowRoot }
---
>     const fn = updateHandlers[args.type]
>     if (typeof fn !== 'function') { return }
> 
>     fn(args.response)
1759d1572
< })()
```

ladamski:
> Updated diff:

I looked at this diff but I'm not quite sure what its supposed to represent?  It doesn't apply (line offsets seem incorrect resulting in `patch: **** `<' expected at line 26 of patch`)

Furthermore it seems to delete many (all?) of the new YT changes (i.e. `createYouTubeBlockingDialog()`) from https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1433/files#diff-b739ace550706db1913f39023ec658272583c9b920f50576ad216ad164f289dcR1011)

Instead I'm going to try:
- diff the current click-to-load.js in this PR against the one it was based upon (i.e. before YT changes were landed), to capture just the content-scope-script related changes
- try to apply that diff to the latest click-to-load.js in the extension, merge the results by hand and copy back over to this PR
- modify the resulting file to fix messaging, etc.

If I can't get this working today, I'm going to put this on the back burner and switch to the MV3 rules work instead, that is more higher risk/urgency

src/utils.js
ladamski: This name is used in the config, etc.  Its not a trivial change to update it unfortunately, this was the most surgical fix that actually worked across the board.

jonathanKingston: Can't we just rename click-to-load.js to click-to-play.js nothing else depends on this naming after it moves here no?

jonathanKingston:
> I looked at this diff but I'm not quite sure what its supposed to represent?

This was for my own reference to evaluate the changes as before. Yeah this shows the changes to the current develop branch in the extension and this patch here.

src/utils.js
jonathanKingston: Testing the following worked for me:

```diff
diff --git a/src/content-scope-features.js b/src/content-scope-features.js
index 86b13f8..4ba4aec 100644
--- a/src/content-scope-features.js
+++ b/src/content-scope-features.js
@@ -34,7 +34,7 @@ export async function load () {
         'fingerprintingScreenSize',
         'fingerprintingTemporaryStorage',
         'navigatorInterface',
-        'clickToLoad'
+        'clickToPlay'
     ]
 
     for (const featureName of featureNames) {
diff --git a/src/utils.js b/src/utils.js
index a2aa148..841b61e 100644
--- a/src/utils.js
+++ b/src/utils.js
@@ -176,8 +176,6 @@ export function iterateDataKey (key, callback) {
 }
 
 export function isFeatureBroken (args, feature) {
-    if (feature === 'clickToLoad') feature = 'clickToPlay'
-
     return isWindowsSpecificFeature(feature)
         ? !args.site.enabledFeatures.includes(feature)
         : args.site.isBroken || args.site.allowlisted || !args.site.enabledFeatures.includes(feature)
```

and rename src/features/click-to-load.js -> src/features/click-to-play.js

src/features/ctl-assets.js
ladamski: Good call I moved it to a new directory: `src/assets`

src/utils.js
ladamski: 👍  done!

src/utils.js
ladamski: Well yeah except we should move to calling it clickToLoad, also the new config is clickToLoad.  I don't really care TBH, we can just keep using clickToPlay.  We'll have to update the new config to revert this too but it can wait.

ladamski: Ok reverted to using CTP, rebased, rebuild, tests pass locally.  Not sure why its complaining in CI, I just built again and there's nothing to push?

jonathanKingston: It's CTP in the config: https://github.com/duckduckgo/privacy-configuration/blob/main/features/click-to-play.json

It's failing the clean tree test, run npm build again and check in any assets changes. If that doesn't work check your node/npm version to match CI.

jonathanKingston: > Well yeah except we should move to calling it clickToLoad, also the new config is clickToLoad. I don't really care TBH, we can just keep using clickToPlay.

Which new config? The tds stuff?
Yeah I don't care what it is; if we're migrating the remote config name then let's do that automated etc and avoid these feature names in a function; there's other util functions that take a config name that would need changing.

ladamski: Yeah lets just punt on this for now, its not blocking.  There was much discussion to call it "click to load" vs "play" so was trying to be consistent, but it can wait.

jonathanKingston (approved):

PR #154, opened 2022-09-29
Simplify tab checks

jonathanKingston:
Changes:

- Renames getTabOrigin to getTabHostname
- Remove getTopLevelURL and replace with getTabHostname
- Explain that args.site.domain is still preferred; this is because in Firefox the URL will come from the background and circumvent the sandboxed iframe limitations that make the response of the method wrong.
- Type the processConfig method
- Change isUnprotectedDomain to accept a hostname instead of a URL object; this simplifies the code further.
- Add a .vscode config file as the generated files being in search is time consuming.


src/utils.js
```diff
@@ -455,18 +441,24 @@ export function isUnprotectedDomain (topLevelUrl, featureList) {
     return unprotectedDomain
 }
 
+/**
+ * @param {{ features: { [x: string]: { state: string; settings: any; exceptions: string[] }; }; unprotectedTemporary: string; }} data
```
shakyShane, line 445: swap `[x: string]` for a `Record<string, { state: string; settings: any; exceptions: string[] }>` like below

shakyShane (changes requested):
just a small about comment about using Record

PR #155, opened 2022-10-02
Collapse empty ad spaces (element hiding)

dharb:
This PR adds the `element-hiding.js` content script, which is injected into pages to hide empty ad spaces. This is similar functionality to what most ad blockers call 'cosmetic filtering' with one key caveat: we are only hiding spaces where ads didn't load due to blocked tracker requests. To accomplish this we use a heuristic to check each ad container selector for visible content, and only hide if no visible content (eg non-tracking ads) is found.

Initially this feature will only be enabled in the macOS app, with other client support to come as follow ups. This feature is v1, we will likely follow up with a more sophisticated approach in the near future.

This script functions based on [Privacy Configuration](https://github.com/duckduckgo/privacy-configuration) rules so that coverage may be expanded remotely without requiring updates to this repo or clients.

To see this functionality in action, navigate to a site that has a large blank space at the top of the page where a tracking banner ad didn't load, and see that the empty space is hidden shortly after page load. A few example sites to check are bloomberg.com, cyclingtips.com, and fandom.com.

Asana link: https://app.asana.com/0/0/1203092629639220/f
Corresponding privacy config PR: https://github.com/duckduckgo/privacy-configuration/pull/508


jonathanKingston:
As mentioned in MM this is working well in Firefox and Chrome nicely too (with https://github.com/duckduckgo/content-scope-scripts/pull/157 for FF).

src/features/element-hiding.js
```diff
@@ -0,0 +1,137 @@
...
+
+    // handle any ad containers that weren't added to the page within 300ms of page load
+    setTimeout(() => {
+        rules.forEach((rule) => {
+            const matchingElementArray = [...document.querySelectorAll(rule.selector)]
+            matchingElementArray.forEach((element) => {
+                collapseDomNode(element, rule.type)
+            })
+        })
+    }, 1000)
```
jonathanKingston, line 81:
```suggestion
    function hideMatchingNodesInner() {
        rules.forEach((rule) => {
            const matchingElementArray = [...document.querySelectorAll(rule.selector)]
            matchingElementArray.forEach((element) => {
                collapseDomNode(element, rule.type)
            })
        })
    }
    // wait 300ms before hiding ad containers so ads have a chance to load
    setTimeout(hideMatchingNodesInner, 300)

    // handle any ad containers that weren't added to the page within 300ms of page load
    setTimeout(hideMatchingNodesInner, 1000)
```

jonathanKingston (approved):
This is great! Thanks for your hard work :)

PR #169, opened 2022-10-14
Document platform key

jonathanKingston:
Documents the required platform key to be passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ The exposed API is a global called contentScopeFeatures and has three methods:
 - init
     - Calls the init method on all of the features
     - This should be passed the arguments object which has the following keys:
+        - 'platform' which is an object with:
+            - 'name' which is a string of 'android', 'ios', 'macos' or 'extension'
         - 'debug' true if debuging should be enabled
         - 'globalPrivacyControlValue' false if the user has disabled GPC.
         - 'sessionKey' a unique session based key.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,6 +12,8 @@ The exposed API is a global called contentScopeFeatures and has three methods:
 - init
     - Calls the init method on all of the features
     - This should be passed the arguments object which has the following keys:
+        - 'platform' which is an object with:
```
marcosholgado, line 15: Although I think this is fine because is needed for the init method, I think we are missing what the apps really need to pass. Something along the lines of what we have in Asana already.

$CONTENT_SCOPE$ - raw remote config object
$USER_UNPROTECTED_DOMAINS$ - an array of user allowlisted domains
$USER_PREFERENCES$ - 
platform: { name: '<ios | macos | extension | android>' }
debug: boolean
globalPrivacyControlValue: boolean
sessionKey: <random UUID string> (used for fingerprinting)

jonathanKingston: Let me add that in a separate section; that's specific to the replacement integration (used by all but the extension)

jonathanKingston: Ok got round to finishing this sorry for the delay. I've added a separate section for replacement keys.

marcosholgado (approved):
LGTM! Thanks @jonathanKingston!

PR #187, opened 2022-11-08
Hide empty ad space improvements

dharb:
This PR updates the way we hide empty ad spaces to decrease the delay between page load and hiding spaces. Previously we were waiting 300ms after page load to run the hide functions in order to give ads a chance to load; now we are running the hide functions immediately on page load and then following up with an unhide function that checks previously hidden elements to see if content was loaded into them after they were hidden.


Sources/ContentScopeScripts/dist/contentScope.js
```diff
@@ -3864,78 +3864,230 @@
...
+      const cachedDisplayProperties = {
+          display: element.style.display,
+          'min-height': element.style.minHeight,
+          height: element.style.height
+      };
+      hiddenElements.set(element, cachedDisplayProperties);
+
+      // apply styles to hide element
       element.style.setProperty('display', 'none', 'important');
+      element.style.setProperty('min-height', '0px', 'important');
```
dharb, line 3964: I added these to address a case where the child element of a leaderboard container has `position: fixed` and a set `min-height` value, so it appears even when the parent element has `display: none`.

Sources/ContentScopeScripts/dist/contentScope.js
```diff
@@ -3864,78 +3864,230 @@
...
+   * Apply relevant hiding rules to page at set intervals
+   * @param {Object[]} rules
+   * @param {string} rules[].selector
+   * @param {string} rules[].type
+   */
+  function applyRules (rules) {
+      // several passes are made to hide & unhide elements. this is necessary because we're not using
+      // a mutation observer but we want to hide/unhide elements as soon as possible, and ads
+      // frequently take from several hundred milliseconds to several seconds to load
+      // check at 0ms, 100ms, 200ms, 300ms, 400ms, 500ms, 1000ms, 1500ms, 2000ms, 2500ms, 3000ms
```
dharb, line 4029: This may be excessive, but there are a few variables (connection speed, website speed, device speed) that impact when ad containers are added to the page. On high end devices with a reasonable connection speed to an average website, most ad containers are on the page well before 500ms. But if you have a slow device, a bad connection, or are loading a slow website, this can take upwards of 1s--I've been testing this feature under each of these conditions and the slowest case I've found was on indiatimes.com, where when navigating from page to page using the top nav the leaderboard container can take ~2.5s to be added to the page.

src/features/element-hiding.js
```diff
@@ -1,78 +1,234 @@
...
+ * Apply relevant hiding rules to page at set intervals
+ * @param {Object[]} rules
+ * @param {string} rules[].selector
+ * @param {string} rules[].type
+ */
+function applyRules (rules) {
+    // several passes are made to hide & unhide elements. this is necessary because we're not using
+    // a mutation observer but we want to hide/unhide elements as soon as possible, and ads
+    // frequently take from several hundred milliseconds to several seconds to load
+    // check at 0ms, 100ms, 200ms, 300ms, 400ms, 500ms, 1000ms, 1500ms, 2000ms, 2500ms, 3000ms
```
dharb, line 170: This may be excessive, but there are a few variables (connection speed, website speed, device speed) that impact when ad containers are added to the page. On high end devices with a reasonable connection speed to an average website, most ad containers are on the page well before 500ms. But if you have a slow device, a bad connection, or are loading a slow website, this can take upwards of 1s--I've been testing this feature under each of these conditions and the slowest case I've found was on indiatimes.com, where when navigating from page to page using the top nav the leaderboard container can take ~2.5s to be added to the page.

src/features/element-hiding.js
```diff
@@ -1,78 +1,234 @@
...
+    const cachedDisplayProperties = {
+        display: element.style.display,
+        'min-height': element.style.minHeight,
+        height: element.style.height
+    }
+    hiddenElements.set(element, cachedDisplayProperties)
+
+    // apply styles to hide element
     element.style.setProperty('display', 'none', 'important')
+    element.style.setProperty('min-height', '0px', 'important')
```
dharb, line 105: I added these to address a case where the child element of a leaderboard container has `position: fixed` and a set `min-height` value, so it appears even when the parent element has `display: none`.

Sources/ContentScopeScripts/dist/contentScope.js
```diff
@@ -3864,78 +3864,230 @@
...
+      case 'closest-empty':
+          // iterate upwards from matching DOM elements until we arrive at previously
+          // hidden element. Unhide element if it contains visible content.
+          if (alreadyHidden) {
+              if (!isDomNodeEmpty(element)) {
+                  unhideNode(element);
+              }
+              break
+          } else {
+              expandNonEmptyDomNode(element.parentNode, rule);
```
jonathanKingston, line 3942: These are pretty similar, either make a function or just share the case statement?

```suggestion
      case 'hide-empty':
      case 'closest-empty':
          if (alreadyHidden && !isDomNodeEmpty(element)) {
              unhideNode(element);
          } else if (type === 'closest-empty') {
              // iterate upwards from matching DOM elements until we arrive at previously
              // hidden element. Unhide element if it contains visible content.
              expandNonEmptyDomNode(element.parentNode, rule);
          }
          break
```

src/features/element-hiding.js
```diff
@@ -1,78 +1,234 @@
...
+    case 'closest-empty':
+        // iterate upwards from matching DOM elements until we arrive at previously
+        // hidden element. Unhide element if it contains visible content.
+        if (alreadyHidden) {
+            if (!isDomNodeEmpty(element)) {
+                unhideNode(element)
+            }
+            break
+        } else {
+            expandNonEmptyDomNode(element.parentNode, rule, element)
```
jonathanKingston, line 81: These are pretty similar, either make a function or just share the case statement?

```suggestion
      case 'hide-empty':
      case 'closest-empty':
          if (alreadyHidden && !isDomNodeEmpty(element)) {
              unhideNode(element);
          } else if (type === 'closest-empty') {
              // iterate upwards from matching DOM elements until we arrive at previously
              // hidden element. Unhide element if it contains visible content.
              expandNonEmptyDomNode(element.parentNode, rule);
          }
          break
```

src/features/element-hiding.js
```diff
@@ -1,78 +1,234 @@
...
+    // we do this in order to display non-tracking ads that aren't blocked at the request level
+    // check at 750ms, 1500ms, 2250ms, 3000ms
+    let unhideIterations = 0
+    const unhideInterval = setInterval(function () {
+        unhideIterations += 1
+        if (unhideIterations === 3) {
+            clearInterval(unhideInterval)
+        }
+        unhideLoadedAds()
+    }, 750)
```
jonathanKingston, line 200: After these intervals can we clear all of out caches `hiddenElements` and `appliedRules`?

src/features/element-hiding.js
```diff
@@ -107,25 +263,23 @@ export function init (args) {
...
-            apply (target, thisArg, args) {
-                hideMatchingDomNodes(activeRules)
-                return DDGReflect.apply(target, thisArg, args)
-            }
-        })
-        historyMethodProxy.overload()
-    }
+    // we use proxy/reflect on history.pushState to call applyRules on page
+    // navigations, and listen for popstate events that indicate a back/forward navigation
+    const historyMethodProxy = new DDGProxy(featureName, History.prototype, 'pushState', {
```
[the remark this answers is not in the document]
dharb: `popstate` listener is a few lines below on L282, I can update the comment to be more clear about that.

Rationale for removing the `replaceState` code is I'm not sure it's necessary - the goal of this block is to activate the feature on single page app navigations. I looked at a bunch of SPAs and they all seem to use `pushState` on navigation to maintain history items. It's possible I'm overlooking something here and don't feel super strongly about it, let me know if you think I should change it back.

src/features/element-hiding.js
dharb: Good call, I'll address.

jonathanKingston (approved):
Looks great! Thanks for all the hard work here.

PR #193, opened 2022-11-16
Defend against missing options on cookie init

sammacbeth:
Fixes breaking tests.


jonathanKingston (approved):
👍🏻  thanks for fixing this

PR #194, opened 2022-11-18
Scope alwaysInitFeatures to extensions only

dharb:
This was causing an issue on the macOS browser where `init` was being called but not `load`, so `loadedPolicyResolve` in cookie.js was throwing an unhandled promise rejection error. This PR simply adds a platform check to the method.


jonathanKingston:
Thanks for jumping on the call and figuring this out!

/cc @sammacbeth I think this is safe to do as we don't need to call init in anything other than the extension due to it being the only platform that calls load for these.

I'll verify in the extension, but this PR unblocks @dharb with his patch as in MacOS we were calling init but not load causing a runtime exception about the promise not existing.

sammacbeth:
@jonathanKingston are you sure the cookie code isn't run at all on other platforms? Can we exclude that code entirely from these bundles to be sure?

jonathanKingston:
Right there's actually other avenues that this code is enabled outside those guards added in: https://github.com/duckduckgo/content-scope-scripts/commit/11f8fb9a7550b9de0f544f90f24bef469dde49b7# so not sure how this worked in testing; for other platforms nether load or init should be triggered.

We could do one of the following:

1. Tree shake at bundle time features that aren't enabled on a platform
2. Prevent cookie from load and init for all platforms except the extension
3. Remove the implementation of `load` for all other platforms than the extension, rename to `preExtensionLoad`. 

1. is likely the most involved to do well, 3. is still the cleanest whilst self documenting 'here be dragons' for the cookie+extension code paths.

jonathanKingston:
I don't hate it, but I think it's adding unnecessary complexity to a problem that's purely extension based as it stands.

I'd rather we do 3. as we've no intention to ever ship this to other platforms (at least in the near term) and then follow up with 1. as this is something we want anyway for perf reasons.

sammacbeth:
I just realized that we already have this block https://github.com/duckduckgo/content-scope-scripts/blob/main/src/features/cookie.js#L71-L74, which protections against what I was worried about, so I'm ok with the PR as-is.

jonathanKingston:
Ah yeah, forgot I added that!
Ok so let's just rebase and build the files (that's the failing tests).

jonathanKingston (approved):

PR #196, opened 2022-11-22
Adding YouTube CTL

franfaccin:
https://app.asana.com/0/0/1203325458386637/f


src/features/click-to-play.js
```diff
@@ -1229,6 +1623,53 @@ function getLearnMoreLink (mode) {
...
+    if (widget.originalElementStyle) {
+        return widget.originalElementStyle
+    }
+
+    const stylesToCopy = ['display', 'visibility', 'position', 'top', 'bottom', 'left', 'right',
+        'transform', 'margin', 'width', 'max-width', 'height', 'max-height']
+    widget.originalElementStyle = {}
+    const allOriginalElementStyles = getComputedStyle(originalElement)
+    for (const key of stylesToCopy) {
+        widget.originalElementStyle[key] = [
```
jonathanKingston, line 1642: Does this array have defined meaning rather than a set of the same data?

- If so, I would instead use an object with named keys. for example:
```
{
  value: allOriginalElementStyles[key],
  priority: originalElement.style.getPropertyPriority(key)
}
```

**However:** I see you indexing the 0th element, are you using the 1st element? Can we remove the priority key altogether? (this would simplify all the array access across the file for this key)

src/features/click-to-play.js
```diff
@@ -696,6 +946,72 @@ const config = {
                 }
             }
         }
+    },
+    'Google LLC': {
```
kzar, line 950: I think we need to change the entity name here, to match the block list changes.
```suggestion
    'Youtube': {
```

src/features/click-to-play.js
```diff
@@ -1029,10 +1355,10 @@ class DuckWidget {
     }
 }
 
-async function initCTL (resp) {
-    for (const entity of Object.keys(config)) {
+async function initCTL (configResponseData) {
```
ladamski, line 1358: Nit: I get renaming stuff but since we are both touching this code, it would be good to keep this to a minimum as it introduces merge conflicts

ladamski (changes requested):

src/features/click-to-play.js
```diff
@@ -1229,6 +1623,53 @@ function getLearnMoreLink (mode) {
...
+            originalElement.style.getPropertyPriority(key)
+        ]
+    }
+
+    // Try to get original height/width styles from element,
+    // otherwise measures current size and stores that
+    const { height: heightViewValue, width: widthViewValue } = originalElement.getBoundingClientRect()
+    let styleHeight = originalElement.height
+    styleHeight += styleHeight.search(/^\d+(\.\d+)*$/g) > -1 ? 'px' : ''
+    widget.originalElementStyle.height[0] = styleHeight || widget.originalElementStyle.height[0] || heightViewValue
```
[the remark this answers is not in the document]
franfaccin: The idea here was to try and make the placeholder reproduce the original sizing of the embedded video, and resize if the original would resize, as it was discussed in the Design Review. But I'm starting to see this being increasingly difficult to maintain and not always getting the results we would like.
I'll fix the size again and ping design about it. They were already aware it was going to be hard to make it work properly.

Just one added note regarding the `If the regex doesn't match we're going to use a raw unit value for the height is that intended?` question. I was checking both the `height` property and the `height` style, because they can have different values, or could be missing, and then `getBoundingClientRect()` would be a fallback. And the height property will return a value without pixels sometimes, that's why I was using the regex to check if only numbers and `.` existed, otherwise we would add the `px`. For other values like `%`, nothing changed. And finally, `getComputedStyle()` would always return the value with its unit, so adding `px` was not necessary. Yes, this is as confusing as it sounds, and that's why I agree with going back to a fixed size, and we can iterate based on user feedback later.
image

image

src/features/click-to-play.js
```diff
@@ -1533,11 +2264,23 @@ const updateHandlers = {
     },
     getDevMode: function (resp) {
         devMode = resp
+    },
+    getSetting: function (resp) {
```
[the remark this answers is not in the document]
franfaccin: Yeap, good catch. This passed right by me from the previous implementation and will be changed now.

src/features/click-to-play.js
franfaccin: 👍 ✅

src/features/click-to-play.js
franfaccin: 👍 ✅

src/features/click-to-play.js
```diff
@@ -696,6 +946,72 @@ const config = {
...
+    Youtube: {
+        domains: [
+            'youtube.com',
+            'youtube-nocookie.com'
+        ],
+        excludedSubdomains: [],
+        excludedDomains: [{
+            domain: 'duckduckgo.com',
+            reason: 'Existing privacy protections for YouTube videos'
+        }],
```
franfaccin, line 959: Remove this.

src/features/click-to-play.js
```diff
@@ -696,6 +946,72 @@ const config = {
                 }
             }
         }
+    },
+    Youtube: {
```
franfaccin, line 950: Comment out before merging

ladamski (approved):

jonathanKingston (approved):
SGTM thanks!

kzar (approved):
_(I have not reviewed this, but Jonathan and Lucas' LGTMs got cleared by a comment change, so rubber-stamping.)_

PR #202, opened 2022-12-05
Add timeout to initial element hiding check

dharb:
This fixes a layout issue on cnn.com/business/tech where the element hiding code was running before the page finished some layout changes, which caused a lingering white space at the top of the page.

Asana link: https://app.asana.com/0/0/1203494166470295/f


jonathanKingston (approved):
Nice!

PR #212, opened 2022-12-14
Bundle tracker lookup and config with extension scripts

sammacbeth:
Exposes variables for the extension to inject:
 1. A lookup data structure for tracker domains
 2. The privacy config at release-time.

Injecting these allows us to check things at feature `init` time and proactively apply protections with more precision. This is mainly of concern for cookie protection, which may have to block cookies before the extension background can send a message to tell the script if exceptions apply.


inject/chrome.js
```diff
@@ -20,7 +20,28 @@ function randomString () {
     return num.toString().replace('0.', '')
 }
 
+/**
+ * @see ../src/trackers.js
+ */
+function isTrackerOrigin (trackerLookup) {
```
sammacbeth, line 26: @jonathanKingston I've just copied and pasted this in instead of importing as it seems the chrome inject.js doesn't go through rollup. Would that be an easy change, that I could tackle in this PR, or would I be best leaving it as-is?

inject/chrome.js
```diff
@@ -20,7 +20,28 @@ function randomString () {
...
+        } else {
+            return false
+        }
+    }
+    return false
+}
+
 function init () {
+    const documentOriginIsTracker = isTrackerOrigin($TRACKER_LOOKUP$)
+    const bundledConfig = $BUNDLED_CONFIG$
```
sammacbeth, line 44: The extension config is pretty massive, and most of it seems irrelevant to C-S-S. Should we prune it somehow before injection?

inject/chrome.js
jonathanKingston: I think the change would be made here: https://github.com/duckduckgo/content-scope-scripts/blob/main/scripts/inject.mjs#L77 I don't think there's any rationale not to do that other than we didn't need it initially. :)

inject/chrome.js
sammacbeth: Yeah, we'll do this on the extension side. I'll add it to that PR.

jonathanKingston (approved):
LGTM :)

sammacbeth:
Can I get a re-approve so I can merge this @jonathanKingston ?

PR #215, opened 2022-12-15
YouTube CTL - design fixes

franfaccin:
https://app.asana.com/0/0/1203493249197456/f


src/features/click-to-play.js
```diff
@@ -998,9 +1016,9 @@ const config = {
     //     },
     //     informationalModal: {
     //         icon: blockedYTVideo,
-    //         messageTitle: 'Enable YouTube previews and reduce privacy?',
+    //         messageTitle: 'Enable all YouTube previews?',
     //         messageBody: 'Showing previews will allow Google (which owns YouTube) to see some of your device’s information, but is still more private than playing the video.',
-    //         confirmButtonText: 'Enable Previews',
+    //         confirmButtonText: 'Enable All Previews',
```
kzar, line 1021: I notice the capitalisation of all (`all` vs `All`) is different for these. If that's intended then feel free to ignore, just mentioning it in case it's a mistake.

kzar (changes requested):

src/features/click-to-play.js
```diff
@@ -2247,14 +2279,14 @@ const updateHandlers = {
         devMode = resp
     },
     getYoutubePreviewsEnabled: function (resp) {
```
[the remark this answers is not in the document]
franfaccin: I don't have strong preferences for this, but I think we should keep the pattern that was already introduced for other `updateHandlers` functions. So I'd like to defer to @ladamski: Lucas, would like for us to start using method shorthands here? I'm happy to update the other functions in this PR if we want it.

src/features/click-to-play.js
franfaccin: Thank you for pointing this out. But it is intentional, as part of the revision of the [copy](https://app.asana.com/0/0/1203533466418466/1203564528435112/f). The buttons have capitalised text, so my understanding is that we're keeping the format.

src/features/click-to-play.js
```diff
@@ -1661,13 +1679,18 @@ function makeTextButton (linkText, mode) {
...
 
-function makeButton (buttonText, mode) {
+function makeButton (buttonText, mode = 'lightMode') {
     const button = document.createElement('button')
     button.style.cssText = styles.button + styles[mode].buttonBackground
-    const textContainer = document.createElement('div')
-    textContainer.style.cssText = styles.buttonTextContainer + styles[mode].buttonFont
-    textContainer.textContent = buttonText
-    button.appendChild(textContainer)
+    button.addEventListener('mouseenter', function () { button.style.cssText += styles[mode].buttonBackgroundHover })
```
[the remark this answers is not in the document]
franfaccin: I've changed has these styles work in the latest commit. I'm adding a `<style/>` to the wrapper element and it contains the button styles, so there is no need to manually update the styles based on button states anymore.

src/features/click-to-play.js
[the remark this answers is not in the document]
kzar: The code is equivalent, the new syntax is just nicer. I don't mind too much, let's skip this for now I'll do it as a follow-up if I have time.

src/features/click-to-play.js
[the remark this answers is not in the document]
kzar: Nice, looks good :+1:

src/features/click-to-play.js
```diff
@@ -1654,20 +1673,90 @@ function copyStylesTo (originalStyles, element) {
...
-function makeButton (buttonText, mode) {
+function makeButton (buttonText, mode = 'lightMode') {
     const button = document.createElement('button')
-    button.style.cssText = styles.button + styles[mode].buttonBackground
-    const textContainer = document.createElement('div')
-    textContainer.style.cssText = styles.buttonTextContainer + styles[mode].buttonFont
-    textContainer.textContent = buttonText
-    button.appendChild(textContainer)
+    button.classList.add('DuckDuckGoButton')
+    button.classList.add(mode !== 'cancelMode' ? 'primary' : 'secondary')
```
kzar, line 1754: Nit: Invert the condition?

```suggestion
    button.classList.add(mode === 'cancelMode' ? 'secondary' : 'primary')
```

src/features/click-to-play.js
```diff
@@ -1654,20 +1673,90 @@ function copyStylesTo (originalStyles, element) {
...
+
 function makeTextButton (linkText, mode) {
     const linkElement = document.createElement('a')
     linkElement.style.cssText = styles.headerLink + styles[mode].linkFont
     linkElement.textContent = linkText
     return linkElement
 }
 
-function makeButton (buttonText, mode) {
+function makeButton (buttonText, mode = 'lightMode') {
```
kzar, line 1751: Any chance you could add a JSDoc comment for this function that explains the idea of the `mode` (possible values and what they do / what they are for)?

src/features/click-to-play.js
```diff
@@ -2099,11 +2182,7 @@ async function createYouTubePreview (originalElement, widget) {
     youTubePreview.id = `yt-ctl-preview-${widget.widgetID}`
     youTubePreview.style.cssText = styles.wrapperDiv + styles.placeholderWrapperDiv
 
-    // Put our custom font-faces inside the wrapper element, since
-    // @font-face does not work inside a shadowRoot.
-    // See https://github.com/mdn/interactive-examples/issues/887.
-    const fontFaceStyleElement = document.createElement('style')
-    fontFaceStyleElement.textContent = styles.fontStyle
+    const fontFaceStyleElement = makeFontFaceStyleElement()
```
kzar, line 2185: Nit: Might as well skip the intermediary variable?
```suggestion 
   youTubePreview.appendChild(makeFontFaceStyleElement())
```

src/features/click-to-play.js
```diff
@@ -2144,7 +2226,8 @@ async function createYouTubePreview (originalElement, widget) {
     topSection.appendChild(titleElement)
 
     /** Text Button on top section */
-    const textButton = makeTextButton(widget.replaceSettings.buttonText, widget.getMode())
+    // Use darkMode styles because of preview background
```
kzar, line 2229: Maybe explain why? E.g. "Use darkMode styles because video preview background can make button text hard to read otherwise"? (Same below.)

kzar (changes requested):
Looks great, just a couple of nits.

kzar (approved):
LGTM :+1:

PR #224, opened 2023-01-03
Include YouTube Click to Load configuration

kzar:
Since it is now up to the extension to decide which Click to Load
entities are enabled, let's include the YouTube Click to Load
configuration ready for when the extension enables it.

Also, let's take care to pass the correct blocking
action ("block-ctl-yt" or "block-ctl-fb") when unblocking the embedded
content.


kzar:
_(These changes are untested so far, because they depend on https://github.com/duckduckgo/privacy-configuration/pull/603, https://github.com/duckduckgo/ddg2dnr/pull/67 and some more extension changes. Shout if you can see anything obvious I'm missing though @franfaccin!)_

franfaccin:
Thank you for the ping @kzar! What comes to mind is that with these changes we can probably update the YT CTL [test configuration](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/integration-test/data/configs/click-to-load-youtube.json) and enable them again ❤️

ladamski (dismissed):

franfaccin (approved):
LGTM ✅

PR #226, opened 2023-01-10
Fix Facebook Click to Load logins and add related JSDoc comment

kzar:
There was a typo in one of the calls to `enableSocialTracker` which
caused Facebook requests to not be allowed properly when the user
clicked to login via Facebook. That in turn broke the "Log in via
Facebook" functionality.

Let's fix that typo, but also add a JSDoc comment to make the function
easier to use correctly in the future.


kzar:
@jonathanKingston I also double-checked the old code and confirmed `isLogin` should be true. See https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/519491f2b8cc77815c15dc4b1e9f60a2ba16a4b3/shared/js/content-scripts/click-to-load.js#L1246-L1255

Sources/ContentScopeScripts/dist/contentScope.js
```diff
@@ -3538,12 +3538,32 @@
...
    *  Messaging to surrogates & extension
    *********************************************************/
+
+  /**
+   * @typedef enableSocialTrackerRequest
+   * @property {string} entity
+   *   The entity to unblock requests for (e.g. "Facebook").
+   * @property {bool} [isLogin=false]
+   *   True if we should "allow social login", defaults to false.
+   * @property {string} action
```
kzar, line 3548: Note: It's also a bug that `action` is not being passed, but that is only for Chrome MV3. I have follow-up changes MV3 changes anyway, so will address that with those.

ladamski (approved):
Looks good to me, tested and it seems to work correctly for both scenarios (FB already logged in and not)

PR #229, opened 2023-01-13
Element hiding: inject strict hide rules as style tag if enabled by privacy config

dharb:
**Asana link:** https://app.asana.com/0/1203557590179903/1203556051691984/f
**Corresponding privacy config PR:** https://github.com/duckduckgo/privacy-configuration/pull/619

**Description**
This PR adds the ability to inject a style tag for strict hide rules. This can be toggled on/off globally and for specific domains using privacy config. If toggled off, element hiding will work exactly as it did prior to this change.


src/features/element-hiding.js
```diff
@@ -157,51 +158,83 @@ function isDomNodeEmpty (node) {
...
+    rules.forEach((rule, i) => {
+        if (i === rules.length - 1) {
+            styleTag = styleTag.concat(rule.selector, '{')
+        } else {
+            styleTag = styleTag.concat(rule.selector, ',')
+        }
+    })
+
+    styleTag = styleTag.concat('display:none!important;min-height:0!important;height:0!important;}</style>')
+    document.head.insertAdjacentHTML('beforeend', styleTag)
```
jonathanKingston, line 235: This is risking XSS through using a variant of innerHTML; instead we should create the style element, innerText the rules and then append the node.

jonathanKingston (changes requested):
We should fix the potential XSS risk here (even if from our own remote). 

Are we not going to allow for remotely disabling the style tag injection?

dharb:
> Are we not going to allow for remotely disabling the style tag injection?

No we are - see [here](https://github.com/duckduckgo/content-scope-scripts/pull/229/files#diff-fe23b6c57bd95dc1c110287c02534dc14b910516ac0dd67fce7610ae7ff8a628R280-R287) and [here](https://github.com/duckduckgo/content-scope-scripts/pull/229/files#diff-fe23b6c57bd95dc1c110287c02534dc14b910516ac0dd67fce7610ae7ff8a628R198-R200), as well as the corresponding privacy config changes in https://github.com/duckduckgo/privacy-configuration/pull/619. We can disable globally or by domain.

I'll address the styletag potential xss issue you flagged now.

src/features/element-hiding.js
dharb: Good catch, updated to address.

jonathanKingston (dismissed):
Looks great to me thanks for making this configurable! :)

jonathanKingston (approved):

PR #230, opened 2023-01-17
Rethink the Click to Load initialisation logic

kzar:
- Avoid messaging the huge placeholder configuration Object to the
  platform, just to have it sent back again.
- Use the already available parentEntity of the current website,
  instead of asking the platform for it.
- Determine if "clicksBeforeSimpleVersion" has been exceeded here,
  instead of relying on the platform to determine that.
- Take care to remove entities from the placeholder configuration
  Object that are not also enabled in the feature configuration.
- Take care to reduce the number of messages sent during
  initialisation with the happy side-effect of fixing a race
  condition.
- Remove the now unused message handlers and rename the
  "enableSocialTracker" message to "unblockClickToLoadContent".


kzar:
@ladamski see [the corresponding (WIP) extension changes here](https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1646). Cc @jonathanKingston

src/features/click-to-play.js
```diff
@@ -2372,9 +2371,48 @@ const updateHandlers = {
...
+            currentEntityData.modalTitle = informationalModal.messageTitle
+            currentEntityData.modalText = informationalModal.messageBody
+            currentEntityData.modalAcceptText = informationalModal.confirmButtonText
+            currentEntityData.modalRejectText = informationalModal.rejectButtonText
+        }
+
+        entityData[entity] = currentEntityData
+    }
+
+    // TODO: Fix the race condition where the getDevMode or
```
ladamski, line 2410: Agree, we should have just one settings getter here.  Looks like we're already getting maxClicks above anyway; could we just pass everything thing above via init (args)?

src/features/click-to-play.js
```diff
@@ -2337,16 +2324,28 @@ async function createYouTubePreview (originalElement, widget) {
...
-    initClickToLoad: function (resp) {
+    getClickToLoadClicks (response) {
+        for (const [entity, clickCount] of Object.entries(response)) {
+            if (entityData[entity]) {
+                entityData[entity].simpleVersion =
+                    clickCount >= entityData[entity].maxClicks
+            }
+        }
+
+        // TODO: Move the below init logic, it's confusing having it inside the
```
ladamski, line 2339: +1 :)

Or call this something generic like "getCTLSettings" and have it return a hash of all the things.

kzar: Good idea, I've combined them into `getClickToLoadState` which fixes the race condition + further reduces the messages required on init :+1:.

src/features/click-to-play.js
```diff
@@ -2337,22 +2324,36 @@ async function createYouTubePreview (originalElement, widget) {
...
             window.addEventListener('load', (event) => {
-                initCTL(resp)
+                initCTL()
             })
         }
     },
-    getDevMode: function (resp) {
-        devMode = resp
-    },
     getYoutubePreviewsEnabled: function (resp) {
```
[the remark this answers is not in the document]
kzar: Yea, I thought I'd leave it in since there's a setter + the `ddg-settings-youtubePreviewsEnabled` logic. But I don't feel strongly about it, Done.

src/features/click-to-play.js
```diff
@@ -1600,15 +1587,15 @@ async function replaceClickToLoadElements (config, targetElement) {
...
  * the page.
- * @param {enableSocialTrackerRequest} message
- * @see {@event ddg-ctp-enableSocialTracker-complete} for the response handler.
+ * @param {unblockClickToLoadContentRequest} message
+ * @see {@event ddg-ctp-unblockClickToLoadContent-complete} for the response handler.
  */
-function enableSocialTracker (message) {
-    sendMessage('enableSocialTracker', message)
+function unblockClickToLoadContent (message) {
+    sendMessage('unblockClickToLoadContent', message)
```
kzar, line 1594: Since we're now in the realm of tidying up the messages, I have taken the opportunity to rename "enableSocialTracker" to "unblockClickToLoadContent", since that name was pretty confusing. The other names weren't too bad though, so I left them alone.

src/features/click-to-play.js
```diff
@@ -2366,17 +2364,49 @@ const updateHandlers = {
             originalWindowDispatchEvent(new OriginalCustomEvent('ddg-ctp-youTubeVideoDetails', { detail: resp }))
         }
     },
-    enableSocialTracker: function (resp) {
-        originalWindowDispatchEvent(new OriginalCustomEvent('ddg-ctp-enableSocialTracker-complete', { detail: resp }))
+    unblockClickToLoadContent () {
+        originalWindowDispatchEvent(new OriginalCustomEvent('ddg-ctp-unblockClickToLoadContent-complete'))
```
kzar, line 2368: Note: I removed the `resp` parameter and stopped passing it through. There is no response and the `ddg-ctp-unblockClickToLoadContent-complete` event listener doesn't use it either.

ladamski (approved):
LGTM!

PR #231, opened 2023-01-20
Fix typo in Click to Load script

kzar:
When adding logic to determine the correct tracker blocking rule
action based on Click to Load entity, I made a small typo. Instead of
`entity`, I attempted to access `this.entity` which didn't exist.


franfaccin (approved):
✅

PR #238, opened 2023-01-27
Localize ClickToLoad

ladamski:
Task: https://app.asana.com/0/inbox/1171671347221380/1202416379154694/1203770191565262/f


ladamski:
@kzar would love to get your feedback on general approach here!  A few things are still in flux (per https://app.asana.com/0/0/1202416379154694/1203870644873836/f), but the overall structure is there

src/features/click-to-play.js
```diff
@@ -606,425 +603,419 @@ const styles = {
...
+
+function getConfig (locale) {
+    const locales = JSON.parse(localesJSON)
+    const fbStrings = locales[locale]['facebook.json']
+
+    return {
+        Facebook: {
+            informationalModal: {
+                icon: blockedFBLogo,
+                messageTitle: `${fbStrings.informationalModalMessageTitle.title}`,
```
GioSensation, line 618: If these are all simple strings there's no need to wrap them in `` `${}` ``, you can just have the variable.
```suggestion
                messageTitle: fbStrings.informationalModalMessageTitle.title,
```

ladamski: Ah good catch, yes templates don't really add anything here!

scripts/buildLocales.js
```diff
@@ -0,0 +1,25 @@
...
+
+const locales = {}
+const localesDirs = fs.readdirSync(localesRoot).filter(f => !f.startsWith('.'))
+for (const l of localesDirs) {
+    locales[l] = {}
+    const dir = `${localesRoot}/${l}`
+    const files = fs.readdirSync(dir)
+    for (const f of files) {
+        const localeJSON = fs.readFileSync(`${dir}/${f}`)
+        locales[l][f] = JSON.parse(localeJSON)
```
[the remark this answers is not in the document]
ladamski: 👍 @jonathanKingston cleaned this up per https://app.asana.com/0/0/1203778992058963/1203876937309551/f

kzar:
> ..should set the locale in `init()`... or inside `getClickToLoadState()`.

My vote would be that the locale is passed to `init()`. That way, all features have access to the locale, which I figure they will need to translate strings too. In other words, I don't think needing to know the user's locale is specific to Click to Load.

src/features/click-to-play.js
```diff
@@ -1,1034 +1,27 @@
...
+let config = {}
+let sharedStrings = {}
+
 // TODO: Remove these redundant data structures and refactor the related code.
 //       There should be no need to have the entity configuration stored in two
 //       places.
 const entities = []
 const entityData = {}
 
-/*********************************************************
```
kzar, line 19: Would you mind splitting the changes up, so that the first PR is to just move the config into the separate file? I think that would make the rest of the changes a lot easier to review.

src/features/click-to-play.js
```diff
@@ -2374,6 +1350,10 @@ const updateHandlers = {
 export function init (args) {
     const websiteOwner = args?.site?.parentEntity
     const settings = args?.featureSettings?.clickToPlay || {}
+    const locale = args?.locale || 'en'
+    const localizedConfig = getConfig(locale)
```
kzar, line 1354: IMO the localisation shouldn't be mixed up with getConfig. I figure the config should just be data, with the string IDs. Then content-scope-scripts should provide a utility function that takes a string ID and the locale and returns the appropriate translation (that way other features could use it too). Then this script calls the translation function whenever it wants to turn one of the string IDs into the text to display to the user.

src/features/click-to-play.js
ladamski: That makes sense, though lets resolve the other conversation first so this will fit whatever pattern we decide on.

src/features/click-to-play.js
ladamski: I revisited this in https://app.asana.com/0/0/1203778992058963/1203923914388384/f, as it seems like a design decision that should be covered there?

scripts/buildLocales.js
```diff
@@ -0,0 +1,31 @@
+import * as fs from 'fs'
```
kzar, line 1: Nit:
```suggestion
import { readdirSync, readFileSync } from 'fs'
```

scripts/buildLocales.js
```diff
@@ -0,0 +1,31 @@
...
+const localesRoot = process.argv[2]
+if (!localesRoot) {
+    console.log('usage: buildLocales <locales/feature dir>')
+}
+
+const locales = {}
+const localesDirs = fs.readdirSync(localesRoot).filter(f => !f.startsWith('.'))
+for (const l of localesDirs) {
+    locales[l] = {}
+    const dir = `${localesRoot}/${l}`
```
kzar, line 14: Probably better to use `path.join`, that takes care to get the separator right depending on platform. See https://nodejs.org/api/path.html#pathjoinpaths

(Same below.)

scripts/buildLocales.js
```diff
@@ -0,0 +1,31 @@
...
+        for (const [key, value] of Object.entries(stringObj)) {
+            if (key !== 'smartling') {
+                locales[l][f][key] = value.title
+            }
+        }
+    }
+}
+
+const output = 'export default `' +
+    JSON.stringify(locales).replace('`', '\\`') +
```
kzar, line 29: I'm not sure the correct way to do this, but this line makes me nervous. I'm thinking the translations are untrusted input (came from a third-party service) and this output is going into our extension, so we should be careful to get escaping right here. What do you think @jonathanKingston?

src/features/click-to-play.js
```diff
@@ -834,7 +840,7 @@ function makeShareFeedbackRow () {
 }
 
 /* FB login replacement button, with hover text */
-function makeLoginButton (buttonText, mode, hoverTextTitle, hoverTextBody, icon, originalElement) {
```
ladamski, line 837: Note that it looks like `hoverTextTitle` had been moved to `styles` but it remained here as a method argument, which seems like a bug?

src/features/click-to-play.js
```diff
@@ -1,13 +1,19 @@
...
 let devMode = false
 let isYoutubePreviewsEnabled = false
 let appID
 
 const titleID = 'DuckDuckGoPrivacyEssentialsCTLElementTitle'
 
+/*********************************************************
+ *  config set by getConfig() at init per localization
+ *********************************************************/
+let config = {}
```
kzar, line 14: Could we have these initialised to null?

src/features/click-to-play.js
```diff
@@ -1,13 +1,19 @@
...
+import { styles, getConfig } from './click-to-play/ctl-config.js'
 
 let devMode = false
 let isYoutubePreviewsEnabled = false
 let appID
 
 const titleID = 'DuckDuckGoPrivacyEssentialsCTLElementTitle'
 
+/*********************************************************
+ *  config set by getConfig() at init per localization
```
kzar, line 12:
```suggestion
// Configuration for how the placeholder elements should look and behave.
// @see {getConfig}
```

unit-test/dirSize.js
```diff
@@ -0,0 +1,16 @@
+import fastFolderSizeSync from 'fast-folder-size/sync.js'
+
+const buildDirs = ['android', 'chrome', 'chrome-mv3', 'firefox', 'integration', 'windows']
+const quotas = [512000, 680960, 512000, 512000, 512000, 512000]
```
kzar, line 4: Could you add a comment to explain the idea of this and why you picked those values? I know we talked about 500kb in Asana, but to someone reading this code it might not be obvious what unit the numbers are in for example.

scripts/buildLocales.js
```diff
@@ -0,0 +1,31 @@
...
+// This script loads the current JSON-based locales files and merges them into
+// a single importable ES module for bundling purposes
+const localesRoot = process.argv[2]
+if (!localesRoot) {
+    console.log('usage: buildLocales <locales/feature dir>')
+}
+
+const locales = {}
+const localesDirs = fs.readdirSync(localesRoot).filter(f => !f.startsWith('.'))
+for (const l of localesDirs) {
```
kzar, line 12: Nit:
```suggestion
for (const locale of localesDirs) {
```

package.json
```diff
@@ -5,7 +5,8 @@
     "postinstall": "npm run copy-sjcl",
     "copy-sjcl": "node scripts/generateSJCL.js",
     "bundle-config": "node scripts/bundleConfig.mjs",
-    "build": "npm run build-firefox && npm run build-chrome && npm run build-apple && npm run build-android && npm run build-windows && npm run build-integration && npm run build-chrome-mv3",
+    "build": "npm run build-locales && npm run build-firefox && npm run build-chrome && npm run build-apple && npm run build-android && npm run build-windows && npm run build-integration && npm run build-chrome-mv3",
+    "build-locales": "mkdir -p build/locales && node scripts/buildLocales src/locales/click-to-play > build/locales/ctl-locales.js",
```
kzar, line 9: I figure this command shouldn't need to hard-code "click-to-play". Couldn't it automatically generate the file for each feature that has translated strings?

scripts/buildLocales.js
ladamski: Good call!

src/features/click-to-play.js
ladamski: Sure tho we are mixing initializers (the next two are empty array & obj)

scripts/buildLocales.js
ladamski: I actually had that originally but I've seen some odd typo bugs due to stuff like 'file of files' .  But NBD, change made. :)

package.json
ladamski: That seems a bit like premature optimization, since I don't know the file / directory structure of upcoming features.  Its easy enough to append another '&& buildLocales' for now, and revise if it becomes an issue?

scripts/buildLocales.js
ladamski: The intent here is not to prevent code injection, but to support the legitimate use of backticks in strings.

My understanding is that using element properties is safe (vs directly in content).  If we want to defense it depth we could go back to using the template literals but that was overkill on my part I think.

package.json
kzar: Well, up to you. I just figure there's no need to hard code the names here, we could have the script list the directories in src/locales and produce a bundle for each one. E.g. since src/locales/click-to-play exists, we produce build/locales/click-to-play-locales.js

scripts/buildLocales.js
kzar: Fair enough, but would you mind taking a quick look though @jonathanKingston? Happy to go with this if you think it's safe.

jonathanKingston: I think this is probably 'good enough' for the reasons Lucas gives; this comes from a known source and we're trying to prevent legitimate use cases and data from breaking the script.

Escaping is hard and I'm sure there's something we're missing here but there's not a 'threat actor'.

If we wish to defend against this, writing a super simple test that imports the file and reads some properties from the translation would ensure the escaping isn't breaking once released.

kzar: OK cool, if you're happy then it sounds good to me :+1:

kzar (approved):

kzar (approved):

PR #255, opened 2023-02-13
Workaround inconsistent entity name for Facebook

kzar:
The extension configuration and block list have inconsistent entity
names for Facebook. The extension configuration uses "Facebook",
whereas the block list uses "Facebook, Inc.". We'll fix that in the
extension configuration soon, but in the mean time let's ensure we
don't display the Click to Load placeholders for first-party Facebook
content!


jonathanKingston:
Is the entity in the blocklist or just privacy-configuration?
We could easily generate the data for both into the config output (rather than manually maintain two entries in the config)

kzar:
> Is the entity in the blocklist or just privacy-configuration? We could easily generate the data for both into the config output (rather than manually maintain two entries in the config)

The entity name is in the the config bundled in this repository, the extension config and the block list.

We already work around the "Facebook" vs "Facebook, Inc." inconsistency in the extension code, so we won't accidentally block first-party Facebook requests. This PR is just to add the same workaround to the content script code, so we don't accidentally show placeholders for first-party Facebook content either.

ladamski (approved):
LGTM

PR #256, opened 2023-02-14
Runtime script checks

jonathanKingston:
Asana: https://app.asana.com/0/72649045549333/1203962730315600/f

Overall the implementation takes constructed <script> elements into the page and turns them into <ddg-runtime-checks> we do this because it creates a benign element that doesn't inject code.
- We use a custom element as it allows for a whole custom interface to the element.
- Additionally it allows us to capture when the node is injected into the DOM with connectedCallback.
- Once the node is injected into the DOM, we transpose all the listeners, props and attrs into a real <script>
  - In the second phase of this implementation, we'll be modifying the script code / properties prior to injection.


src/features/runtime-checks.js
```diff
@@ -0,0 +1,249 @@
...
+    customElements.define('ddg-runtime-checks', DDGRuntimeChecks)
+}
+
+export function init (args) {
+    const domain = args.site.domain
+    const domains = getFeatureSetting(featureName, args, 'domains') || []
+    const enabled = domains.find((rule) => {
+        if (rule.matchAll) {
+            return true
+        }
```
sammacbeth, line 237: Under the `domains` setting, there is a list of `rule`s. Isn't this the wrong way around?

Also, if one rule contains `matchAll: true`, then it will nullify all the rest of the rules, right? Why not make that a top-level setting?

src/features/runtime-checks.js
```diff
@@ -0,0 +1,249 @@
...
+        return matchHostname(domain, rule.domain)
```
sammacbeth, line 238: How many `rules`/`domains` do we expect to have in this list?

src/features/runtime-checks.js
```diff
@@ -0,0 +1,249 @@
...
+            }
+            return Reflect.apply(fn, scope, args)
+        }
+    })
+    proxy.overload()
+    initialCreateElement = proxy._native
+}
+
+export function load () {
+    customElements.define('ddg-runtime-checks', DDGRuntimeChecks)
```
sammacbeth, line 228: This [can throw](https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/define), for example if the extension reloads and reruns this code, or if there is a conflict with existing custom elements in the page. Should we catch and handle that error here?

src/features/runtime-checks.js
```diff
@@ -0,0 +1,249 @@
...
+        if (el) {
+            return el.removeEventListener(...args)
+        }
+        this.#listeners = this.#listeners.filter((listener) => {
+            return listener[0] !== args[0] || listener[1] !== args[1]
+        })
+    }
+
+    toString () {
+        // TODO
```
sammacbeth, line 139: Is there more to do here? If so, maybe add a description of what this should do.

src/features/runtime-checks.js
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * Overrides the instanceof checks to make the custom element interface pass an instanceof check
+ * @param {Object} elementInterface
+ */
+function overloadInstanceOfChecks (elementInterface) {
+    const proxy = new Proxy(elementInterface[Symbol.hasInstance], {
+        apply (fn, scope, args) {
+            if (args[0] instanceof DDGRuntimeChecks) {
+                return true
+            }
```
sammacbeth, line 181: Does this make `x instanceof A` always true if x is a `DDGRuntimeChecks`? Why do we need to do that?

sammacbeth (commented):
Added some comments and questions.

src/features/runtime-checks.js
jonathanKingston: Fixed by translating the first letter to upper case, there will be some edge cases that might be wrong but that's fine for now as we're only looking to support script for now.

src/features/runtime-checks.js
jonathanKingston: so that ddg-runtime-check behaves like script.

This means if the page checks this; it pretends to be a script (same as tagname etc) so they should continue interacting with it.

src/features/runtime-checks.js
jonathanKingston: We don't have any idempotency guarantees for C-S-S (in fact in Apple code we have to do a grubby little hack to prevent it from triggering twice).

Additionally the page shouldn't run first (as you well know this isn't always the case); so yeah I'll catch this but it "shouldn't ever happen™️"

src/features/runtime-checks.js
jonathanKingston: The structure is:

```
    "settings": {
        "domains": [
            {
                "domain": "mozilla.org"
            }
        ]
    }
```

This is to allow progressive rollout of the feature, I take your point I could make a `matchAllDomains` key though.

The intention here is to delete this after initial rollout.

src/features/runtime-checks.js
jonathanKingston: Max 100 before rolling out to all domains.

shakyShane:
@jonathanKingston and I have spoken about this, we decided on the following

1. We should add tests for a common pattern where a listener is added just before the DOM node is appended to the document, for example.

```javascript
const script = document.createElement("script")
script.onload = () => console.log("onload")
//                                 ^ does this fire as expected
```

2. We also discussed about the need to verify that something like a React App that controls a script tag doesn't become 'broken' if state changes and the DOM tree no longer matches

shakyShane (approved):

PR #258, opened 2023-02-15
Check only first match and nest under rule.domain

jonathanKingston:


src/features/runtime-checks.js
```diff
@@ -214,8 +214,8 @@ function shouldInterrogate (tagName) {
     }
     const stack = getStack()
     const scriptOrigins = [...getStackTraceOrigins(stack)]
-    const isInterestingHost = scriptOrigins.some(origin => {
-        return stackDomains.find(hostname => matchHostname(origin, hostname))
+    const isInterestingHost = scriptOrigins.find(origin => {
```
shakyShane, line 217:
```suggestion
    const interestingHost = scriptOrigins.find(origin => {
```

src/features/runtime-checks.js
```diff
@@ -214,8 +214,8 @@ function shouldInterrogate (tagName) {
...
+        return stackDomains.find(rule => matchHostname(origin, rule.domain))
     })
     return isInterestingHost
```
shakyShane, line 220:
```suggestion
    return interestingHost
```

shakyShane (changes requested):

shakyShane (approved):

PR #261, opened 2023-02-17
Runtime check fix tag name

jonathanKingston:
I guess I'll add this to my TOFIX list, ensure we add TS to all files!


shakyShane (dismissed):

shakyShane:
yeah, I'd remove the opt-ins, and per-file opt out instead, because that way new files are checked by default

shakyShane (approved):

PR #267, opened 2023-02-22
TSC by default for new files (second attempt)

shakyShane:
Suppressing existing errors, which allows us to ensure that all NEW files are checked by default.

https://app.asana.com/0/1201614831475344/1203979574128023/f


jonathanKingston (approved):
Cheers for this, I rebased and removed the build files only.

PR #287, opened 2023-02-27
Display Click to Load placeholders on demand

kzar:
Historically, the Click to Load placeholders were only drawn once at
the point page load completed. That had two problems:
1. Sometimes placeholders would take too long to display, if the page
   took a long time to load fully.
2. Sometimes placeholders would not be displayed at all, if the
   blocked content was only created after page load had finished.

To fix that, let's display the placeholders when instructed by the
platform using a "displayClickToLoadPlaceholders" update message.

Let's also restructure the response messages, so that they can be
clearly distinguished.

### DIFF
--- a/inject/chrome.js
+++ b/inject/chrome.js
@@ -126,16 +126,25 @@ function init () {
         }
     })
 
-    window.addEventListener('sendMessageProxy' + messageSecret, (m) => {
-        // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
-        const messageType = m.detail.messageType
+    window.addEventListener('sendMessageProxy' + messageSecret, event => {
+        event.stopImmediatePropagation()
+
+        if (!(event instanceof CustomEvent) || !event?.detail) {
+            return console.warn('no details in sendMessage proxy', event)
+        }
+
+        const messageType = event.detail?.messageType
         if (!allowedMessages.includes(messageType)) {
             return console.warn('Ignoring invalid sendMessage messageType', messageType)
         }
-        // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
-        chrome.runtime.sendMessage(m && m.detail, response => {
-            const msg = { func: messageType, response }
-            const stringifiedArgs = JSON.stringify({ detail: msg })
+
+        chrome.runtime.sendMessage(event.detail, response => {
+            const message = {
+                messageType: 'response',
+                responseMessageType: messageType,
+                response
+            }
+            const stringifiedArgs = JSON.stringify(message)
             const callRandomUpdateFunction = `
                 window.${reusableMethodName}('${reusableSecret}', ${stringifiedArgs});
             `
--- a/src/features/click-to-play.js
+++ b/src/features/click-to-play.js
@@ -20,6 +20,9 @@ let sharedStrings = null
 const entities = []
 const entityData = {}
 
+let readyResolver
+const ready = new Promise(resolve => { readyResolver = resolve })
+
 /*********************************************************
  *  Widget Replacement logic
  *********************************************************/
@@ -364,22 +367,6 @@ class DuckWidget {
     }
 }
 
-/**
- * Initialise the Click to Load feature, once the necessary details have been
- * returned by the platform.
- * @returns {Promise}
- */
-async function initCTL () {
-    await replaceClickToLoadElements()
-
-    window.addEventListener('ddg-ctp-replace-element', ({ target }) => {
-        replaceClickToLoadElements(target)
-    }, { capture: true })
-
-    // Inform surrogate scripts that CTP is ready
-    originalWindowDispatchEvent(createCustomEvent('ddg-ctp-ready'))
-}
-
 function replaceTrackingElement (widget, trackingElement, placeholderElement, hideTrackingElement = false, currentPlaceholder = null) {
     widget.dispatchEvent(trackingElement, 'ddg-ctp-tracking-element')
 
@@ -492,11 +479,14 @@ async function replaceYouTubeCTL (trackingElement, widget, togglePlaceholder = f
     }
 
     // Show YouTube Preview for embedded video
+    // TODO: Fix the hideTrackingElement option and reenable, or remove it. It's
+    //       disabled for YouTube videos so far since it caused multiple
+    //       placeholders to be displayed on the page.
     if (isYoutubePreviewsEnabled === true) {
         const { youTubePreview, shadowRoot } = await createYouTubePreview(trackingElement, widget)
         const currentPlaceholder = togglePlaceholder ? document.getElementById(`yt-ctl-dialog-${widget.widgetID}`) : null
         replaceTrackingElement(
-            widget, trackingElement, youTubePreview, /* hideTrackingElement= */ true, currentPlaceholder
+            widget, trackingElement, youTubePreview, /* hideTrackingElement= */ false, currentPlaceholder
         )
         showExtraUnblockIfShortPlaceholder(shadowRoot, youTubePreview)
 
@@ -506,7 +496,7 @@ async function replaceYouTubeCTL (trackingElement, widget, togglePlaceholder = f
         const { blockingDialog, shadowRoot } = await createYouTubeBlockingDialog(trackingElement, widget)
         const currentPlaceholder = togglePlaceholder ? document.getElementById(`yt-ctl-preview-${widget.widgetID}`) : null
         replaceTrackingElement(
-            widget, trackingElement, blockingDialog, /* hideTrackingElement= */ true, currentPlaceholder
+            widget, trackingElement, blockingDialog, /* hideTrackingElement= */ false, currentPlaceholder
         )
         showExtraUnblockIfShortPlaceholder(shadowRoot, blockingDialog)
     }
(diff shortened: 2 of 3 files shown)


src/features/click-to-play.js
```diff
@@ -20,6 +20,9 @@ let sharedStrings = null
 const entities = []
 const entityData = {}
 
+let readyResolver
+const ready = new Promise(resolve => { readyResolver = resolve })
```
kzar, line 24: (This is used so that we can block replacing placeholders only until the feature has finished starting up.)

src/features/click-to-play.js
```diff
@@ -1346,6 +1336,8 @@ const updateHandlers = {
     }
 }
 
+const knownMessage = Object.prototype.hasOwnProperty.bind(messageResponseHandlers)
```
kzar, line 1339: (This is to avoid functions inherited from Objects.prototype being called.)

src/features/click-to-play.js
```diff
@@ -1411,17 +1403,25 @@ export function init (args) {
...
-    if (typeof fn !== 'function') { return }
+    // Message response.
+    if (typeof args?.detail?.func === 'string' && knownMessage(args.detail.func)) {
+        return messageResponseHandlers[args.detail.func](args.detail.response)
+    }
 
-    fn(detail.response)
+    // Other known update messages.
+    if (args?.detail?.type === 'displayClickToLoadPlaceholders') {
+        // TODO: Pass `args.detail.ruleAction` through, that way only content
```
kzar, line 1422: I actually think this is OK for now, but I thought it made sense to pass the rule action through ready for if we ever decide to do this. Otherwise, changing the message on all platforms would be a pain.

src/features/click-to-play.js
```diff
@@ -364,22 +367,6 @@ class DuckWidget {
     }
 }
 
-/**
- * Initialise the Click to Load feature, once the necessary details have been
- * returned by the platform.
- * @returns {Promise}
- */
-async function initCTL () {
```
kzar, line 372: Having both `init` and `initCTL` was confusing. Especially with these changes, where `initCTL` needed to do even less work. I thought it would be better to remove this therefore, but I'm still hoping to move more of the logic into `init` in the future.

inject/android.js
```diff
@@ -8,11 +8,11 @@ function init () {
     if (isGloballyDisabled(processedConfig)) {
         return
     }
-    // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
```
kzar, line 11: I fixed a TypeScript issue for the files I'm changing, but that caused TypeScript to complain that these unrelated `@ts-expect-error`s were no longer needed... so that's why I had to remove them all!

inject/chrome.js
```diff
@@ -126,16 +126,25 @@ function init () {
         }
     })
 
-    window.addEventListener('sendMessageProxy' + messageSecret, (m) => {
-        // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
-        const messageType = m.detail.messageType
+    window.addEventListener('sendMessageProxy' + messageSecret, event => {
+        event.stopImmediatePropagation()
+
+        if (!(event instanceof CustomEvent) || !event?.detail) {
```
kzar, line 132: (The `CustomEvent` check is to keep TypeScript happy...)

src/features/click-to-play.js
```diff
@@ -1346,6 +1336,8 @@ const updateHandlers = {
     }
 }
 
+const knownMessageResponseType = Object.prototype.hasOwnProperty.bind(messageResponseHandlers)
```
kzar, line 1339: This is to avoid response messages being able to trigger functions that were inherited from Object.prototype to be called.

src/features/click-to-play.js
```diff
@@ -1411,17 +1403,32 @@ export function init (args) {
...
+    if (messageType === 'response') {
+        const messageResponseType = message?.responseMessageType
+        if (messageResponseType && knownMessageResponseType(messageResponseType)) {
+            return messageResponseHandlers[messageResponseType](message.response)
+        }
+    }
+
+    // Other known update messages.
+    if (messageType === 'displayClickToLoadPlaceholders') {
+        // TODO: Pass `message.options.ruleAction` through, that way only
```
kzar, line 1429: I actually think this is OK for now, but I thought it made sense to pass the rule action through ready for if we ever decide to do this. Otherwise, changing the message on all platforms would be a pain.

src/globals.d.ts
```diff
@@ -3,3 +3,8 @@ declare function exportFunction(fn: Function, desc: object, out: object): void;
 declare function exportFunction(fn: Function, desc: object): void;
 declare function cloneInto(fn: object, desc: object, out: object): void;
 declare function cloneInto(fn: object, desc: object): void;
+declare namespace  contentScopeFeatures {
```
kzar, line 6: This is just to keep TypeScript happy, since it turns out the `/* global ... */` comment only works for ESLint.

src/features/click-to-play.js
```diff
@@ -1411,17 +1403,32 @@ export function init (args) {
...
     sendMessage('getClickToLoadState')
 }
 
-export function update (args) {
-    const detail = args && args.detail
-    if (!(detail && detail.func)) { return }
+export function update (message) {
+    // TODO: Once all Click to Load messages include the feature property, drop
+    //       messages that don't include the feature property too.
+    if (message?.feature && message?.feature !== 'clickToLoad') return
```
kzar, line 1414: Update "messages" are sent to all enabled features, so messages of one feature could clash with another. I made a start adding a feature property here, but in the future we should ensure it's sent more consistently.

kzar: Yea, if the content-scope-scripts abstractions takes care of this in the future then it sounds good to me. I just thought I'd keep the changes for this PR minimal, while making the Click to Load messages make sense (since ports are in progress).

jonathanKingston: It's only on the extension platforms currently and you're changing out how all the messaging is working.

Maybe I'm misunderstanding the complexity here but we can probably move this if statement up a level to only trigger to the correct feature in this PR safely. (I've not looked at update in a while so might be totally misremembering)

Anyway I think it's serious enough to scope as a task before we end up with a runtime issues overtriggering etc.

kzar: We can't move the feature check up a level yet unfortunately, since message responses don't include the "feature" property. I started looking at including feature in message resposnes, but then the original message needed to include the "feature" property too, and to add those we needed to change all Click to Load messages to use Objects for the message payload so the "feature" could be included. Like I say, with this PR I'm trying to keep the changes minimal while making the API make sense for folks porting Click to Load to other platforms.

src/features/click-to-play.js
```diff
@@ -1411,17 +1403,32 @@ export function init (args) {
...
+    // Message responses.
+    if (messageType === 'response') {
+        const messageResponseType = message?.responseMessageType
+        if (messageResponseType && knownMessageResponseType(messageResponseType)) {
+            return messageResponseHandlers[messageResponseType](message.response)
+        }
+    }
+
+    // Other known update messages.
+    if (messageType === 'displayClickToLoadPlaceholders') {
```
ladamski, line 1428: The issue with this approach is that it will cause us to re-enumerate the DOM for each tracker that we block at https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1738/files#diff-ed5e7ba3132fdd67f6264c179d3641e8959578d350a29280c36f58ab228c52f5R260, which can result in a lot of redundant enumeration.  For https://privacy-test-pages.glitch.me/privacy-protections/click-to-load/ we do 8 passes when 1 would do.

A mitigation might be to do a set a semaphore that a refresh is needed (f not already set), and then do a setTimeout for like 10ms, fire `replaceClickToLoadElements()` and clear the semaphore.  That way we'd enumerate likely once (or every 10ms at most) vs repeatedly for a number of blocked elements at load.

kzar: Yea, the test pages are pretty much the worst case for this approach. A bunch of blocked elements, on one page, that were all ready before page load, so the event fires multiple times but there are not more elements found each time. Usually though, this new behaviour is what we want. We don't want to wait for page load to display placeholders, nor assume that because we didn't find any blocked elements in the past, that there aren't any in the future.

I agree with your suggested mitigation, if performance proves to be an issue I'd be up for doing something like that to reduce the number of refreshes that happen in quick succession. I'd prefer not to add that for now though, I figure there's a good chance it's a case of premature optimisation. (I would also take more care to use the provided ruleAction at that point.) What do you think?

ladamski: For FB perhaps but for YT pages of videos aren't that uncommon.  Take https://www.surfingpaddling.com/dropping-in-surf-show for example.  Granted its already a slow page to load but with YT protections enabled it basically ceases to render (for me, in Chrome).

kzar: _(As discussed in MatterMost, it turned out the (broken) `hideTrackingElement` option was causing an exponential number of placeholder elements to be added. Which caused websites like that with a large number of videos to be slow or crash. I have disabled that option for now, since I have a follow-up task to remove/fix it already anyway.)_

ladamski: I spent some time profile this with and without a setTimeout hack, and didn't notice much difference so I think we're good to go here!

ladamski:
Ok I'm having one issue but its annoying to repro.  Basically after the browser has been open for a couple hours CTL is no longer being applied to any page (whether I reload, open a new tab, etc).  Maybe the background script is getting killed and not restarting?

kzar:
> Ok I'm having one issue but its annoying to repro. Basically after the browser has been open for a while (couple hours) CTL is no longer being applied to any page (whether I reload, open a new tab, etc). Maybe the background script is getting killed and not restarting?

Weird. OK I have [filed a task in Asana](https://app.asana.com/0/0/1204105573463789/f) to investigate that.

PR #295, opened 2023-03-06
Add safari-mv3 platform

sammacbeth:
This is basically a copy of the `chrome-mv3` entrypoint, but we spam the `ddg-secret` event until we see a response. This is required because the content-script loading order on Safari is non-deterministic, so, in most cases, the initial event on load will not be seen by the isolated content-script.


package.json
```diff
@@ -10,11 +10,12 @@
     "postinstall": "npm run copy-sjcl",
     "copy-sjcl": "node scripts/generateSJCL.js",
     "bundle-config": "node scripts/bundleConfig.mjs",
-    "build": "npm run build-locales && npm run build-firefox && npm run build-chrome && npm run build-apple && npm run build-android && npm run build-windows && npm run build-integration && npm run build-chrome-mv3",
+    "build": "npm run build-locales && npm run build-firefox && npm run build-chrome && npm run build-apple && npm run build-android && npm run build-windows && npm run build-integration && npm run build-chrome-mv3 && npm run build-safari-mv3",
```
jonathanKingston, line 13: 😢

jonathanKingston: One for a refactor task at some point.

PR #296, opened 2023-03-06
Reland Display Click to Load placeholders on demand

kzar:
These changes were landed and reverted last week, since some of the
integration tests started failing. But it turned out, that those failures
were just down to Puppeteer reporting blocked requests differently if
the corresponding elements are removed from the DOM very quickly.
See https://github.com/duckduckgo/duckduckgo-privacy-extension/commit/8b5d3898cdb09676c887b6d933121e7b748bd341

The original commit message:

Historically, the Click to Load placeholders were only drawn once at
the point page load completed. That had two problems:
1. Sometimes placeholders would take too long to display, if the page
   took a long time to load fully.
2. Sometimes placeholders would not be displayed at all, if the
   blocked content was only created after page load had finished.

To fix that, let's display the placeholders when instructed by the
platform using a "displayClickToLoadPlaceholders" update message.

Let's also restructure the response messages, so that they can be
clearly distinguished.

### DIFF
--- a/inject/chrome.js
+++ b/inject/chrome.js
@@ -126,16 +126,25 @@ function init () {
         }
     })
 
-    window.addEventListener('sendMessageProxy' + messageSecret, (m) => {
-        // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
-        const messageType = m.detail.messageType
+    window.addEventListener('sendMessageProxy' + messageSecret, event => {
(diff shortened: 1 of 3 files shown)


ladamski (approved):
Works for me, though given the FB CTL tests seem to be timing out for me pretty much 100% of the time we should actually try to fix those soon (perhaps by switching to local hosting)

PR #308, opened 2023-03-10
Overload Document.prototype createElement instead of document

jonathanKingston:


integration-test/test-runtime-checks.js
```diff
@@ -385,4 +385,51 @@ describe('Runtime checks: should allow element modification', () => {
...
+        const scriptResult = await page.evaluate(
+            () => {
+                // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
+                window.scriptDocumentPrototypeRan = false
+                const scriptElement = Document.prototype.createElement.call(window.document, 'script')
+                scriptElement.innerText = 'window.scriptDocumentPrototypeRan = true'
+                scriptElement.id = 'scriptDocumentPrototype'
+                scriptElement.setAttribute('type', 'application/javascript')
+                document.body.appendChild(scriptElement)
+                const hadInspectorNode = !!document.querySelector('ddg-runtime-checks')
```
muodov, line 415: Would this be better?
```suggestion
                const hadInspectorNode = scriptElement === document.querySelector('ddg-runtime-checks')
```

jonathanKingston: Possibly, I mean ideally nothing makes this detectable haha.

Let me make this as a separate PR as we're doing it everywhere else in the file.

PR #319, opened 2023-03-14
Improve element hiding empty check to avoid potentially hiding visible content

dharb:
Asana link: https://app.asana.com/0/1203557590179903/1204090946699947/f
Corresponding privacy test page update: https://github.com/duckduckgo/privacy-test-pages/pull/130

While we try to be precise in the selector rules we use for element hiding, we've run into a few cases now where we accidentally hid visible content that matched an ad selector. This PR adds several guardrails to prevent hiding visible content even when elements match common ad selectors.

Specifically, this PR:
- avoids hiding elements that contain various media tags such as `<embed>`, `<object>`, `<audio>`, and `<map>`
- avoids hiding elements that contain form tags such as `<form>`, `<input>`, `<textarea>`, `<select>`, `<option>`, and `<button>`
- avoids hiding elements that contain real images - we previously weren't counting images as visible content since ad containers frequently include tracking pixels and small ad logos (like adchoices). Now we the size of contained images and don't hide when they're > 20px width or height.

### DIFF
--- a/src/features/element-hiding.js
+++ b/src/features/element-hiding.js
@@ -137,17 +137,27 @@ function isDomNodeEmpty (node) {
     })
 
     const visibleText = parsedNode.innerText.trim().toLocaleLowerCase().replace(/:$/, '')
-    const mediaContent = parsedNode.querySelector('video,canvas,picture')
+    const mediaContent = parsedNode.querySelector('video,canvas,embed,object,audio,map')
+    const formContent = parsedNode.querySelector('form,input,textarea,select,option,button')
     const frameElements = [...parsedNode.querySelectorAll('iframe')]
+    // query original node instead of parsedNode for img elements since heuristic relies
+    // on size of image elements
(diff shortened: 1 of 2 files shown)


src/features/element-hiding.js
```diff
@@ -137,17 +137,27 @@ function isDomNodeEmpty (node) {
     })
 
     const visibleText = parsedNode.innerText.trim().toLocaleLowerCase().replace(/:$/, '')
-    const mediaContent = parsedNode.querySelector('video,canvas,picture')
+    const mediaContent = parsedNode.querySelector('video,canvas,embed,object,audio,map')
+    const formContent = parsedNode.querySelector('form,input,textarea,select,option,button')
```
jonathanKingston, line 141: Does it make sense to combine mediaContent and formContent as a single selector that we pass through from the config.

jonathanKingston (changes requested):
Can we make the selectors configurable?

jonathanKingston (approved):

PR #321, opened 2023-03-15
Replace private method support with _ for now

jonathanKingston:
This fixes a syntax issue with Android that we're having.


joshliebe (approved):
LGTM

PR #324, opened 2023-03-17
Fix toString DDGProxy response

jonathanKingston:
Fixing the toString of any use of DDGProxy.

https://app.asana.com/0/1200971092846006/1204199159492456/f

This resolves differently for Firefox and Chrome.

### DIFF
--- a/integration-test/test-utils.js
+++ b/integration-test/test-utils.js
@@ -1,0 +1,36 @@
+/**
+ *  Tests for utils
+ */
+import { setup } from './helpers/harness.js'
+
+describe('Ensure utils behave as expected', () => {
+    let browser
+    let server
+    let teardown
+    let setupServer
+    let gotoAndWait
+    beforeAll(async () => {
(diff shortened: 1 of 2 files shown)


integration-test/test-utils.js
```diff
@@ -0,0 +1,36 @@
...
+    beforeAll(async () => {
+        ({ browser, setupServer, teardown, gotoAndWait } = await setup({ withExtension: true }))
+        server = setupServer()
+    })
+    afterAll(async () => {
+        await server?.close()
+        await teardown()
+    })
+
+    fit('should toString DDGProxy correctly', async () => {
```
sammacbeth, line 21:
```suggestion
    it('should toString DDGProxy correctly', async () => {
```

jonathanKingston:
> Looks good, except the `fit` call.

Lol I knew that would get me one day! Thanks for noticing.

sammacbeth (approved):

PR #344, opened 2023-03-25
Remove the need for the build replacements for all non chrome platforms

jonathanKingston:
Replace the need for string replacement of the code for all platforms except MV2 Chrome.


shakyShane (approved):
perfect

PR #345, opened 2023-03-25
Move to using a class for consistency improvements

jonathanKingston:
Make features a class, this allows us to migrate args to being a type safe simple to use API instead.

Ignoring whitespace changes here is advisable here when viewing the PR, the number of changes is pretty minimal concentrating around content-scope-features.js and the content-features.js file.

https://app.asana.com/0/1201614831475344/1204257980027290/f

### DIFF
--- a/src/content-feature.js
+++ b/src/content-feature.js
@@ -1,0 +1,32 @@
+import { getFeatureSetting, getFeatureSettingEnabled, matchHostname } from './utils'
+
+export default class ContentFeature {
+    constructor (featureName) {
+        this.name = featureName
+        this._args = null
+    }
+
+    getFeatureSetting (featureKeyName) {
+        return getFeatureSetting(this.name, this._args, featureKeyName)
+    }
+
+    getFeatureSettingEnabled (featureKeyName) {
+        return getFeatureSettingEnabled(this.name, this._args, featureKeyName)
+    }
+
+    matchDomainFeatureSetting (featureKeyName) {
+        const domains = this.getFeatureSetting(featureKeyName)
+        return domains.find((rule) => {
+            return matchHostname(this._args.site.domain, rule.domain)
+        })
+    }
+
+    load () {
+    }
+
+    init () {
+    }
+
+    update () {
+    }
+}
--- a/src/content-feature.js
+++ b/src/content-feature.js
@@ -6,6 +6,17 @@ export default class ContentFeature {
         this._args = null
     }
 
+    /**
+     * @param {import('./utils').Platform} platform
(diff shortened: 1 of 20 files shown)


src/features/click-to-load.js
```diff
@@ -1348,92 +1349,94 @@ const messageResponseHandlers = {
...
 const knownMessageResponseType = Object.prototype.hasOwnProperty.bind(messageResponseHandlers)
 
-export function init (args) {
-    const websiteOwner = args?.site?.parentEntity
-    const settings = args?.featureSettings?.clickToLoad || {}
-    const locale = args?.locale || 'en'
-    const localizedConfig = getConfig(locale)
-    config = localizedConfig.config
-    sharedStrings = localizedConfig.sharedStrings
+export default class DuckPlayer extends ContentFeature {
```
shakyShane, line 1352: Wrong name here? 

```suggestion
export default class ClickToLoadFeature extends ContentFeature {
```

src/content-feature.js
```diff
@@ -0,0 +1,59 @@
...
+     */
+    set platform (platform) {
+        this._platform = platform
+    }
+
+    get platform () {
+        return this._platform
+    }
+
+    getFeatureAttr (attrName, defaultValue) {
```
shakyShane, line 20: I think you already mentioned coming back and adding types to these in a follow up?

Assuming they were untyped before too? if so, fine by me to reduce the size of the initial change 💪🏻

src/utils.js
```diff
@@ -486,26 +500,28 @@ export function processConfig (data, userList, preferences, platformSpecificFeat
         return feature.state === 'enabled' && !isUnprotectedDomain(topLevelHostname, feature.exceptions)
     }).concat(platformSpecificFeaturesNotInRemoteConfig) // only disable platform specific features if it's explicitly disabled in remote config
     const isBroken = isUnprotectedDomain(topLevelHostname, data.unprotectedTemporary)
-    preferences.site = {
+    /** @type {Record<string, any>} */
+    const output = Object.assign({}, preferences)
```
shakyShane, line 504: not a blocker of course, but I wouldn't default to `Object.assign({}, foo)` for this case - just because it's a mutating operation on the first arg and easy to get wrong.  

I'd do this instead:

```js
const output = { ...preferences } 
```

shakyShane (changes requested):
@jonathanKingston Looks great - definitely an improvement already, and can see it getting even better shortly too as we plumb in things such as shared messaging etc :)

Just left a couple of tiny comments

PR #346, opened 2023-03-25
tests: updated the unit tests to check file size only

shakyShane:


shakyShane:
* main
  * **PR #346** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/346" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/346?utm_source=stack-comment).

unit-test/dirSize.js
```diff
@@ -1,19 +1,45 @@
...
-describe('Expect build size of', () => {
-    for (const build in buildDirs) {
-        const dir = `${rootDir}/${buildDirs[build]}`
-        const sizeLimit = quotas[build]
+const checks = {
+    android: [
+        { kind: 'maxFileSize', value: 512000, path: join(BUILD, 'android/contentScope.js') }
+    ],
+    chrome: [
+        { kind: 'maxFileSize', value: 716800, path: join(BUILD, 'chrome/inject.js') }
```
jonathanKingston, line 14: We should preserve the previous comment, I'd prefer we make the whole file simpler not more complex unless you're foreseeing using this for the HTML too?

shakyShane: yep it would be good to have a way to add more checks.

if you think this is 'too complex', we can revert and add the other checks else where 👍🏻

shakyShane: The old version:

1: Only works for a directory, which no longer makes sense
2: didn’t support apple
3: used a 3rd party lib

so yes, it’s now more complex to solve those

jonathanKingston: Nice missed that:

- Please remove the other lib :)
- Can we use a single int for the output size and make a 40% difference calculation for Chrome?

> if you think this is 'too complex', we can revert and add the other checks else where 👍🏻

Nah, if there's intent to do this soon that's fine by me; I just didn't want the if statement and 'kind' complexity if there wasn't the need for it.

shakyShane: done 👍🏻

jonathanKingston (approved):

PR #349, opened 2023-03-26
Add stack trace check to cookie protection

jonathanKingston:
See: https://app.asana.com/0/0/1204257980027287/f

See also: https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1847


src/features/cookie.js
```diff
@@ -176,16 +211,24 @@ export default class CookieFeature extends ContentFeature {
 
     init (args) {
         if (args.cookie) {
+            const trackerLookup = cookiePolicy.trackerLookup
             cookiePolicy = args.cookie
+            cookiePolicy.trackerLookup = trackerLookup
```
sammacbeth, line 216: Can we not assign `trackerLookup` to a different global, rather than doing this copy?

inject/chrome.js
```diff
@@ -40,6 +40,7 @@ function randomString () {
 function init () {
     // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
     const documentOriginIsTracker = isTrackerOrigin($TRACKER_LOOKUP$)
+    const trackerLookup = JSON.stringify($TRACKER_LOOKUP$)
```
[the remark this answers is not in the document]
jonathanKingston: When we spoke about this, we kinda agreed the alternative is message passing across multiple frames with a fair amount of bloat of complexity.

Without measuring the performance (which we can now do pretty easily given the performance markers I added https://github.com/duckduckgo/content-scope-scripts/pull/396) I'd reason that it's largely tit-for-tat performance gains depending on website.

We can fast follow with changing this up without it impacting the feature itself.

jonathanKingston:
@shakyShane do you mind doing a first pass here as there's lots of C-S-S infra being changed then I can pass it back to @sammacbeth?

sammacbeth (dismissed):
Mostly looks fine to me. Just two minor comments/questions.

package.json
```diff
@@ -19,6 +19,7 @@
     "build-android": "node scripts/inject.js --platform android",
     "build-windows": "node scripts/inject.js --platform windows",
     "build-integration": "node scripts/inject.js --platform integration",
+    "bundle-trackers": "node scripts/bundleTrackers.mjs > build/tracker-lookup.json",
```
shakyShane, line 22: I don't think this works on windows - just do the write inside the `bundleTrackers.mjs` to fix

scripts/bundleTrackers.mjs
```diff
@@ -0,0 +1,25 @@
...
+    } else {
+        node[domainParts[0]] = {}
+        insert(domainParts.slice(1), node[domainParts[0]])
+    }
+}
+Object.keys(tds.trackers).forEach((tracker) => {
+    insert(tracker.split('.').reverse(), trackerLookupTrie)
+})
+
+console.log(JSON.stringify(trackerLookupTrie))
```
shakyShane, line 25: as mentioned in another comment, just write directly to disk within node here and it'll work on windows then

shakyShane:
@jonathanKingston - I just left 2 notes about Windows support, if not relevant then feel free to ignore :)

src/features/cookie.js
```diff
@@ -43,26 +53,49 @@ function shouldBlockNonTrackingCookie () {
...
+    return cookiePolicy.isFrame && !cookiePolicy.isTracker && cookiePolicy.isThirdPartyFrame
 }
 
 export default class CookieFeature extends ContentFeature {
-    load (args) {
+    load () {
         // Feature is only relevant to the extension and windows, we should skip for other platforms for now as the config testing is broken.
         if (this.platform.name !== 'extension' && this.platform.name !== 'windows') {
             return
         }
```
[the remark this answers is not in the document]
jonathanKingston: After testing can be removed yeah.

We can now also remove features from the compiled bundled now; thanks @shakyShane so we no-longer need to do these kinda hacks for unsupported code 😄

package.json
jonathanKingston: I was trying to avoid having different code paths for each platform (extension uses this). I'll think of some way to resolve.

scripts/bundleTrackers.mjs
jonathanKingston: Kept this for the extension but allowed for --output in the default case here for C-S-S

shakyShane (approved):

src/features/cookie.js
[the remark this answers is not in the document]
jonathanKingston: After testing MacOS I'm pretty convinced this is all working as expected so removed this line.

sammacbeth (approved):

jonathanKingston:
Merging this for now, there might be some follow up for Windows (which is currently broken anyway so we're not too concerned here).

PR #351, opened 2023-03-27
Abstract away method calling and call native captured prototype

jonathanKingston:
https://app.asana.com/0/0/1204243453023943/f

### DIFF
--- a/src/features/runtime-checks.js
+++ b/src/features/runtime-checks.js
@@ -24,6 +24,14 @@ let elementRemovalTimeout
 const featureName = 'runtimeChecks'
 const taintSymbol = Symbol(featureName)
 const supportedSinks = ['src']
+// Store the original methods so we can call them without any side effects
+const defaultElementMethods = {
+    setAttribute: HTMLElement.prototype.setAttribute,
+    getAttribute: HTMLElement.prototype.getAttribute,
+    removeAttribute: HTMLElement.prototype.removeAttribute,
+    remove: HTMLElement.prototype.remove,
+    removeChild: HTMLElement.prototype.removeChild
+}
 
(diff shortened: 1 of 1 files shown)


src/features/runtime-checks.js
```diff
@@ -274,19 +285,12 @@ class DDGRuntimeChecks extends HTMLElement {
...
     }
 
+    // @ts-expect-error TS node return here
     removeChild (child) {
-        const el = this._getElement()
-        if (el) {
-            return el.removeChild(child)
-        }
-        return super.removeChild(child)
+        return this._callMethod('removeChild', child)
```
jonathanKingston, line 293: @shakyShane I couldn't get the type magic to align here.

The error is:
```
Property 'removeChild' in type 'DDGRuntimeChecks' is not assignable to the same property in base type 'HTMLElement'.
  Type '(child: any) => Node' is not assignable to type '<T extends Node>(child: T) => T'.
    Type 'Node' is not assignable to type 'T'.
      'Node' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'Node'.ts(2416)
```

This also breaks the `customElements.define` method below also.

shakyShane (approved):

src/features/runtime-checks.js
jonathanKingston: Yeah I tried assigning to many other things, I think the only way is perhaps adding a T on the class itself. Let me try that just in case it's that simple.

We have tests for many of these functions, I don't think this one specifically.

jonathanKingston: Yeah didn't work, I'm just going to merge for now. The important methods like setAttribute have tests and correctly break if not working etc.

PR #357, opened 2023-03-28
Remove unused (and broken) hideTrackingElement Click to Load code path

kzar:
The hideTrackingElement code path was originally intended for use with
YouTube Click to Load placeholders. The thinking was to keep the
original YouTube video elements on the page, in case the website
referenced the original element when creating a video.

Unfortunately that didn't work out, since keeping the original element
on the page caused problems such as duplicate placeholders being
displayed. We already stopped using that code path, let's tidy things
up and remove it.

## Testing notes
- Click to Load YouTube is not fixed with this change, this is just to lay the groundwork.
- Ensure the Playwright Click to Load Facebook integration tests still pass, and that Click to Load Facebook works OK in practice.

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -316,19 +316,6 @@ class DuckWidget {
                         break
                     }
 
-                    // If hidden, restore the tracking element's styles to make
-                    // it visible again.
-                    if (this.originalElementStyle) {
-                        for (const key of ['display', 'visibility']) {
-                            const { value, priority } = this.originalElementStyle[key]
-                            if (value) {
-                                fbElement.style.setProperty(key, value, priority)
-                            } else {
-                                fbElement.style.removeProperty(key)
-                            }
-                        }
-                    }
-
                     /*
                     * Modify the overlay to include a Facebook iFrame, which
                     * starts invisible. Once loaded, fade out and remove the overlay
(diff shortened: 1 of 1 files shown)


src/features/click-to-load.js
```diff
@@ -632,44 +596,25 @@ function getLearnMoreLink (mode) {
...
-
-/**
- * Copy list of styles to provided element
- * @param {{[key: string]: string[]}} originalStyles Object with styles read from original element.
- * @param {Element} element Node element to have the styles copied to
- */
-function copyStylesTo (originalStyles, element) {
-    const { display, visibility, ...filteredStyles } = originalStyles
-    const cssText = Object.keys(filteredStyles).reduce((cssAcc, key) => (cssAcc + `${key}: ${filteredStyles[key].value};`), '')
-    element.style.cssText += cssText
```
franfaccin, line 672: I don't think we can remove the original function completely. I remember I added this to try and copy relevant CSS styles from the original element so that our placeholders sit better on the page. In particular, videos that have `position: absolute` will appear on top of empty space that the page had originally intended for the video. Since we by default use `position: relative`, our placeholder will then stack bellow this empty area. And we can't just always use `position: absolute` because not all pages are using styles like this, that's why we need to try and copy relevant styles.
Here is an example that appeared during Design Review: https://www.raspberrypi.com/news/this-robots-extra-arm-gives-it-whole-body-control/
This is how it looks now:
image


This is how it looked previously:
image

src/features/click-to-load.js
```diff
@@ -367,33 +354,11 @@ class DuckWidget {
     }
 }
 
-function replaceTrackingElement (widget, trackingElement, placeholderElement, hideTrackingElement = false, currentPlaceholder = null) {
+function replaceTrackingElement (widget, trackingElement, placeholderElement, currentPlaceholder = null) {
```
franfaccin, line 357: I'm not sure if this is caused by any of these changes, it probably isn't, but Log In with Facebook is not working for me.
- https://privacy-test-pages.glitch.me/privacy-protections/click-to-load/ : The modal appears for the first button, but then nothing happens. For the second button, it gets stuck with "Calling FB SDK..."
image

- https://en.boardgamearena.com/account?page=newuser&src=login : The custom FB button doesn't work
- https://www.babyartikel.de/en/login: Our FB button gets replaced only after clicking the second time

kzar: The first one is expected.

I'm not sure about the other two. I can reproduce, but I can also reproduce with the released extension (and before these changes.) I will open a task to investigate.

src/features/click-to-load.js
kzar: Great catch, thanks for pointing this out. I think I've sorted this now, that example renders OK for me on Chrome MV2 and MV3.

src/features/click-to-load.js
kzar: OK, yea second one is an unrelated issue that I've now pinned down. We need to start redirecting requests `https://connect.facebook.net/en_US/all.js` instead of blocking them. I'm not sure about the last website, but the Facebook login flow there seems kind of broken anyway.

franfaccin: Thank you for taking a look into this! 🙌  I'll resolve this since it's unrelated to this PR.

src/features/click-to-load.js
franfaccin: Great fix 🙌

franfaccin (approved):
I've tested again and all seems to be working well. Great job ✅

PR #360, opened 2023-03-29
Handle short Click to Load placeholders more consistently

kzar:
When Click to Load placeholders are displayed inside a short parent
element, we take care to make the placeholder shorter and display an
extra button in the header. There were two code-paths for this, let's
combine them and also take care to check that
placeholder.parentElement exists.


franfaccin (approved):
LGTM ✅

PR #364, opened 2023-03-30
Ignore monitoring for DOM0 event listeners as they're already handled

jonathanKingston:
We can't really test for this as it creates a crash in Chrome.
The following is the test I as using:
https://github.com/duckduckgo/duckduckgo-privacy-extension/commit/dc3efa29676cc5daca02d13bb32877b74e63d254

### DIFF
--- a/src/features/runtime-checks.js
+++ b/src/features/runtime-checks.js
@@ -106,6 +106,8 @@ class DDGRuntimeChecks extends HTMLElement {
                     },
                     set (value) {
                         if (shouldFilterKey(this.#tagName, 'property', prop)) return
+                        // Ignore any properties that start with 'on' as they are only relevant once and already handled
+                        if (prop.startsWith('on')) return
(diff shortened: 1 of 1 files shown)


shakyShane (dismissed):

src/features/runtime-checks.js
```diff
@@ -106,6 +106,8 @@ class DDGRuntimeChecks extends HTMLElement {
                     },
                     set (value) {
                         if (shouldFilterKey(this.#tagName, 'property', prop)) return
+                        // Ignore any properties that start with 'on' as they are only relevant once and already handled
+                        if (prop.startsWith('on')) return
```
shakyShane, line 110: just a sanity check, are _all_ `on*` props problematic, or are there any that are not to do with events? I don't think there are, but wanted to ask anyway :)

shakyShane (approved):

PR #365, opened 2023-03-31
Avoid showing placeholders for the same element multiple times

kzar:
Even without the hideTrackingElement code path, there are some
situations that can lead to duplicate placeholders from being shown
for an element. Let's explicitly prevent that from happening.


franfaccin (approved):
LGTM ✅

PR #366, opened 2023-03-31
Site specific settings features

jonathanKingston:
Allows us to remotely configure ContentScopeScript features by config.

https://app.asana.com/0/0/1204282844952067/f

### DIFF
--- a/src/content-feature.js
+++ b/src/content-feature.js
@@ -36,6 +36,18 @@ export default class ContentFeature {
      * @returns {any}
      */
     getFeatureSetting (featureKeyName) {
+        const result = this._getFeatureSettingIndividual(featureKeyName)
+        if (featureKeyName === 'domains') {
+            throw new Error('domains is a reserved feature setting key name')
+        }
+        const domainResult = this.matchDomainFeatureSetting('domains')?.settings?.[featureKeyName]
+        if (domainResult !== undefined) {
+            return domainResult
+        }
+        return result
+    }
+
+    _getFeatureSettingIndividual (featureKeyName) {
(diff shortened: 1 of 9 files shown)


src/features/runtime-checks.js
```diff
@@ -461,7 +461,7 @@ export default class RuntimeChecks extends ContentFeature {
     init () {
         let enabled = this.getFeatureSettingEnabled('matchAllDomains')
         if (!enabled) {
-            enabled = !!this.matchDomainFeatureSetting('domains')
```
jonathanKingston, line 464: I changed the return value here to be an array

src/content-feature.js
```diff
@@ -18,33 +19,64 @@ export default class ContentFeature {
...
      */
     getFeatureSetting (featureKeyName) {
-        return getFeatureSetting(this.name, this._args, featureKeyName)
+        let result = this._getFeatureSetting()
+        if (featureKeyName === 'domains') {
+            throw new Error('domains is a reserved feature setting key name')
+        }
+        const domainMatch = this.matchDomainFeatureSetting('domains').sort((a, b) => {
+            return a.domain.length - b.domain.length
+        })
```
shakyShane, line 46: how often is `getFeatureSetting` called? just wondering if there's anything perf-wise to measure/check here? might not be, but just flagging as this is likely mutating the underlying array over and over

jonathanKingston: Overall I think 10-15 times a page, it's probably ok. I think we could memoise the whole function response if it becomes an issue later.

shakyShane (approved):

PR #372, opened 2023-04-01
Runtime checks referrer policy

jonathanKingston:
https://app.asana.com/0/0/1204314451512673/f

### DIFF
--- a/integration-test/test-pages/runtime-checks/config/referrer-stripping.json
+++ b/integration-test/test-pages/runtime-checks/config/referrer-stripping.json
@@ -1,0 +1,25 @@
+{
+    "features": {
+        "runtimeChecks": {
+            "state": "enabled",
+            "exceptions": [],
+            "settings": {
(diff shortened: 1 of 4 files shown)


integration-test/test-pages/runtime-checks/config/referrer-stripping.json
```diff
@@ -0,0 +1,25 @@
...
+                "overloadInstanceOf": "enabled",
+                "injectGlobalStyles": "disabled",
+                "shadowDom": "enabled",
+                "domains": [
+                ],
+                "stackDomains": [
+                ],
+                "tagModifiers": {
+                    "script": {
+                        "referrerPolicy": "origin"
```
jonathanKingston, line 19: I think this should probably be:
```
"script": {
    "treatments": {
        "referrerPolicy": {
            "value": "origin"
        }
    }
}
```

PR #373, opened 2023-04-02
Add in exception handle debugging

jonathanKingston:


src/features/exception-handler.js
```diff
@@ -0,0 +1,19 @@
+import { postDebugMessage } from '../utils.js'
+import ContentFeature from '../content-feature.js'
+
+export default class ExceptionHandler extends ContentFeature {
+    init () {
+        // Report to the debugger panel if an uncaught exception occurs
+        function handleUncaughtException (e) {
+            postDebugMessage('error', {
```
GuiltyDolphin, line 8: I wonder whether it might be worth using a name like `jsException` in case we want `error` for something else in the future. Just a suggestion and I'm fine either way.

GuiltyDolphin (approved):

GuiltyDolphin (approved):

PR #374, opened 2023-04-02
fix: support building on windows

shakyShane:
https://app.asana.com/0/0/1204315805721508/f

**Before:** 

We had a number of things preventing building on Windows

- `mkdir`/`cp` unix-y style commands directly in `package.json`
- `import.meta.url` needs a platform check when you try to use it for absolute file paths
- `npm run a; npm run b` format is not supported on Windows

**After**

Have everything as node scripts for cross-platform support


shakyShane:
* main
  * **PR #374** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/374" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/374?utm_source=stack-comment).

package.json
```diff
@@ -11,15 +11,14 @@
...
+    "build-locales": "node scripts/buildLocales.js --dir src/locales/click-to-load --output build/locales/ctl-locales.js",
+    "build-firefox": "node scripts/inject.js --entry inject/mozilla.js --output build/firefox/inject.js --platform firefox",
+    "build-chrome": "node scripts/inject.js --entry inject/chrome.js --output build/chrome/inject.js --platform chrome",
+    "build-chrome-mv3": "node scripts/inject.js --entry inject/chrome-mv3.js --output build/chrome-mv3/inject.js --platform chrome-mv3",
+    "build-apple": "node scripts/inject.js --entry inject/apple.js --output Sources/ContentScopeScripts/dist/contentScope.js --platform apple",
+    "build-android": "node scripts/inject.js --entry inject/android.js --output build/android/contentScope.js --platform android",
+    "build-windows": "node scripts/inject.js --entry inject/windows.js --output build/windows/contentScope.js --platform windows",
+    "build-integration": "node scripts/inject.js --entry inject/integration.js --output build/integration/contentScope.js --output integration-test/extension/contentScope.js --platform integration",
```
shakyShane, line 21: The script supports multiple outputs, so no need for a separate 'copy' script 💪🏻

package.json
```diff
@@ -62,6 +61,7 @@
     "rollup": "^3.20.2",
     "rollup-plugin-dynamic-import-variables": "^1.1.0",
     "typedoc": "^0.23.28",
-    "typescript": "^5.0.3"
+    "typescript": "^5.0.3",
+    "minimist": "^1.2.8"
```
shakyShane, line 65: I was going to use the built-in arg parser in node, but it's marked as 'experimental' - so adding this single-file dependency in the mean time

scripts/generateSJCL.js
```diff
@@ -12,6 +12,11 @@ async function init () {
         return
     }
 
+    if (process.platform === 'win32') {
```
shakyShane, line 15: might need to improve this if we need this on windows - but shouldn't be a blocker here since the file is vendored anyway, so it would show up in CI if there were build updates etc

jonathanKingston: Yeah so long as we don't depend on win32 to build the code I think it's fine.

jonathanKingston: (This will break fingerprinting-canvas so if you're testing that expect it not to work well)

packages/special-pages/index.js
```diff
@@ -13,8 +13,9 @@
  */
 import { join } from 'node:path'
 import { existsSync, cpSync, rmSync } from 'node:fs'
-const CWD = new URL('.', import.meta.url).pathname
-const ROOT = new URL('../../', import.meta.url).pathname
+import { cwd } from '../../scripts/script-utils.js'
+const CWD = cwd(import.meta.url)
+const ROOT = join(CWD, '../../')
```
jonathanKingston, line 18: 😢

package.json
jonathanKingston: Nice we needed this change for sourcemaps anyway.

scripts/inject.js
```diff
@@ -1,45 +1,61 @@
...
-        initOther('inject/chrome-mv3.js', 'chrome_mv3')
-    } else {
-        initChrome()
+/**
+ * @param {string} entry
+ * @param {string} platformName
+ */
+async function bundle (entry, platformName) {
+    if (platformName === 'firefox') {
+        return initOther(entry, platformName)
```
jonathanKingston, line 13: Do we need to expose 'entry' to the build script? The risk of pointing to the wrong file is pretty catastrophic.

scripts/inject.js
```diff
@@ -1,45 +1,61 @@
...
+    return outputScript
 }
 
-init()
+// verify the input
+const requiredFields = ['entry', 'output', 'platform']
+const args = parseArgs(process.argv.slice(2), requiredFields)
+
+// run the build and write the files
+bundle(args.entry, args.platform)
```
jonathanKingston, line 56: Can we lob this in an `async init` to remove `.then`s?

scripts/inject.js
shakyShane: let's put both entry + output into the script?  I'd prefer that anyway, easier to track changes over time too in git

scripts/generateSJCL.js
shakyShane: but the code is vendored already, so it only breaks windows if there's a change. regardless let's change to check for 'perl' anyway

jonathanKingston (approved):

PR #375, opened 2023-04-03
Use original video element's ID for Click to Load placeholders

kzar:
Previously, we used a (hopefully) unique ID for each YouTube Click to
Load placeholder, so that we could find it again if the placeholder
needed to be replaced (when toggling video previews). The problem with
that was that the website will sometimes need to select the video
element by ID.

Until now, that problem didn't generally surface, because we were
waiting until after page load to display the placeholders. Page
scripts could generally select the video element before page
load. Since that's no longer the case, let's use the same ID for the
placeholder as the original video element had, and keep track of the
placeholders inside the DuckWidget class.


franfaccin (approved):
IDs are added as expected 👍  ✅

PR #380, opened 2023-04-03
Add Android messaging integration

joshliebe:
https://app.asana.com/0/0/1204325992866076/f

Adds the messaging integration for sending and receiving messages from Android.


inject/android.js
```diff
@@ -2,9 +2,23 @@
  * @module Android integration
  * @category Content Scope Scripts Integrations
  */
-import { load, init } from '../src/content-scope-features.js'
+import { load, init, update } from '../src/content-scope-features.js'
 import { processConfig, isGloballyDisabled } from './../src/utils'
 
+const allowedMessages = [
```
jonathanKingston, line 8: Can we drop a utils/messaging file into https://github.com/duckduckgo/content-scope-scripts/tree/main/inject or similar and remove all this copy and paste?

inject/android.js
```diff
@@ -16,10 +30,34 @@ function initCode () {
...
+            return console.warn('Ignoring invalid sendMessage messageType', messageType)
+        }
+
+        const message = {
+            type: messageType,
+            options: event.detail?.options
+        }
+        const stringifiedArgs = JSON.stringify(message)
+        // @ts-ignore
+        window.ClickToLoadAndroid.process(stringifiedArgs)
```
jonathanKingston, line 59: This is exposed through Android right?
I think we could probably make this not CTL specific; even if we're acknowledging that it's legacy upfront. `window.legacyContentScopeScriptsMessageHandler....` etc

joshliebe: That makes sense, I've renamed them to `legacyContentScopeScriptsMessageHandler` and `legacyContentScopeScriptsMessageCallback`.

inject/android.js
joshliebe: I copied this from inject `chrome.js`, @kzar would be best to comment what we do here

jonathanKingston: We can leave for now and just file a clean up PR, a lot of this is on the extension side too (which is my fault https://app.asana.com/0/1201614831475344/1204299256427937/f) we could clean them all up as a single task perhaps.

Just so long as we're all aware that it's in 4 places is fine.

inject/android.js
```diff
@@ -16,10 +31,34 @@ function initCode () {
         platform: processedConfig.platform
     })
 
+    processedConfig.messageSecret = messageSecret
     init(processedConfig)
 
-    // Not supported:
-    // update(message)
+    // @ts-ignore
+    window.legacyContentScopeScriptsMessageCallback = (message) => {
```
kzar, line 38: Not tested, but I suggest we harden this somewhat:

```suggestion
const secret = PROVIDED_BY_ANDROID
const secretMethodName = ALSO_PROVIDED_BY_ANDROID // Not containing above secret.

const wrappedUpdate = (providedSecret, ...args) => {
    if (providedSecret === secret) {
        update(...args);
    }
}.bind()

Object.defineProperty(window, secretMethodName, {
    value: wrappedUpdate
})

// Then Android code can call window[secretMethodName](secret, ...)
```

Notes:
 - `.bind()` prevents wrappedUpdate.toString() from displaying the function's contents.
 - Object.defineProperty (without specifying configurable or writable) won't allow the website to redefine the function.
 - Checking the secret should hopefully prevent the website calling update() itself.

kzar (changes requested):

inject/android.js
```diff
@@ -16,10 +31,34 @@ function initCode () {
...
+            return console.warn('Ignoring invalid sendMessage messageType', messageType)
+        }
+
+        const message = {
+            type: messageType,
+            options: event.detail?.options
+        }
+        const stringifiedArgs = JSON.stringify(message)
+        // @ts-ignore
+        window.legacyContentScopeScriptsMessageHandler.process(stringifiedArgs)
```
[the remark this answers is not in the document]
joshliebe: Yes, this is used for sending messages back to the app.
To do this I use the `addJavascriptInterface()` method on `WebView` to add the interface.

I don't have any control where it puts this unfortunately. This is from the WebView docs:
_"Because the object is exposed to all the frames, any frame could obtain the object name and call methods on it. There is no way to tell the calling frame's origin from the app side, so the app must not assume that the caller is trustworthy unless the app can guarantee that no third party content is ever loaded into the WebView even inside an iframe."_

So what I can do is check the secret on the other side and if it doesn't match then ignore the message.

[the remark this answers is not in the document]
kzar: OK got it. If possible, I would:
1. Put this directly in the `window` Object, instead of `window.legacyContentScopeScriptsMessageHandler`.
2. Give it a random name, that's then also passed over from the Android code with the other secrets.

Passing the secret into the function makes sense, but we do need to make sure the website can't change the function. If something like `window.legacyContentScopeScriptsMessageHandler.process = evilFunction` is possible, the secret could get intercepted. `Object.getOwnPropertyDescriptor(window.legacyContentScopeScriptsMessageHandler, 'process')` should tell us if that's configurable or writable. Alternatively, something like `window.legacyContentScopeScriptsMessageHandler.process = alert` could be a quick test.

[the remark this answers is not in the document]
jonathanKingston: 1. isn't possible
2. is possible :)

> If something like window.legacyContentScopeScriptsMessageHandler.process = evilFunction is possible

iirc we never checked if this wrapping is possible; see "Spoofing communication boundaries" under https://app.asana.com/0/944732614786175/1202203984294184/f I think assignment was already checked but there's other ways listed there.

inject/android.js
joshliebe: I've added these changes and it works as expected 👍

inject/android.js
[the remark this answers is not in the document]
joshliebe: 1. Isn't possible as Jonathan stated, we can only add the interface onto the `window`
2. I've updated the code to pass in a random name from Android 👍

@kzar I ran `Object.getOwnPropertyDescriptor(window.legacyContentScopeScriptsMessageHandler, 'process')` and result is `{"writable":true,"enumerable":true,"configurable":true}`. Let me know if there's anything else I can do here.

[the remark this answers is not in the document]
jonathanKingston: I wonder if we can/should take the windows approach of grabbing the handler then deleting it.

inject/android.js
```diff
@@ -16,10 +25,43 @@ function initCode () {
...
+        if (!allowedMessages.includes(messageType)) {
+            return console.warn('Ignoring invalid sendMessage messageType', messageType)
+        }
+
+        const message = {
+            type: messageType,
+            options: event.detail?.options,
+            secret: messageSecret
+        }
+        const stringifiedArgs = JSON.stringify(message)
```
kzar, line 61: Is this step required? If so, is it possible to at least avoid passing the secret into `JSON.stringify()`? The website might have messed with that function, so if possible I'd like to avoid giving it the secret.

inject/android.js
```diff
@@ -16,10 +25,43 @@ function initCode () {
...
-    // update(message)
+    window.addEventListener('sendMessageProxy' + messageSecret, event => {
+        event.stopImmediatePropagation()
+
+        if (!(event instanceof CustomEvent) || !event?.detail) {
+            return console.warn('no details in sendMessage proxy', event)
+        }
+
+        const messageType = event.detail?.messageType
+        if (!allowedMessages.includes(messageType)) {
```
kzar, line 52: Note: It could be worth checking the message is allowed in the Android code as well.

inject/android.js
```diff
@@ -16,10 +25,43 @@ function initCode () {
         platform: processedConfig.platform
     })
 
+    const messageSecret = processedConfig.messageSecret
+    const messageCallback = processedConfig.messageCallback
```
kzar, line 29: Nit: Could you add comments for `messageCallback` and `messageInterface` to make it clear which is for sending messages to the Android code and which is for receiving messages from the Android code?

inject/android.js
```diff
@@ -16,10 +25,43 @@ function initCode () {
...
+        }
+
+        const message = {
+            type: messageType,
+            options: event.detail?.options,
+            secret: messageSecret
+        }
+        const stringifiedArgs = JSON.stringify(message)
+        // @ts-ignore
+        window[messageInterface].process(stringifiedArgs)
```
kzar, line 63:
```suggestion
        // Earlier in the file...
        const sendMessageToAndroid = window[messageInterface].process.bind(window[messageInterface])
        
        // Here
        sendMessageToAndroid(secret, stringifiedArgs)
```

That's about as good as we can do to prevent them messing with `window.messageInterface` + has the bonus of giving the wrapped function a clearer name.

kzar (changes requested):
Looking great, a few more suggestions

inject/android.js
joshliebe: I've sent the secret as a separate argument in process 👍

inject/android.js
joshliebe: I've made a task to remind me to do that 👍

inject/android.js
```diff
@@ -16,10 +25,47 @@ function initCode () {
...
+        if (providedSecret === messageSecret) {
+            update(...args)
+        }
+    }).bind()
+
+    Object.defineProperty(window, messageCallback, {
+        value: wrappedUpdate
+    })
+
     init(processedConfig)
```
kzar, line 44: Nit: It would be better to move the init() call just below the `sendMessageToAndroid` lines. E.g.

```javascript
const sendMessageToAndroid = window[messageInterface].process.bind(window[messageInterface])
delete window[messageInterface]

init(processedConfig)

window.addEventListener('sendMessageProxy' + messageSecret, event => {
```

kzar (approved):
LGTM % One nit

kzar (approved):
LGTM

PR #383, opened 2023-04-04
Support arbitary depth config overloading of DOM properties

jonathanKingston:
See: https://app.asana.com/0/0/1204271448367771/f

Allows us to override properties to any depth within the DOM.

### DIFF
--- a/src/features/runtime-checks.js
+++ b/src/features/runtime-checks.js
@@ -187,11 +187,20 @@ class DDGRuntimeChecks extends HTMLElement {
             for (const [key, value] of scope) {
                 const varOutName = [...scopePath, key].join('_')
                 if (value instanceof Map) {
+                    const proxyName = `_proxyFor_${varOutName}`
+                    output += `
+                    let ${proxyName}
+                    if (${scopePath.join('?.')}?.${key} === undefined) {
+                        ${proxyName} = Object.bind(null);
+                    } else {
+                        ${proxyName} = ${scopePath.join('.')}.${key};
+                    }
+                    `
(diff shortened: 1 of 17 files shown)


src/features/runtime-checks.js
```diff
@@ -157,32 +162,64 @@ class DDGRuntimeChecks extends HTMLElement {
...
+         */
+        function stringifyScope (scope, scopePath) {
+            let output = ''
+            for (const [key, value] of scope) {
+                const varOutName = [...scopePath, key].join('_')
+                if (value instanceof Map) {
+                    const proxyName = `_proxyFor_${varOutName}`
+                    output += `
+                    let ${proxyName}
+                    if (${scopePath.join('?.')}?.${key} === undefined) {
```
shakyShane, line 193: with this not be transpiled, I'm assuming you're ok with the support level of optional chaining?

shakyShane (dismissed):
My only comments on this would be around testing the string output in the future too, to compliment the integration testing. 
 
Since we're just composing a complicated string - I think it would be good to have a base output to compare against (like an actual JS file). It might help surface any problems and would be good for understanding the intention.

On the other hand, it's very hard to construct a string that can be compared against a base one - let's just keep it in mind imo

src/features/runtime-checks.js
jonathanKingston: 👀 checks notes

jonathanKingston: I think we're good based on the pixels I'm seeing.

shakyShane (approved):
Nice!

PR #385, opened 2023-04-05
Clean indenting of generated code

jonathanKingston:


src/features/runtime-checks/script-overload.js
```diff
@@ -61,44 +124,6 @@ export function wrapScriptCodeOverload (code, config) {
         currentScope.set(pathOut, value)
     }
 
-    /**
-     * Output scope variable definitions to arbitrary depth
-     */
-    function stringifyScope (scope, scopePath) {
```
jonathanKingston, line 67: Moved out a bunch of these functions to the top level to improve readability.

src/features/runtime-checks/script-overload.js
```diff
@@ -1,5 +1,96 @@
 import { processAttr } from '../../utils.js'
 
+/**
+ * Indent a code block using braces
+ * @param {string} string
+ * @returns {string}
+ */
+function removeIndent (string) {
+    const lines = string.split('\n')
```
jonathanKingston, line 9: This was much simpler than trying to make tagged template strings play ball, I spent an hour or so on trying to re-indent the values to the in place indent; it was a whole sea of sadness.

I'm pretty sure this will be faster perf wise and it appears to solve most use-cases.

shakyShane (approved):

PR #390, opened 2023-04-05
Ensure Click to Load events are dispatched at the right times

kzar:
The Click to Load content script needs to communicate with the Click
to Load surrogate scripts sometimes. This is handled by dispatching
events on the page.

If those events are dispatched _before_ the receiving script has
loaded, the events will be missed and aspects of the feature
broken. So let's wait until page load before dispatching them.

That is made harder, by the fact that some of the events must be
dispatched targetting the tracking elements. That way, the receiver
can maintain a lookup of placeholder element <-> tracking element. If
the tracking element is removed _before_ an event is dispatched that
targets it, the event will no longer bubble up the DOM and be
received. To avoid that from happening, let's hide the tracking
element until the events are fired, and then unhide + remove it from
the DOM.

Worse still, it's also important that those events targetting the
tracking (and placeholder) elements are fired _before_ the
"ddg-ctp-ready" event. So let's also take care to wait for all the
element events to fire before sending it!

Note: This all still assumes that the website does not load the
      scripts (that are redirected to surrogate scripts) after page
      load has completed.

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -27,8 +27,19 @@ const entityData = {}
 // Used to avoid displaying placeholders for the same tracking element twice.
 const knownTrackingElements = new WeakSet()
 
-let readyResolver
-const ready = new Promise(resolve => { readyResolver = resolve })
+// Promise that is resolved when the Click to Load feature init() function has
+// finished its work, enough that it's now safe to replace elements with
+// placeholders.
+let readyToDisplayPlaceholdersResolver
+const readyToDisplayPlaceholders = new Promise(resolve => {
+    readyToDisplayPlaceholdersResolver = resolve
+})
+
+// Promise that is resolved when the page has finished loading (and
+// readyToDisplayPlaceholders has resolved). Wait for this before sending
+// essential messages to surrogate scripts.
+let afterPageLoadResolver
+const afterPageLoad = new Promise(resolve => { afterPageLoadResolver = resolve })
 
 /*********************************************************
  *  Widget Replacement logic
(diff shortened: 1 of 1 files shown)


src/features/click-to-load.js
```diff
@@ -24,8 +24,20 @@ const entityData = {}
 // Used to avoid displaying placeholders for the same tracking element twice.
 const knownTrackingElements = new WeakSet()
```
franfaccin, line 26: I'm getting this error when toggling on/off the previews:
image

kzar: I've not been able to reproduce that, did rebuilding with latest extension (and tracker-surrogates) code help?

src/features/click-to-load.js
```diff
@@ -499,7 +566,7 @@ function showExtraUnblockIfShortPlaceholder (shadowRoot, placeholder) {
  *   in the document will be replaced instead.
  */
 async function replaceClickToLoadElements (targetElement) {
-    await ready
+    await afterInit
```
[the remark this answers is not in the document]
kzar: > It seems like you have only one call site into this method in update, are you able to await there for pageload, init etc rather than scattering awaiting through the code?

We could potentially wait for `afterInit` before calling `replaceClickToLoadElements` like you suggest, though I think having the await inside the function makes more sense personally.

We don't want to wait for page load before replacing elements though, since we want the placeholders to show up right away.

sammacbeth (changes requested):
I'm trying to get my head around the `afterInit` and `afterPageLoad` phases. It seems that:
 1. `afterInit`:
     - resolves after we get a `getClickToLoadState` message via `update` (unrelated to the class's `init` method).
     - is used in `replaceClickToLoadElements` (which is called once from `update`).
     - is awaited in `init` to delay triggering of `afterPageLoad`
 2. `afterPageLoad`:
     - resolves after the page load event fires AND `afterInit` has resolved.
     - is used in `replaceTrackingElement` to delay message dispatch
     - is used in `init` to send a final `ddg-ctp-ready` event after the `replaceTrackingElement` events have finished dispatching.

I think we can make this a bit easier to understand with some minor changes:
 - For me, the `afterInit` naming is confusing, as we are `awaiting` it in the `init` function. We should name it something to differentiate it from `init`.
 - As `afterPageLoad` is equivalent to `Promise.all(afterInit, pageLoaded)`, which `pageLoaded` is a promise that resolves after the `load` event, we could define it as such and avoid having the confusing async page loaded resolver in the `init` function.
 - As `inProgressElementReplacements` happens synchronously after `afterPageLoad` resolves, then I don't think we need to await those promises. Instead, we just need to wait one tick after `afterPageLoad`. As this is dispatching an async event anyway, I'm not even sure the tick makes a difference.

Do those suggestions make sense? Happy to jump on a call to discuss it further.

kzar:
@sammacbeth Thanks for taking a look so quickly :+1: 

> For me, the `afterInit` naming is confusing, as we are `awaiting` it in the `init` function. We should name it something to differentiate it from `init`.

Good point, I'll rename that.

> As `afterPageLoad` is equivalent to `Promise.all(afterInit, pageLoaded)`, which `pageLoaded` is a promise that resolves after the `load` event, we could define it as such and avoid having the confusing async page loaded resolver in the `init` function.

I mean, I do define it as such. But if you like I can move that outside of the init function, if that's what you mean? (If not, could you explain what you're suggesting?)

>  As `inProgressElementReplacements` happens synchronously after `afterPageLoad` resolves, then I don't think we need to await those promises. Instead, we just need to wait one tick after `afterPageLoad`. As this is dispatching an async event anyway, I'm not even sure the tick makes a difference.

In my testing, it did make a difference unfortunately. (I lost maybe a day or two trying to figure out WTF was going on there!) If you want to reproduce, [enable the last integration tests](https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/main/integration-test/click-to-load-youtube.spec.js#L166), make your changes and then see if the tests still pass after content-scope-scripts and the extension are rebuilt. (The video in question is the one on the test page that shows a coloured border, and can be controlled by play/stop buttons in the website.)

That said, I'll try again to see if I can make that work. If we can figure out a simpler approach that works reliably then I'm all for it.

kzar:
> That said, I'll try again to see if I can make that work. If we can figure out a simpler approach that works reliably then I'm all for it.

I'm not sure why it didn't work last time I tried, but happily it worked this time. I think you were right, that we could just wait for the next tick :sunglasses:

src/features/click-to-load.js
kzar: _(For the record, we discussed this in MatterMost and after fetching the latest node_modules this issue went away.)_

sammacbeth:
Thanks @kzar - that's already an improvement. 

> I mean, I do define it as such. But if you like I can move that outside of the init function, if that's what you mean? (If not, could you explain what you're suggesting?)

It is implemented as that, yes, but at some complexity cost. You have to look into the init function and follow the async flow to learn that it resolves when the latest of the page load event and `readyToDisplayPlaceholders` resolves. IMO, having `afterPageLoad` synced to the page load event allows us to put the full definition of that promise in a single place, then use `Promise.all([afterPageLoad, readyToDisplayPlaceholders])` which is more descriptive about what we're waiting for here.

I also plaid around with how we can make the flow simpler, and found that we can remove `async` from almost every function here, then remove the corresponding redundant `await`. This makes it easier to reason about the order of operations because without the `async`/`await` we _know_ that the code is sync and won't race. Having made those changes, I committed them, so you can cherry-pick if you like: https://github.com/duckduckgo/content-scope-scripts/commit/155731bd0bc883c62a6ae61ab5397e3b279257ee. I tested this with the YT CTL tests in the extension and they worked for me locally, so this patch doesn't change the code behavior.

kzar:
> It is implemented as that, yes, but at some complexity cost. You have to look into the init function and follow the async flow to learn that it resolves when the latest of the page load event and `readyToDisplayPlaceholders` resolves. IMO, having `afterPageLoad` synced to the page load event allows us to put the full definition of that promise in a single place, then use `Promise.all([afterPageLoad, readyToDisplayPlaceholders])` which is more descriptive about what we're waiting for here.

I think at this point it's subjective. To me, having this inside the init function is clearer and easier to follow.

> I also plaid around with how we can make the flow simpler, and found that we can remove `async` from almost every function here, then remove the corresponding redundant `await`. This makes it easier to reason about the order of operations because without the `async`/`await` we _know_ that the code is sync and won't race. Having made those changes, I committed them, so you can cherry-pick if you like: [155731b](https://github.com/duckduckgo/content-scope-scripts/commit/155731bd0bc883c62a6ae61ab5397e3b279257ee). I tested this with the YT CTL tests in the extension and they worked for me locally, so this patch doesn't change the code behavior.

Thanks, but those are unrelated changes. I intend to come back to getting the remaining lining enabled (and passing) for this file, which will involve a similar change, but for now I want to focus on getting the events dispatched correctly (and therefore Click to Load YouTube working again.)

sammacbeth (approved):
Let's follow up with the `async`/`await` changes in a clean-up PR.

PR #392, opened 2023-04-06
Take care to set hasMozProxies for Firefox builds again

kzar:
The build script was checking if one of the command line arguments
were "firefox", before setting a variable. Unfortunately, the command
line arguments were moved around[1] which broke this check. Let's fix
that, copying the parsing logic used elsewhere.

1 - https://github.com/duckduckgo/content-scope-scripts/commit/03751df2274120e7e007cf64bcffcc19363977bc

### DIFF
--- a/scripts/utils/build.js
+++ b/scripts/utils/build.js
@@ -4,6 +4,7 @@ import replace from '@rollup/plugin-replace'
 import resolve from '@rollup/plugin-node-resolve'
 import dynamicImportVariables from 'rollup-plugin-dynamic-import-variables'
 import { runtimeInjected } from '../../src/features.js'
+import { parseArgs } from '../script-utils.js'
 
 /**
  * This is a helper function to require all files in a directory
@@ -46,7 +47,8 @@ function runtimeInjections () {
 export async function rollupScript (scriptPath, name, supportsMozProxies = true) {
     let mozProxies = false
     // The code is using a global, that we define here which means once tree shaken we get a browser specific output.
-    if (process.argv[2] === 'firefox' && supportsMozProxies) {
+    const args = parseArgs(process.argv.slice(2), ['platform'])
+    if (args.platform === 'firefox' && supportsMozProxies) {
         mozProxies = true
     }
     const inputOptions = {


jonathanKingston:
Overall this is _fine_, but I would like to understand _why_ given we rely on the check everywhere else also.

> For whatever reason, that check does not seem to always work

This is check now is essentially is an isFirefox check based on the build script.
Is there a reason that `cloneInto` isn't defined ever?

kzar:
> Overall this is _fine_, but I would like to understand _why_ given we rely on the check everywhere else also.
> 
> > For whatever reason, that check does not seem to always work
> 
> This is check now is essentially is an isFirefox check based on the build script. Is there a reason that `cloneInto` isn't defined ever?

Honestly, I don't know. I just noticed `hasMozProxies` was false and that was causing `cloneInto` to not be called when it needed to be. I'm not sure how a lot of this Firefox proxy code works. But I figure this change is fairly harmless since 1. the logic was used previously, when the Click to load code lived in the extension, and 2. I figure checking if a function exists before calling it is pretty unlikely to go wrong.

What do you think?

jonathanKingston:
It is harmless but I'm very concerned we've already broken Firefox with build changes.

Let's continue to investigate; this matches my experience of local Firefox development lately; just been too slammed to report it.

kzar:
> It is harmless but I'm very concerned we've already broken Firefox with build changes.
> 
> Let's continue to investigate; this matches my experience of local Firefox development lately; just been too slammed to report it.

So it seems that [this check fails](https://github.com/duckduckgo/content-scope-scripts/blob/9ccbd3faf44f4e6a56e4f3effeada3820cdc4ec5/scripts/utils/build.js#L49) as `process.argv[2]` is `"--platform"` instead of `"firefox"`. The command being run by `npm run build-firefox` is `node scripts/inject.js --platform firefox` , so that makes sense.

kzar:
This is ready for review again now @shakyShane / @jonathanKingston . The failure seems to be unrelated (an arbitrary bundle size being exceeded.)

kzar:
Closing as @jonathanKingston is working on a similar change.

PR #396, opened 2023-04-08
Performance monitoring

jonathanKingston:
Adds some very basic performance monitoring.

To use this in the extension, you need to use the devtools-panel.html flow to get debug: true set on the tab.
This sets marks by default but never will monitor unless the debug flag is set.

Opening the performance monitor in the browser, you can zoom in and see each part of the code:
image

### DIFF
--- a/src/performance.js
+++ b/src/performance.js
@@ -20,7 +20,7 @@ export class PerformanceMonitor {
     /**
      * Measure all performance markers
      */
-    measure () {
+    measureAll () {
         this.marks.forEach((mark) => {
             mark.measure()
         })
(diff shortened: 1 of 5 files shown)


src/performance.js
```diff
@@ -0,0 +1,49 @@
...
+    mark (name) {
+        const mark = new PerformanceMark(name)
+        this.marks.push(mark)
+        return mark
+    }
+
+    /**
+     * Measure all performance markers
+     */
+    measure () {
```
shakyShane, line 23:
```suggestion
    measureAll () {
```

shakyShane: NAB: because you can `measure` a `Mark`, how about making this `measureAll` for call-site clarity?

shakyShane (approved):
Looks great - added a comment but feel free to ignore :)

src/performance.js
jonathanKingston: Renamed :)

shakyShane (approved):

PR #415, opened 2023-04-12
Partially revert "Tighten linting" changes to Click to Load code

kzar:
Recently, we adjusted the linting rules to make them stricter. While
doing that, we made some changes to the Click to Load code to make the
linting pass. Unfortunately, that complicated ongoing (and more
urgent) changes to that code that are about to land.

Let's revert the linting changes to the Click to Load code therefore,
while keeping the linting passing + adding a note to revisit.

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -1,3 +1,7 @@
+// TODO - Remove these comments to enable full linting.
+/* eslint-disable @typescript-eslint/ban-ts-comment, @typescript-eslint/await-thenable, require-await */
+// @ts-nocheck
+
 import { createCustomEvent, sendMessage, OriginalCustomEvent, originalWindowDispatchEvent } from '../utils.js'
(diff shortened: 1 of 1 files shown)


jonathanKingston:
What was the issue? Did you apply https://github.com/duckduckgo/content-scope-scripts/pull/411 over your patch.

jonathanKingston:
Ok sure, please get a **real review** before merging https://github.com/duckduckgo/content-scope-scripts/pull/390

There's clear race conditions and unnecessary complexity in that code that put the entire platform at risk.

jonathanKingston (approved):

kzar:
> Ok sure, please get a **real review** before merging #390

Well, for what it's worth I think the ongoing review with @franfaccin _is_ a "real" review (whatever that means). She knows that code better than pretty much anyone.

> There's clear race conditions and unnecessary complexity in that code that put the entire platform at risk.

I guess we can agree to disagree there, since I'll be having ongoing discussions about the best approach on the review with @franfaccin  and now @sammacbeth  as well.

PR #422, opened 2023-04-13
Improve narrow YouTube Click to Load placeholders

kzar:
When the YouTube Click to Load placeholders are very narrow, the
information text takes up too much space and some of the other
elements are lost. Let's tweak narrow placeholders, to hide that
information text and move the "Learn more" link. This isn't the best
approach (using @container queries[1] might be better), but this code
is going to be refactored shortly anyway.

1 - https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -589,6 +589,7 @@ function replaceYouTubeCTL (trackingElement, widget) {
         resizeElementToMatch(oldPlaceholder || trackingElement, blockingDialog)
         replaceTrackingElement(widget, trackingElement, blockingDialog)
         showExtraUnblockIfShortPlaceholder(shadowRoot, blockingDialog)
+        hideInfoTextIfNarrowPlaceholder(shadowRoot, blockingDialog, 460)
     }
 }
 
@@ -625,6 +626,44 @@ function showExtraUnblockIfShortPlaceholder (shadowRoot, placeholder) {
             innerDiv.style.maxHeight = parentHeight + 'px'
             innerDiv.style.overflow = 'hidden'
         }
+    }
+}
+
+/**
+ * Hide the info text (and move the "Learn More" link) if the placeholder is too
+ * narrow.
+ * @param {ShadowRoot} shadowRoot
+ * @param {HTMLElement} placeholder Placeholder for tracking element
+ * @param {number} narrowWidth
+ *    Maximum placeholder width (in pixels) for the placeholder to be considered
+ *    narrow.
+ */
+function hideInfoTextIfNarrowPlaceholder (shadowRoot, placeholder, narrowWidth) {
+    const { width: placeholderWidth } = placeholder.getBoundingClientRect()
+    if (placeholderWidth > 0 && placeholderWidth <= narrowWidth) {
+        const buttonContainer =
+              shadowRoot.querySelector('.DuckDuckGoButton.primary')?.parentElement
+        const contentTitle = shadowRoot.getElementById('contentTitle')
+        const infoText = shadowRoot.getElementById('infoText')
+        /** @type {HTMLElement?} */
+        const learnMoreLink = shadowRoot.getElementById('learnMoreLink')
+
+        // These elements will exist, but this check keeps TypeScript happy.
+        if (!buttonContainer || !contentTitle || !infoText || !learnMoreLink) {
+            return
+        }
+
+        // Remove the information text.
+        infoText.remove()
+        learnMoreLink.remove()
+
+        // Append the "Learn More" link to the title.
+        contentTitle.innerText += '. '
+        learnMoreLink.style.removeProperty('font-size')
+        contentTitle.appendChild(learnMoreLink)
+
+        // Improve margin/padding, to ensure as much is displayed as possible.
+        buttonContainer.style.removeProperty('margin')
     }
 }
 
@@ -749,6 +788,7 @@ function getLearnMoreLink (mode = 'lightMode') {
     linkElement.href = 'https://help.duckduckgo.com/duckduckgo-help-pages/privacy/embedded-content-protection/'
     linkElement.target = '_blank'
     linkElement.textContent = sharedStrings.learnMore
+    linkElement.id = 'learnMoreLink'
     return linkElement
 }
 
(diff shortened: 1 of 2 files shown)


src/features/click-to-load.js
```diff
@@ -492,6 +493,40 @@ function showExtraUnblockIfShortPlaceholder (shadowRoot, placeholder) {
...
+ * Hide the info text (and move the "Learn More" link) if the placeholder is too
+ * narrow.
+ * @param {Element} shadowRoot
+ * @param {Element} placeholder Placeholder for tracking element
+ * @param {number} narrowWidth
+ *    Maximum placeholder width (in pixels) for the placeholder to be considered
+ *    narrow.
+ */
+function hideInfoTextIfNarrowPlaceholder (shadowRoot, placeholder, narrowWidth) {
+    const { height: placeholderHeight } = window.getComputedStyle(placeholder)
```
kzar, line 506: (I need to change this to `getBoundingClientRect()` after my zoom call.)

src/features/click-to-load.js
```diff
@@ -628,6 +629,48 @@ function showExtraUnblockIfShortPlaceholder (shadowRoot, placeholder) {
...
+        learnMoreLink.remove()
+
+        // Append the "Learn More" link to the title.
+        contentTitle.innerText += '. '
+        learnMoreLink.style.removeProperty('font-size')
+        contentTitle.appendChild(learnMoreLink)
+
+        // Improve margin/padding, to ensure as much is displayed as possible.
+        buttonContainer.style.removeProperty('margin')
+        innerDiv.style.height = 'calc(100% - 80px)'
```
franfaccin, line 670: Could you tell a bit more about why the `calc(100% - 80px)` height?

src/features/click-to-load.js
```diff
@@ -749,6 +792,7 @@ function getLearnMoreLink (mode = 'lightMode') {
     linkElement.href = 'https://help.duckduckgo.com/duckduckgo-help-pages/privacy/embedded-content-protection/'
     linkElement.target = '_blank'
     linkElement.textContent = sharedStrings.learnMore
+    linkElement.classList.add('learnMoreLink')
```
franfaccin, line 795: Any particular motivations for doing a class here and not an id like in the other elements?

unit-test/verify-artifacts.js
```diff
@@ -6,7 +6,7 @@ import { cwd } from '../scripts/script-utils.js'
 const ROOT = join(cwd(import.meta.url), '..')
 const BUILD = join(ROOT, 'build')
 const APPLE_BUILD = join(ROOT, 'Sources/ContentScopeScripts/dist')
-const CSS_OUTPUT_SIZE = 551000
+const CSS_OUTPUT_SIZE = 560000
```
franfaccin, line 9: Just general question, is there a limit to how much we're willing to increase the CSS output? Or for now this is just to keep track of it?

src/features/click-to-load.js
```diff
@@ -589,6 +589,7 @@ function replaceYouTubeCTL (trackingElement, widget) {
         resizeElementToMatch(oldPlaceholder || trackingElement, blockingDialog)
         replaceTrackingElement(widget, trackingElement, blockingDialog)
         showExtraUnblockIfShortPlaceholder(shadowRoot, blockingDialog)
+        hideInfoTextIfNarrowPlaceholder(shadowRoot, blockingDialog, 460)
```
franfaccin, line 592: We now have a situation where we can trigger both the extra unblock because the placeholder is short, and the shorter copy version. Maybe we can make `showExtraUnblockIfShortPlaceholder()` test for an even shorter height, since the button is visible more often with these changes.

image

franfaccin (commented):
I think you're handling well these changes. I wonder if it would be best to create a new text in the config instead of moving the "learn more" around. But there wouldn't be much gain either way, and like you said, we will likely refactor this code soon, so it should be fine to leave it like this to unblock you. I left some other general comments, ping me if you have any questions :)

unit-test/verify-artifacts.js
kzar: Yea, that's a good question. The original idea was so that we'd know when we hit half-a-megabyte, but since then we've been slowly increasing it as the test fails.

src/features/click-to-load.js
kzar: Argh, good point!

I'm not sure how to fix that though, both of these functions aren't ideal. For example, `showExtraUnblockIfShortPlaceholder` doesn't take into account the placeholder's position inside the parent element. [Take the short-placeholder test case](https://privacy-test-pages.glitch.me/privacy-protections/youtube-click-to-load/) where the parent element would be big enough for the placeholder's primary unblock button to be visible, except there is some text above the placeholder.

I'm inclined to leave this as is, and hope it will be sorted out by your refactor. What do you think?

src/features/click-to-load.js
kzar: The existing height is `calc(100% - 30px);` and I needed it to be a bit shorter. So with trial and error I found `calc(100% - 80px);` looked about right. I'm not sure why the original height was `calc(100% - 30px);` though, that was before my time.

src/features/click-to-load.js
kzar: > I wonder if it would be best to create a new text in the config instead of moving the "learn more" around.

Yea, you're right there. Moving the string around like this isn't correct really, and with the new preview sub-heading string the link is included. I figured this is pragmatic for now though, in particular since we're doing the same thing when adding this particular link - appending it to a div with the text content of `hoverTextBody + ' '`.

kzar:
> I wonder if it would be best to create a new text in the config instead of moving the "learn more" around.

Yea, you're right there. Moving the string around like this isn't correct really, and with the new preview sub-heading string the link is included. I figured this is pragmatic for now though, in particular since we're doing the same thing when adding this particular link originally - we append it to a div with the text content of `hoverTextBody + ' '`.

src/features/click-to-load.js
franfaccin: I think I was the one that added `calc(100% - 30px);`, and it was to in order to fit the "Share feedback" link under the placeholder. That's why I found it odd to have `80px` here now. You got to 80px because of the site where Costas was testing? If so, I think we can continue with 30px, as that website has some particular cropping of the video area and we shouldn't be tailoring the styles based only on that.

src/features/click-to-load.js
franfaccin: My suggestion was just to lower the height that `showExtraUnblockIfShortPlaceholder()` is already checking. But I just noticed this is the same for FB CTL, so it is better if we handle this during the refactor.

src/features/click-to-load.js
kzar: Fair enough, I have removed that for now :+1:

franfaccin (approved):

PR #432, opened 2023-04-14
Fix linting issues in Click to Load code

kzar:
There were a number of ESLint and TypeScript linting issues in the
Click to Load code. Let's fix those and enable full linting for the
file.

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -1,11 +1,16 @@
-// TODO - Remove these comments to enable full linting.
-/* eslint-disable @typescript-eslint/ban-ts-comment, @typescript-eslint/await-thenable, require-await */
-// @ts-nocheck
-
-import { createCustomEvent, sendMessage, OriginalCustomEvent, originalWindowDispatchEvent } from '../utils.js'
+import { createCustomEvent, sendMessage, originalWindowDispatchEvent } from '../utils.js'
 import { logoImg, loadingImages, closeIcon } from './click-to-load/ctl-assets.js'
 import { styles, getConfig } from './click-to-load/ctl-config.js'
 import ContentFeature from '../content-feature.js'
+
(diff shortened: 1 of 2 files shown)


kzar:
@jonathanKingston As promised, here is a PR to get the Click to Load code passing linting. I have left this as a draft while I test I didn't break anything.

src/features/click-to-load.js
```diff
@@ -1389,18 +1590,19 @@ export default class ClickToLoad extends ContentFeature {
         }
 
         // Listen for events from "surrogate" scripts.
-        addEventListener('ddg-ctp', (event) => {
+        addEventListener('ddg-ctp', (/** @type {CustomEvent} */ event) => {
             if (!event.detail) return
```
jonathanKingston, line 1594: I think this is not narrowing TS unfortunately, the following would remove the need for some of the ?'s below
```suggestion
            if (!('detail' in event)) return
```

src/features/click-to-load.js
```diff
@@ -54,10 +67,17 @@ class DuckWidget {
         this.gatherDataElements()
         this.entity = entity
         this.widgetID = Math.random()
+        this.autoplay = false
```
jonathanKingston, line 70: Why was this added?

src/features/click-to-load.js
kzar: Done.

src/features/click-to-load.js
kzar: We were setting this later, which caused TypeScript to get upset that DuckWidget didn't have an `autoplay` property. Better to set it when constructing the Object instead, plus that keeps TypeScript happy when we set the property to true later.

jonathanKingston (approved):
LGTM thanks for this :)

PR #436, opened 2023-04-17
Update Click to Load YouTube (preview-enabled) footer text

kzar:
During internal review, it was requested that we change the
footer-text for Click to Load YouTube placeholders.


ladamski (approved):
Tested this and LGTM!  The only small nit maybe is that the "Learn more" currently opens in the active tab, I would expect this to open a new tab or window instead?

ladamski (approved):
👍

PR #447, opened 2023-04-19
Per-platform feature bundling

shakyShane:
Allow platforms to opt-in to features they want bundled

### DIFF
--- a/scripts/inject.js
+++ b/scripts/inject.js
@@ -59,7 +59,7 @@ async function initOther (injectScriptPath, platformName) {
  */
 async function initChrome (entry, platformName) {
     const replaceString = '/* global contentScopeFeatures */'
-    const injectScript = await rollupScript({ scriptPath: entry })
+    const injectScript = await rollupScript({ scriptPath: entry, platform: platformName })
     const contentScope = await rollupScript({
         scriptPath: contentScopePath,
         name: contentScopeName,
--- a/scripts/inject.js
+++ b/scripts/inject.js
@@ -14,6 +14,10 @@ const builds = {
         input: 'inject/apple.js',
         output: ['Sources/ContentScopeScripts/dist/contentScope.js']
     },
+    'apple-isolated': {
+        input: 'inject/apple.js',
+        output: ['Sources/ContentScopeScripts/dist/contentScopeIsolated.js']
+    },
     android: {
         input: 'inject/android.js',
         output: ['build/android/contentScope.js']
@@ -43,18 +47,28 @@ const builds = {
 async function initOther (injectScriptPath, platformName) {
     const supportsMozProxies = platformName === 'firefox'
     const identName = `inject${camelcase(platformName)}`
-    const injectScript = await rollupScript(injectScriptPath, identName, supportsMozProxies)
+    const injectScript = await rollupScript({
+        scriptPath: injectScriptPath,
+        name: identName,
+        supportsMozProxies,
+        platform: platformName
+    })
     const outputScript = injectScript
     return outputScript
 }
 
 /**
  * @param {string} entry
+ * @param {string} platformName
  */
-async function initChrome (entry) {
+async function initChrome (entry, platformName) {
     const replaceString = '/* global contentScopeFeatures */'
-    const injectScript = await rollupScript(entry)
-    const contentScope = await rollupScript(contentScopePath, contentScopeName)
+    const injectScript = await rollupScript({ scriptPath: entry })
+    const contentScope = await rollupScript({
+        scriptPath: contentScopePath,
+        name: contentScopeName,
+        platform: platformName
+    })
     // Encode in URI format to prevent breakage (we could choose to just escape ` instead)
     // NB: .replace(/\r\n/g, "\n") is needed because in Windows rollup generates CRLF line endings
     const encodedString = encodeURI(contentScope.toString().replace(/\r\n/g, '\n'))
@@ -74,7 +88,7 @@ async function init () {
 
     let output
     if (args.platform === 'chrome') {
-        output = await initChrome(build.input)
+        output = await initChrome(build.input, args.platform)
     } else {
         output = await initOther(build.input, args.platform)
     }
--- a/scripts/inject.js
+++ b/scripts/inject.js
@@ -14,10 +14,6 @@ const builds = {
         input: 'inject/apple.js',
         output: ['Sources/ContentScopeScripts/dist/contentScope.js']
     },
-    'apple-isolated': {
-        input: 'inject/apple.js',
-        output: ['Sources/ContentScopeScripts/dist/contentScopeIsolated.js']
-    },
     android: {
         input: 'inject/android.js',
         output: ['build/android/contentScope.js']
--- a/scripts/utils/build.js
+++ b/scripts/utils/build.js
@@ -133,10 +133,10 @@ function loadFeatures (platform, featureNames = platformSupport[platform]) {
             if (id !== pluginId) return null
 
             // convert a list of feature names to
-            const imports = featureNames.map((f, index) => {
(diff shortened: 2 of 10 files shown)


scripts/inject.js
```diff
@@ -43,7 +47,12 @@ const builds = {
 async function initOther (injectScriptPath, platformName) {
     const supportsMozProxies = platformName === 'firefox'
     const identName = `inject${camelcase(platformName)}`
-    const injectScript = await rollupScript(injectScriptPath, identName, supportsMozProxies)
+    const injectScript = await rollupScript({
+        scriptPath: injectScriptPath,
+        name: identName,
+        supportsMozProxies,
+        platform: platformName
```
shakyShane, line 54: this is just converting to an object to pas 'platform' through

scripts/utils/build.js
```diff
@@ -69,11 +93,12 @@ export async function rollupScript (scriptPath, name, supportsMozProxies = false
                 preventAssignment: true,
                 values: {
                     // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
-                    mozProxies
+                    mozProxies,
+                    'import.meta.injectName': JSON.stringify(platform)
```
shakyShane, line 97: needed because we can't use 'args.platform.name' for this purpose

scripts/utils/build.js
```diff
@@ -89,3 +114,39 @@ export async function rollupScript (scriptPath, name, supportsMozProxies = false
...
+            // convert the pathname to a feature name
+            const featureName = base.replace(/-(.)/g, (match, letter) => {
+                return letter.toUpperCase()
+            }).slice(0, -3)
+
+            // if the feature is listed for this platform, include it
+            if (featureNames.includes(featureName)) return null
+
+            // if we get here the feature should be skipped/stubbed
+            return 'export default ' + JSON.stringify('__stub__' + base)
```
shakyShane, line 149: this is all a bit of a hack, but I had the idea of how to prevent Duck Player bloating every bundle by simply 'skipping' some modules in a low-risk-easy-to-ship way. 

as always fully open to other ideas 🙏🏻

src/content-scope-features.js
```diff
@@ -41,6 +42,10 @@ export function load (args) {
         return
     }
 
+    const featureNames = typeof import.meta.injectName === 'string'
+        ? platformSupport[import.meta.injectName]
+        : []
```
shakyShane, line 47: here I'm passing through what I call the `injectName` (open to better naming ideas) - we need this because `args.platform.name` is not quite the thing we want here. We want to reference the platform name based on when it was built, not the runtime. 

runtime from the apple build:

- args.platform.name = 'ios`
- args.platform.name = 'macos`

scripts/inject.js
```diff
@@ -14,6 +14,10 @@ const builds = {
         input: 'inject/apple.js',
         output: ['Sources/ContentScopeScripts/dist/contentScope.js']
     },
+    'apple-isolated': {
```
jonathanKingston, line 17: Remove for this PR

scripts/inject.js
```diff
@@ -53,8 +62,11 @@ async function initOther (injectScriptPath, platformName) {
  */
 async function initChrome (entry) {
     const replaceString = '/* global contentScopeFeatures */'
-    const injectScript = await rollupScript(entry)
-    const contentScope = await rollupScript(contentScopePath, contentScopeName)
+    const injectScript = await rollupScript({ scriptPath: entry })
+    const contentScope = await rollupScript({
```
jonathanKingston, line 66: Is this working for Chrome? You're missing platform here.

scripts/utils/build.js
```diff
@@ -69,11 +93,12 @@ export async function rollupScript (scriptPath, name, supportsMozProxies = false
...
                 values: {
                     // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
-                    mozProxies
+                    mozProxies,
+                    'import.meta.injectName': JSON.stringify(platform)
                 }
             }),
             prefixPlugin(prefixMessage)
-        ]
+        ].filter(Boolean)
```
jonathanKingston, line 101: This adds complexity when it's not needed.
I'd handle the lack of platform in stubFeatures and just do a noop essentially.

scripts/utils/build.js
jonathanKingston: I wonder if we should just fork the import module instead, that's why it's always included and not tree shaked away.

jonathanKingston: Looks like they support 'include' now, can we check if that works?

unit-test/platform-features.js
```diff
@@ -0,0 +1,21 @@
+import { rollupScript } from '../scripts/utils/build.js'
+
+describe('Output validation', () => {
+    fit('Given the correct config we should generate expected code output', async () => {
```
jonathanKingston, line 4:
```suggestion
    it('Given the correct config we should generate expected code output', async () => {
```

shakyShane:
* main
  * **PR #447** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/447" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/447?utm_source=stack-comment).

src/features.js
```diff
@@ -13,13 +11,52 @@ export const featureNames = [
...
+
+const otherFeatures = /** @type {const} */([
+    'clickToLoad',
+    'windowsPermissionUsage',
+    'webCompat'
+])
+
+/** @typedef {baseFeatures[number]|otherFeatures[number]} FeatureName */
+/** @type {Record<string, FeatureName[]>} */
+export const platformSupport = {
```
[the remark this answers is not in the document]
jonathanKingston: SGTM, we can always improve later if it becomes a burden or anything.

unit-test/config.json
```diff
@@ -3,7 +3,8 @@
   "spec_files": [
     "**/*[sS]pec.?(m)js",
     "**/*.js",
-    "!unit-test/script-overload-snapshots/out/*.js"
+    "!unit-test/script-overload-snapshots/out/*.js",
```
jonathanKingston, line 6: Can you explain why we need this?

scripts/utils/build.js
```diff
@@ -54,38 +57,111 @@ function prefixPlugin (prefixMessage) {
...
+        resolveId (id) {
+            if (id === pluginId) return id
+            return null
+        },
+        load (id) {
+            if (id !== pluginId) return null
+
+            // convert a list of feature names to
+            const imports = featureNames.map((f, index) => {
+                const fileName = f.replace(/([a-zA-Z])(?=[A-Z0-9])/g, '$1-').toLowerCase()
```
jonathanKingston, line 137: Can we share this regex with the `getFilename`?

shakyShane: oh yeah, that was the point of moving it haha, missed it

unit-test/config.json
shakyShane: 'fixtures' is just a fake file that I wanted in place to test as much of the code-flow as possible - without this exclusion `jasmine` was trying to load it as a test file :)

scripts/utils/build.js
```diff
@@ -54,38 +57,111 @@ function prefixPlugin (prefixMessage) {
...
+                // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
+                mozProxies,
+                'import.meta.injectName': JSON.stringify(platform)
+            }
+        }),
+        prefixPlugin(prefixMessage)
+    ]
+
+    // if 'platform' was provided, ensure we include the dyn
+    if (platform) {
```
jonathanKingston, line 97: Do we need this optionality?

jonathanKingston: Put another way, why is rollupscript not being passed platform always?

PR #448, opened 2023-04-20
Remove bundled ProximaNova font

sammacbeth:
- Removes ProximaNova font in CTL assets.
- Adds a mechanism for the platform to send asset URLs for custom fonts on `init`, allowing CTL fonts to be customized.
- We can use this in the extension to use bundled fonts that work in all locales (see https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/1892/commits/3ca91a675d7cbb2d5b8a512135edcee2f1f3d327)
- For platforms that don't send asset URLs, they get system default fonts that match the privacy-dashboard.


src/features/click-to-load.js
```diff
@@ -1615,6 +1616,10 @@ export default class ClickToLoad extends ContentFeature {
         const localizedConfig = getConfig(locale)
         config = localizedConfig.config
         sharedStrings = localizedConfig.sharedStrings
+        // update styles if asset config was sent
+        if (args?.assets) {
+            styles = getStyles(args.assets)
```
jonathanKingston, line 1621: Can we expose these on the class somehow? I'm trying to deprecate `args` lately.

src/features/click-to-load.js
```diff
@@ -22,6 +22,7 @@ const titleID = 'DuckDuckGoPrivacyEssentialsCTLElementTitle'
 // @see {getConfig}
 let config = null
 let sharedStrings = null
+let styles = getStyles()
```
jonathanKingston, line 25: Do we ever use this before `init`?

sammacbeth: No, but this prevents a typescript error...

jonathanKingston: Can we change the type to make it optional? I think that reduces the risk of unexpected side effects

kzar (approved):

jonathanKingston (approved):
Thanks :)

PR #394, opened 2023-04-07
Revert "Add Android messaging integration"

marcosholgado:
Reverts duckduckgo/content-scope-scripts#380

PR #450, opened 2023-04-20
Add Android messaging integration

joshliebe:
Reverts duckduckgo/content-scope-scripts#394

### DIFF
--- a/inject/android.js
+++ b/inject/android.js
@@ -2,8 +2,17 @@
  * @module Android integration
  * @category Content Scope Scripts Integrations
  */
-import { load, init } from '../src/content-scope-features.js'
+import { load, init, update } from '../src/content-scope-features.js'
 import { processConfig, isGloballyDisabled } from './../src/utils'
+
+const allowedMessages = [
+    'getClickToLoadState',
+    'getYouTubeVideoDetails',
+    'openShareFeedbackPage',
+    'setYoutubePreviewsEnabled',
+    'unblockClickToLoadContent',
+    'updateYouTubeCTLAddedFlag'
+]
 
 function initCode () {
     // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
@@ -16,10 +25,47 @@ function initCode () {
         platform: processedConfig.platform
     })
 
+    const messageSecret = processedConfig.messageSecret
+    // Receives messages from the platform
+    const messageCallback = processedConfig.messageCallback
+    // Sends messages to the platform
+    const messageInterface = processedConfig.messageInterface
+
+    const wrappedUpdate = ((providedSecret, ...args) => {
+        if (providedSecret === messageSecret) {
+            update(...args)
+        }
+    }).bind()
+
+    Object.defineProperty(window, messageCallback, {
+        value: wrappedUpdate
+    })
+
+    // @ts-ignore
+    const sendMessageToAndroid = window[messageInterface].process.bind(window[messageInterface])
+    delete window[messageInterface]
+
     init(processedConfig)
 
-    // Not supported:
-    // update(message)
+    window.addEventListener('sendMessageProxy' + messageSecret, event => {
+        event.stopImmediatePropagation()
+
+        if (!(event instanceof CustomEvent) || !event?.detail) {
+            return console.warn('no details in sendMessage proxy', event)
+        }
+
+        const messageType = event.detail?.messageType
+        if (!allowedMessages.includes(messageType)) {
+            return console.warn('Ignoring invalid sendMessage messageType', messageType)
+        }
+
+        const message = {
+            type: messageType,
+            options: event.detail?.options
+        }
+        const stringifiedArgs = JSON.stringify(message)
+        sendMessageToAndroid(stringifiedArgs, messageSecret)
+    })
 }
 
 initCode()
--- a/inject/android.js
+++ b/inject/android.js
@@ -35,13 +35,14 @@ function initCode () {
         if (providedSecret === messageSecret) {
             update(...args)
         }
+        // eslint-disable-next-line no-extra-bind
     }).bind(this)
 
     Object.defineProperty(window, messageCallback, {
         value: wrappedUpdate
     })
 
-    // @ts-expect-error
+    // @ts-expect-error - Android message interface defined elsewhere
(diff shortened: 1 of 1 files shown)


inject/android.js
```diff
@@ -16,10 +25,47 @@ function initCode () {
...
+        if (providedSecret === messageSecret) {
+            update(...args)
+        }
+    }).bind()
+
+    Object.defineProperty(window, messageCallback, {
+        value: wrappedUpdate
+    })
+
+    // @ts-ignore
```
jonathanKingston, line 44: Can you use ts-expect-error instead?

inject/android.js
```diff
@@ -16,10 +25,48 @@ function initCode () {
...
+    const messageCallback = processedConfig.messageCallback
+    // Sends messages to the platform
+    const messageInterface = processedConfig.messageInterface
+
+    const wrappedUpdate = ((providedSecret, ...args) => {
+        if (providedSecret === messageSecret) {
+            update(...args)
+        }
+        // eslint-disable-next-line no-extra-bind
+    }).bind(this)
```
joshliebe, line 39: @kzar I've suppressed this: `ESLint: The function binding is unnecessary`.
Let me know if this is the correct approach.

jonathanKingston: Just remove it, arrow functions can't be bound.

inject/android.js
```diff
@@ -35,13 +35,14 @@ function initCode () {
...
+        // eslint-disable-next-line no-extra-bind
     }).bind(this)
 
     Object.defineProperty(window, messageCallback, {
         value: wrappedUpdate
     })
 
-    // @ts-expect-error
+    // @ts-expect-error - Android message interface defined elsewhere
     const sendMessageToAndroid = window[messageInterface].process.bind(window[messageInterface])
```
jonathanKingston, line 46: Was this the issue why it was broken?

Whatever it was can we account for it not being present if we wish to land it in isolation?

shakyShane: I've been looking at this too, I think it's better to remove the global interface entirely, like this: 

```js
const originalHandler = window[messageInterface]
const sendMessageToAndroid = originalHandler.process.bind(originalHandler)
Reflect.deleteProperty(window, messageInterface) // doesn't throw

```

kzar: I don't understand @shakyShane , are you saying `Reflect.deleteProperty(window, messageInterface)` is better than `delete window[messageInterface]` here?

shakyShane: Ahh sorry @kzar I mis-read - I was going to make the point that you should remove the entire thing, not just process, but I can see you are doing that 👍🏻

regardless I think you need the reference to window.<interface> for the binding to live on after it's deleted, that's why I did the binding in 2 steps as shown above.

Might be wrong on that, I can verify :)

shakyShane: Updated my comment, just verified this approach on Android and it works :) sorry for the noise, I thought I'd encountered issues doing it this way in the past, but this looks fine in the simulator :)

inject/android.js
jonathanKingston: [See the note in the Spec](https://262.ecma-international.org/6.0/#sec-arrow-function-definitions-runtime-semantics-evaluation) _"An ArrowFunction does not define local bindings for arguments, super, this, or new.target. Any reference to arguments, super, this, or new.target within an ArrowFunction must resolve to a binding in a lexically enclosing environment."_

Even if you bind it's not possible, this is always the enclosed scope and the toString is never different:
```js
const other = {
  toString() {return 'meep'}
}
function beep() {
  const x = (() => {
    console.log('x', this)
  }).bind(other)
  console.log(1, x, x.toString(), x())
  const y = (() => {
    console.log('y', this)
  }).bind(other)
  console.log(2, y, y.toString(), y())
}
beep.prototype.toString = function () {
  return 'xxx'
}
beep()
```

> The bind here helps avoid wrappedUpdate.toString() revealing the messaging secret, which is worth doing.

Can you explain what you mean?

inject/android.js
jonathanKingston: Can we answer these:

- Was this the issue why it was broken (and was reverted)?
- Whatever it was can we account for it not being present if we wish to land it in isolation (to the Android changes)?

joshliebe: @jonathanKingston 
- Yes, it was looking for the interface which was not merged yet on the Android side
- I've since merged the Android messaging changes which means the interface is now there: https://github.com/duckduckgo/Android/pull/3083
- And I've run the action to check if this branch will cause any problems with the privacy tests: https://github.com/duckduckgo/Android/actions/runs/4758472939/jobs/8456577141
- So merging this is safe ✅

inject/android.js
kzar: > Can you explain what you mean?

If the website does `window[messageCallback].toString()` (after figuring out what `messageCallback` is), we don't want the value of `messageSecret` to be revealed. Using bind like I'm recommending avoids that:

```javascript
const example = ((providedSecret, ...args) => {
        if (providedSecret === messageSecret) {
            update(...args)
        }
        // eslint-disable-next-line no-extra-bind
    })
example.toString()

const wrappedExample = example.bind(this)
wrappedExample.toString()
```

inject/android.js
kzar: I think @jonathanKingston  has a point though, it wouldn't harm to first check `window[messageInterface]` exists before attempting to use it. How about something like this?

```javascript
let sendMessageToAndroid
if (Object.prototype.hasOwnProperty.call(window, messageInterface)) {
  sendMessageToAndroid = window[messageInterface].process.bind(window[messageInterface])
  delete window[messageInterface]
} else {
  sendMessageToAndroid = () => { console.error('Android messaging interface not available') }
}
```

inject/android.js
jonathanKingston: Ah! Cool thanks for explaining.

I think because bind is invalid you're essentially doing `.bind(undefined)` the lint still fails using that though 🥲

inject/android.js
joshliebe: I added that and it doesn't break the rest of the script now if the interface is not present 👍
I still needed to suppress `process` not existing on the window, so I made the explanation more explicit.

kzar (approved):
LGTM

jonathanKingston (approved):
LGTM thank you!

PR #452, opened 2023-04-20
Workaround declarativeNetRequest element collapsing bug

kzar:
There's a bug[1] with the declarativeNetRequest API that means that
after blocked iframes are collapsed (hidden) they sometimes can't be
uncollapsed (shown again) after they load successfully. That causes
Click to Load to get messed up for Chrome MV3 builds of the extension,
as the content is sometimes not shown to the user again after it has
been unblocked. The bug is triggered when the collapsed iframe is
removed from the DOM.

As a workaround, let's load a benign data URI in blocked iframes
before removing them from the DOM. That way they are uncollapsed, and
so if they are added back to the page they will be visible to the
user.

1 - https://crbug.com/1428971

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -485,6 +485,16 @@ function replaceTrackingElement (widget, trackingElement, placeholderElement) {
     ]
     elementToReplace.style.setProperty('display', 'none', 'important')
 
+    // When iframes are blocked by the declarativeNetRequest API, they are
+    // collapsed (hidden) automatically. Unfortunately however, there's a bug
+    // that stops them from being uncollapsed (shown again) if they are removed
+    // from the DOM after they are collapsed. As a workaround, have the iframe
+    // load a benign data URI, so that it's uncollapsed, before removing it from
+    // the DOM. See https://crbug.com/1428971
+    const originalSrc = elementToReplace.src
+    elementToReplace.src =
+        'data:text/plain;charset=utf-8;base64,' + btoa('https://crbug.com/1428971')
+
     // Add the placeholder element to the page.
     elementToReplace.parentElement.insertBefore(
         placeholderElement, elementToReplace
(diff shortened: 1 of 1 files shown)


src/features/click-to-load.js
```diff
@@ -484,6 +484,16 @@ function replaceTrackingElement (widget, trackingElement, placeholderElement) {
...
 
+    // When iframes are blocked by the declarativeNetRequest API, they are
+    // collapsed (hidden) automatically. Unfortunately however, there's a bug
+    // that stops them from being uncollapsed (shown again) if they are removed
+    // from the DOM after they are collapsed. As a workaround, have the iframe
+    // load a benign data URI, so that it's uncollapsed, before removing it from
+    // the DOM. See https://crbug.com/1428971
+    const originalSrc = elementToReplace.src
+    elementToReplace.src =
+        'data:text/plain;charset=utf-8;base64,aHR0cHM6Ly9jcmJ1Zy5jb20vMTQyODk3MQ=='
```
jonathanKingston, line 495: Encoding the bug number is a little weird, can we use `btoa`?

kzar: Encoding the bug link is nice, because it's short but provides context to someone who notices the "request" in the developer tools. Sure, we can use `btoa` if you like, done.

jonathanKingston: > Encoding the bug link is nice

What's the chance the user will see this?

kzar: > What's the chance the user will see this?

I assumed this was a rhetorical question, but after talking in MatterMost I realise it wasn't. I don't know how to answer this though, I'm not sure what the probability of a user viewing the network request panel for a website with a Click to Placeholder is.

jonathanKingston: I mean you're making an iframe text the Chrome bug, is there a chance a user would see it ever? (I don't care about the network panel at all).

I don't have the complete context of when the placeholder bug happens and if it impacts real users in general.

kzar: Ah right, got it. No the user can't ever see it in the page. We've already hidden the element at this point, which is on top of the element collapsing that happens on Chrome MV3.

kzar: Flow is something like hide -> maybe collapse + uncollapse by loading data URI -> remove from DOM -> unhide.

jonathanKingston (approved):

PR #457, opened 2023-04-21
Dispatch Click to Load ready event again when necessary

kzar:
The "ddg-ctp-ready" event is dispatched by the Click to Load
content-scope-script, to let the surrogate scripts know when the
feature is ready. Very occasionally that event is dispatched too
early, and so the surrogate script misses it. Workaround that by
dispatching the event again, if we see that a surrogate script
finished loading _after_ page load has finished.

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -1702,6 +1702,16 @@ export default class ClickToLoad extends ContentFeature {
         }
         await afterPageLoad
 
+        // On some websites, the "ddg-ctp-ready" event is occasionally
+        // dispatched too early, before the listener is ready to receive it.
+        // To counter that, catch "ddg-ctp-surrogate-load" events dispatched
+        // _after_ page, so the "ddg-ctp-ready" event can be dispatched again.
+        window.addEventListener(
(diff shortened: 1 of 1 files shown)


ladamski (commented):
Would this mean that `ddg-ctp-ready` sometimes fires twice, and that would result in multiple firings of `onClickToPlayReady()` in the surrogate?  

If this second path is more reliable would it make sense to only use that?  Or put a guard in `onClickToPlayReady()` to handle redundant calls?

kzar:
> Would this mean that `ddg-ctp-ready` sometimes fires twice, and that would result in multiple firings of `onClickToPlayReady()` in the surrogate?

Our event listener for `ddg-ctp-ready` uses the `once: true` option for that reason, to avoid it catching the event twice.

> If this second path is more reliable would it make sense to only use that?

Yea, it might make sense to do that in the future. Thing is it would add quite a bit of complexity, keeping track of when the various surrogate scripts are ready and dispatching the relevant events only after that. I'd like to avoid it if we can though, and so far I've only ever see this happen very occasionally on one website.

PR #471, opened 2023-04-25
Add support for set and get attribute with a NS

### LINKED ISSUE #462: Script sources set using setAttributeNS(null, 'src', <url>) are ignored
For example, the following will try to load a script from `https://<hostname>/undefined` rather than log.js:
```html
<!DOCTYPE html>
<html>
    <head>
        <meta charset="utf-8">
        <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.6.3/jquery.js"></script>
        <script>
$(() => {
  var s = document.createElement("script");
  s.setAttributeNS(null, "src", "log.js"); // log.js just calls console.log("ok") to prove it's working
  document.body.appendChild(s);
});
        </script>
    </head>
    <body>
    </body>
</html>
```

Skyward Qmlativ sites use setAttributeNS when cloning scripts, which breaks a lot of pages.  It's easy to work around by checking for the null namespace and calling setAttribute(), but I wanted to make sure you knew about it as well.

jonathanKingston:
Fixes #462

Captures calls with `null` and passes them through the trapped function.

### DIFF
--- a/integration-test/test-pages/runtime-checks/pages/basic-run.html
+++ b/integration-test/test-pages/runtime-checks/pages/basic-run.html
@@ -176,6 +176,32 @@
             ];
         });
 
+        test('Support namespaces', async () => {
+            const scriptElement = document.createElement('script');
+            scriptElement.id = 'scriptyNS';
+            scriptElement.setAttribute('type', 'application/javascript');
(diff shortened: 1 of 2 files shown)


dharb (approved):
👍

PR #479, opened 2023-04-29
priv info

jonathanKingston:
Adds some privileged info so we can send debugging reports on search for JSE.


jonathanKingston:
Closing: inactive since 2023, a stale WIP/experiment that no longer applies cleanly to `main` (large divergence). Clearing from the open queue — please reopen or re-create from current `main` if still wanted.

PR #480, opened 2023-04-29
Handle WebGL reading of pixels

jonathanKingston:
The 70 lines of shader noise is all boiler plate. I need to validate it some more.

Largely this is doing the same as all the other protections we have, the only slight difference is we're reusing the page context webgl to push the content changes into.

I think maybe we should create another offscreen canvas to then make the changes to other wise it will suffer from re-enterance issues.


jonathanKingston:
Closing: inactive since 2023, a stale WIP/experiment that no longer applies cleanly to `main` (large divergence). Clearing from the open queue — please reopen or re-create from current `main` if still wanted.

PR #506, opened 2023-05-11
Support arrays of domains for patching settings

jonathanKingston:
3. of https://app.asana.com/0/1200890834746050/1204550881134809/f

This adds support for array matching of domain keys to reduce overheads if we wanted a large number of domains to behave a certain way.


GuiltyDolphin (approved):
The logic looks fine to me.

PR #508, opened 2023-05-11
Update styletag injection to use <style> for platforms other than firefox

dharb:
In order to bypass `style-src` CSP restrictions in Firefox, we use a link element. However this creates a CSP issue on other platforms on websites that have `style-src 'unsafe-inline'` in their CSP, which is relatively common: we're not able to inject a link element. This PR updates our injection technique to only use link elements in Firefox, and normal style elements elsewhere.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -23,16 +23,26 @@ export function getInjectionElement () {
     return document.head || document.documentElement
 }
 
+
+// Tests don't define this variable so fallback to behave like chrome
+const hasMozProxies = typeof mozProxies !== 'undefined' ? mozProxies : false
+
 /**
  * Creates a script element with the given code to avoid Firefox CSP restrictions.
  * @param {string} css
  * @returns {HTMLLinkElement}
  */
 export function createStyleElement (css) {
-    const style = document.createElement('link')
-    style.href = 'data:text/css,' + encodeURIComponent(css)
-    style.setAttribute('rel', 'stylesheet')
-    style.setAttribute('type', 'text/css')
+    let style
+    if (!!hasMozProxies) {
+        const style = document.createElement('link')
+        style.href = 'data:text/css,' + encodeURIComponent(css)
+        style.setAttribute('rel', 'stylesheet')
+        style.setAttribute('type', 'text/css')
+    } else {
+        style = document.createElement('style')
+        style.innerText = css
+    }
     return style
 }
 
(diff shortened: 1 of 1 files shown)


src/utils.js
```diff
@@ -23,16 +23,26 @@ export function getInjectionElement () {
...
  * @param {string} css
  * @returns {HTMLLinkElement}
  */
 export function createStyleElement (css) {
-    const style = document.createElement('link')
-    style.href = 'data:text/css,' + encodeURIComponent(css)
-    style.setAttribute('rel', 'stylesheet')
-    style.setAttribute('type', 'text/css')
+    let style
+    if (!!hasMozProxies) {
```
jonathanKingston, line 37: This is a boolean so we're good to just do the following.

```suggestion
    if (hasMozProxies) {
```

src/utils.js
```diff
@@ -23,16 +23,25 @@ export function getInjectionElement () {
     return document.head || document.documentElement
 }
 
+// Tests don't define this variable so fallback to behave like chrome
+const hasMozProxies = typeof mozProxies !== 'undefined' ? mozProxies : false
+
 /**
  * Creates a script element with the given code to avoid Firefox CSP restrictions.
  * @param {string} css
  * @returns {HTMLLinkElement}
```
jonathanKingston, line 32:
```suggestion
 * @returns {HTMLLinkElement | HTMLStyleElement}
```

jonathanKingston, line 32:
```suggestion
 * @returns {HTMLLinkElement | HTMLStyleElement}
```

src/utils.js
```diff
@@ -23,16 +23,25 @@ export function getInjectionElement () {
...
+ * @returns {HTMLLinkElement | HTMLStyleElement}
  */
 export function createStyleElement (css) {
-    const style = document.createElement('link')
-    style.href = 'data:text/css,' + encodeURIComponent(css)
-    style.setAttribute('rel', 'stylesheet')
-    style.setAttribute('type', 'text/css')
+    let style
+    if (hasMozProxies) {
+        const style = document.createElement('link')
```
jonathanKingston, line 37:
```suggestion
    let style
    if (hasMozProxies) {
       style = document.createElement('link')
```

dharb: 🤦‍♂️

jonathanKingston (approved):

PR #511, opened 2023-05-12
Add replaceElement scheme and monitorProperties

jonathanKingston:


dharb (approved):
Tested and this seems to work quite well - I was able to resolve the adobe/logrocket issues entirely by enabling `replaceElement`.

integration-test/test-pages/runtime-checks/config/replace-element.json
```diff
@@ -0,0 +1,21 @@
...
+            "state": "enabled",
+            "exceptions": [],
+            "settings": {
+                "taintCheck": "enabled",
+                "matchAllDomains": "enabled",
+                "matchAllStackDomains": "enabled",
+                "overloadInstanceOf": "enabled",
+                "elementRemovalTimeout": 100,
+                "injectGlobalStyles": "enabled",
+                "replaceElement": "enabled",
```
dharb, line 13: Should we add `monitorProperties` here as well? I realize probably not entirely necessary since it's default enabled, but for consistency.

dharb (approved):
Looks good and seems to work well for me. Appreciate the thorough tests.

integration-test/test-pages/runtime-checks/config/replace-element.json
jonathanKingston: The other test relies on the default here too. I can clean this up after once we remove the algo.

PR #518, opened 2023-05-15
Change interface check to HTMLDocument again

jonathanKingston:
https://app.asana.com/0/1200277586140538/1204601197399295/f

### DIFF
--- a/src/content-scope-features.js
+++ b/src/content-scope-features.js
@@ -8,7 +8,8 @@ import platformFeatures from 'ddg:platformFeatures'
 function shouldRun () {
     // don't inject into non-HTML documents (such as XML documents)
     // but do inject into XHTML documents
-    if (document instanceof Document === false && (
+    // Should check HTMLDocument as Document is an alias for XMLDocument also.
+    if (document instanceof HTMLDocument === false && (
         document instanceof XMLDocument === false ||
         document.createElement('div') instanceof HTMLDivElement === false
     )) {


src/content-scope-features.js
```diff
@@ -8,7 +8,8 @@ import platformFeatures from 'ddg:platformFeatures'
 function shouldRun () {
     // don't inject into non-HTML documents (such as XML documents)
     // but do inject into XHTML documents
-    if (document instanceof Document === false && (
+    // Should check HTMLDocument as Document is an alias for XMLDocument also.
+    if (document instanceof HTMLDocument === false && (
         document instanceof XMLDocument === false ||
```
sammacbeth, line 13: Isn't `document instanceof XMLDocument` === `true` for XML docs, and `false` for HTML?

jonathanKingston: Yes that's correct.

We adopted this logic from elsewhere. It's pretty unreadable and should be refactored but I was restoring it back to what it previously was recently.

My understanding is:

- `document instanceof HTMLDocument` can be false sometimes (due to weird page setup overwriting HTMLDocument perhaps?)
- We always don't run on XMLDocument's
- If this is still potentially (X)HTML we check if div's match `HTMLDivElement`

The reason this started failing is `document instanceof HTMLDocument` was changed to `Document` which matches both `XMLDocument` and `HTMLDocument`.

TL;DR I think we should write some tests here and clean up, but I'm a bit strapped for time so I suggest we just write up an issue to fix?

sammacbeth (approved):

src/content-scope-features.js
jonathanKingston: Task: https://app.asana.com/0/1201614831475344/1204609715941102/f

PR #521, opened 2023-05-16
Disable script overloads for non scripts

jonathanKingston:
Ignores modifying script content of non scripts:
https://app.asana.com/0/0/1204613163970758/f

### DIFF
--- a/src/features/runtime-checks.js
+++ b/src/features/runtime-checks.js
@@ -44,6 +44,25 @@ const defaultElementMethods = {
 }
 const supportedTrustedTypes = 'TrustedScriptURL' in window
 
+const jsMimeTypes = [
+    'text/javascript',
+    'text/ecmascript',
+    'application/javascript',
+    'application/ecmascript',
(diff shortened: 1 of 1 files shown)


src/features/runtime-checks.js
```diff
@@ -44,6 +44,25 @@ const defaultElementMethods = {
 }
 const supportedTrustedTypes = 'TrustedScriptURL' in window
 
+const jsMimeTypes = [
+    'text/javascript',
```
jonathanKingston, line 48: https://searchfox.org/mozilla-central/source/dom/base/nsContentUtils.cpp#7638-7654

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/common/shared_storage/module_script_downloader_unittest.cc;l=205-221?q=text%2Fjavascript1.0%20language:cpp

dharb (approved):
👍 thanks!

dharb (approved):

PR #525, opened 2023-05-17
Click to load: mobile block placeholder

franfaccin:


src/features/click-to-load/assets/ctl-placeholder-block.css
```diff
@@ -0,0 +1,336 @@
+* {
+    font-family: DuckDuckGoPrivacyEssentials, system, -apple-system, system-ui, BlinkMacSystemFont, 'Segoe UI', Roboto,
+        Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
+}
```
GioSensation, line 4: I think the shadow DOM will still inherit some properties from the host page. This has bitten us in the past, where text would render with the wrong alignment or font style. Not sure if this is already taken care of elsewhere, but it may be worth resetting the shadow properties. Here's a list of the inherited properties: https://stackoverflow.com/questions/5612302/which-css-properties-are-inherited#5612360. I haven't vetted it, just a quick search.

GioSensation (commented):
This is looking great 💪. The code seems well structured and easy to follow. I haven't run the code and I'm not familiar with the codebase so half of my points are likely moot 🙃. Anyway, a few comments from me, mostly minor stuff 🙂.

src/features/click-to-load/components/ctl-placeholder-blocked.js
```diff
@@ -0,0 +1,187 @@
...
+        if (useSlimCard) {
+            container.classList.add('slim-card')
+        }
+        if (withFeedback) {
+            container.classList.add('with-feedback-link')
+        }
+
+        const learnMoreLink = this.createLearnMoreLink()
+
+        container.innerHTML = html`
```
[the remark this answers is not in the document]
kzar: Good point, for example the localised strings?

src/features/click-to-load/components/ctl-placeholder-blocked.js
```diff
@@ -0,0 +1,187 @@
+import { html } from '../../../dom-utils'
+import css from '../assets/ctl-placeholder-block.css'
+import { logoImg as daxImg } from '../ctl-assets'
+
+export class DDGCtlPlaceholderBlocked extends HTMLElement {
```
kzar, line 5: Could you add a comment to explain what this class is for? I know it's kind of in the name, but I'm thinking if this is only for non-preview placeholders and only for mobile it would be nice to explain.

kzar: Or maybe just add some text to the start of the comment for the constructor, to explain how it's used and what it's used for?

src/features/click-to-load.js
```diff
@@ -597,11 +618,38 @@ function replaceYouTubeCTL (trackingElement, widget) {
...
         widget.autoplay = false
         const oldPlaceholder = widget.placeholderElement
-        const { blockingDialog, shadowRoot } = createYouTubeBlockingDialog(trackingElement, widget)
-        resizeElementToMatch(oldPlaceholder || trackingElement, blockingDialog)
-        replaceTrackingElement(widget, trackingElement, blockingDialog)
-        showExtraUnblockIfShortPlaceholder(shadowRoot, blockingDialog)
-        hideInfoTextIfNarrowPlaceholder(shadowRoot, blockingDialog, 460)
+
+        if (isMobileApp) {
+            const mobileBlockedPlaceholder = new DDGCtlPlaceholderBlocked({
```
kzar, line 623: It's cool how you've set this up, but it took me a few reads to grok. I think a comment to explain usage for the constructor would help as I mentioned elsewhere. Also though, perhaps ending the class name with "Element" would help make it clearer that this class inherits from HTMLElement too?

src/features/click-to-load/components/ctl-placeholder-blocked.js
franfaccin: done 👍

src/features/click-to-load/components/ctl-placeholder-blocked.js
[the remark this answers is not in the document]
franfaccin: Good point. We don't get external strings, they're all translations that come from Smartling. But for the YouTube preview placeholder, we will get a title and an image externally, so I'll keeping this in mind for my next PR.

Either way, the [templating that Shane introduced](https://github.com/duckduckgo/content-scope-scripts/blob/main/src/dom-utils.js) seems to sanitize the interpolated strings. Here is an example of me passing a string with a button for the placeholder title in the constructor, and that got sanitized.

image

src/features/click-to-load/components/ctl-placeholder-blocked.js
```diff
@@ -0,0 +1,187 @@
...
+     * @returns {HTMLDivElement}
+     */
+    createShareFeedbackLink = () => {
+        const { sharedStrings } = this.params
+
+        const container = document.createElement('div')
+        container.classList.add('ddg-ctl-feedback-row')
+
+        container.innerHTML = html`
+            <a class="ddg-ctl-feedback-link" target="_blank" href="#">${sharedStrings.shareFeedback}</a>
```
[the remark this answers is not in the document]
GioSensation: If there is navigation an anchor tag is actually better. Not a blocker, though. Sorry for the noise.

src/features/click-to-load/components/ctl-placeholder-blocked.js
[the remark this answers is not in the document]
shakyShane: yeah, we added the `html` template tag helper exactly for this scenario 👍🏻

src/features/click-to-load/assets/ctl-placeholder-block.css
```diff
@@ -0,0 +1,329 @@
...
+    margin: 0;
+}
+.DuckDuckGoButton.tertiary {
+    color: rgba(0, 0, 0, 0.84);
+    background-color: transparent;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    padding: 0px 16px;
+    border: 1px solid rgba(0, 0, 0, 0.12);
```
shakyShane, line 55: Let's move all values that match design tokens in Figma into a `variables.css` file as CSS custom properties. Then you can re-use that file in this + any new components whilst sharing the values.

src/features/click-to-load/components/ctl-placeholder-blocked.js
```diff
@@ -0,0 +1,281 @@
...
+                </div>
+                <div class="ddg-ctl-placeholder-card-body">
+                    <div class="ddg-ctl-placeholder-card-body-text">${body} ${learnMoreLink}</div>
+                    <button class="DuckDuckGoButton tertiary ddg-ctl-unblock-btn" type="button">
+                        <div>${unblockBtnText}</div>
+                    </button>
+                </div>
+                ${withToggle
+        ? html`<div class="ddg-ctl-placeholder-card-footer">${this.createToggleButton()}</div> `
+        : ''}
```
shakyShane, line 138: The templating system here makes it difficult to do nice ternaries inline like this, I would extract this (and any others ) to in-scope variables instead, just to prevent having to deal with these formatting issues :)

shakyShane (changes requested):
@franfaccin great stuff! 

I tried to reflect everything we spoke about in our call - let me know if I missed anything or if you have any questions :)

src/features/click-to-load/components/ctl-placeholder-blocked.js
```diff
@@ -0,0 +1,281 @@
...
+        >
+            ${sharedStrings.learnMore}
+        </a>`
+    }
+
+    /**
+     * Creates a Feedback Link container row
+     * @returns {HTMLDivElement}
+     */
+    createShareFeedbackLink = () => {
```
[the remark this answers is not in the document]
franfaccin: Firefox is still not happy after all the changes :/  I'm adding a comment to the first arrow function with context. We can try to debug it more as a follow up.

image

franfaccin:
This is ready for review again :)

shakyShane (approved):

PR #526, opened 2023-05-18
Fix scope overload access issue

jonathanKingston:
Fixes an issue where we're getting 'Uncaught TypeError: Illegal invocation' from the generated JS proxy.


dharb (approved):
Looks good and works well, thanks!

PR #527, opened 2023-05-19
Harmful API protections

muodov:
This PR adds (fingerprint) protections from some privacy-harmful APIs
Project link: https://app.asana.com/0/72649045549333/1202019585755193/f

- Large diff: I tried to keep the commits clean, consider reviewing commit-by-commit

---
  - https://app.asana.com/0/0/1204646827330362


src/features/harmful-apis.js
```diff
@@ -0,0 +1,443 @@
...
+        }
+        const eventsToBlock = settings.filterEvents || []
+        if (eventsToBlock.length > 0) {
+            for (const eventName of eventsToBlock) {
+                const dom0HandlerName = `on${eventName}`
+                if (dom0HandlerName in globalThis) {
+                    delete globalThis[dom0HandlerName]
+                }
+            }
+            wrapMethod(EventTarget.prototype, 'addEventListener', function (nativeImpl, type, ...restArgs) {
```
jonathanKingston, line 73: Double wrapping things has been an issue before for us. I would prefer we didn't but also I don't really want to refactor mid PR.

Are you at all concerned by it? (performance and page detectability wise).

This API is called a lot so we should add a lot more scrutiny to the review here.

src/features/harmful-apis.js
```diff
@@ -0,0 +1,443 @@
...
+                if (eventsToBlock.includes(type) && this === globalThis) {
+                    return
+                }
+                return nativeImpl.call(this, type, ...restArgs)
```
jonathanKingston, line 77: Elsewhere we've used `Reflect.apply` this prevents the page overloading `.call` by using the object internals.

src/features/harmful-apis.js
muodov: Could you elaborate what kind of issues you had with double wrapping?
Performance is not my biggest concern because we are adding overhead on `addEventListener()` itself, not when the handler is called.
Page detectability is just the same as any other wrapping we do: there are certainly ways to detect this, but you'll need to jump through some hoops.

src/features/harmful-apis.js
muodov: I've changed it

src/features/harmful-apis.js
jonathanKingston: document.cookie was much more complex and had async bits.... so I think we can ignore. It just gives me the ick but as I can't really articulate why it's fine :)

muodov: I feel you 😅 It is a very common API indeed, so we gotta be careful, but I also couldn't think of any critical flaw of this approach

integration-test/playwright/harmful-apis.spec.js
```diff
@@ -0,0 +1,140 @@
...
+            return new Promise(resolve => {
+                window.addEventListener('results-ready', () => {
+                    // @ts-expect-error - this is added by the test framework
+                    resolve(window.results)
+                })
+            })
+        })
+        await this.page.getByTestId('render-results').click()
+        return await resultsPromise
+    }
```
shakyShane, line 67: can you separate these actions+assertions from the 'enabled' call (which is just the setup) - just inline it inside the `test` above, or create a separate method on this class. 

The examples in the duckplayer.spec.js are the format we want to move forward with, but since this is a single test I don't think it's an issue to just inline the actions/assertions into the `test` <- that's what I did in the windows-permissions tests

playwright.config.js
```diff
@@ -5,7 +5,8 @@ export default defineConfig({
         {
             name: 'windows',
             testMatch: [
-                'integration-test/playwright/duckplayer.spec.js'
+                'integration-test/playwright/duckplayer.spec.js',
+                'integration-test/playwright/harmful-apis.spec.js'
```
shakyShane, line 9: Ahh! this made me realise - we accidentally excluded the original tests for windows permissions in this commit: 02e53bb9dc9c92e42a86a6aebaa0cc32ada1c002 oops! I'll create a separate PR for that - but it makes me think if there's some redundant overlap? 

Perhaps on your branch you could re-enable the windows-permissions tests and evaluate if your work makes any of it obsolete?

src/features/harmful-apis.js
```diff
@@ -0,0 +1,450 @@
...
+        this.blockWindowPlacementApi(this.getFeatureSetting('WindowPlacement'))
+        this.blockWebBluetoothApi(this.getFeatureSetting('WebBluetooth'))
+        this.blockWebUsbApi(this.getFeatureSetting('WebUsb'))
+        this.blockWebSerialApi(this.getFeatureSetting('WebSerial'))
+        this.blockWebHidApi(this.getFeatureSetting('WebHid'))
+        this.blockWebMidiApi(this.getFeatureSetting('WebMidi'))
+        this.removeIdleDetectionApi(this.getFeatureSetting('IdleDetection'))
+        this.removeWebNfcApi(this.getFeatureSetting('WebNfc'))
+        this.filterStorageManagerApi(this.getFeatureSetting('StorageManager'))
+    }
```
shakyShane, line 55: `this.getFeatureSetting()` can return undefined - but all of your methods expect to be able to check 'settings.protected'

Is this because you're going to always supply defaults in the config? Even if so, until we have schema verification + runtime checking I'd be tempted to convert everything to 

```js
if (!settings?.protected) {

}
```

Feel free to ignore if there's a reason this wouldn't be needed

shakyShane (changes requested):

src/features/harmful-apis.js
muodov: Right, the [default config](https://github.com/duckduckgo/privacy-configuration/blob/harmful-apis/features/harmful-apis.json) is supposed to always have these keys. If I understand correctly, this will always be loaded, right?
I think your suggestion makes sense though, I'll change it.

playwright.config.js
muodov: Yeah, I did actually move some code from the windows permissions in this PR. Re-enabled the test again, it still works well

shakyShane (approved):
Thanks! 

All looking good to me now :)

PR #528, opened 2023-05-22
Fix issue with inline parent elements

franfaccin:
Fix issues related to breakage reports for YouTube CTL.

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -617,7 +617,11 @@ function showExtraUnblockIfShortPlaceholder (shadowRoot, placeholder) {
     if (!placeholder.parentElement) {
         return
     }
-
+    const parentStyles = window.getComputedStyle(placeholder.parentElement)
+    // Inline elements, like span or p, don't have a set height that we can trust.
+    if (parentStyles.display === 'inline') {
+        return
+    }
     const { height: placeholderHeight } = placeholder.getBoundingClientRect()
     const { height: parentHeight } = placeholder.parentElement.getBoundingClientRect()
 
@@ -634,7 +638,7 @@ function showExtraUnblockIfShortPlaceholder (shadowRoot, placeholder) {
         /** @type {HTMLElement?} */
         const innerDiv = shadowRoot.querySelector('.DuckDuckGoSocialContainer')
         if (innerDiv) {
-            innerDiv.style.minHeight = ''
+            innerDiv.style.minHeight = 'initial'
             innerDiv.style.maxHeight = parentHeight + 'px'
             innerDiv.style.overflow = 'hidden'
         }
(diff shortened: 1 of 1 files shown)


src/features/click-to-load.js
```diff
@@ -617,7 +617,11 @@ function showExtraUnblockIfShortPlaceholder (shadowRoot, placeholder) {
     if (!placeholder.parentElement) {
         return
     }
-
+    const parentStyles = window.getComputedStyle(placeholder.parentElement)
+    // Inline elements, like span or p, don't have a set height that we can trust.
+    if (parentStyles.display === 'inline') {
```
jonathanKingston, line 622: What happens when there's nested spans?
Do we need to traverse to the nearest non inline element for height instead?

src/features/click-to-load.js
```diff
@@ -831,6 +835,21 @@ function resizeElementToMatch (sourceElement, targetElement) {
     for (const key of stylesToCopy) {
         targetElement.style[key] = computedStyle[key]
     }
+    // Sometimes, after copying 'top', 'bottom', 'left', 'right' from an absolute
+    // positioned element, the 'inset' property is automatically set to a value
+    // trying to replicate that positioning. But that often is not accurate, so we
+    // make sure to reset it back to 'auto' here.
+    targetElement.style.inset = 'auto'
```
kzar, line 842: As mentioned in MatterMost, this `inset = 'auto'` line is causing placeholders to be incorrectly positioned on a few websites. Suggest we just remove this for now.

src/features/click-to-load.js
kzar: Yea, this is a fair point :+1: .

I have tested quite a bit today though and Fran's changes make the placeholders display far more reliably. Suggest we go with this for now, and expand to cover edge-cases like that if it proves to be an issue in practice.

kzar: (Fran, in the mean time perhaps add a comment to note Jonathan's point? It might be useful to expand the logic to consider grandparent etc elements in the future.)

franfaccin: I was also thinking about that too, @jonathanKingston, although it might have other implications. As we go up the DOM tree, there are more elements being involved and the styles might not be as straightforward to verify. I think we can start with this and observe other reports for examples where this is not enough.

Either way, I'll add a comment so we know what to address if necessary.

src/features/click-to-load.js
franfaccin: 👍

src/features/click-to-load.js
franfaccin: done

kzar (approved):
LGTM

PR #535, opened 2023-05-24
Wrapper utils

muodov:
This adds some utility functions for overriding native descriptors

---
  - https://app.asana.com/0/0/1204646827330356


src/features/fingerprinting-hardware.js
```diff
@@ -1,28 +1,19 @@
-import { overrideProperty } from '../utils'
 import ContentFeature from '../content-feature'
+import { wrapProperty } from '../wrapper-utils'
 
 export default class FingerprintingHardware extends ContentFeature {
     init () {
-        const Navigator = globalThis.Navigator
-        const navigator = globalThis.navigator
-
-        overrideProperty('keyboard', {
```
jonathanKingston, line 9: I don't really understand the move from `overrideProperty` to wrap other than it's arguments are different and the API is different (this makes the diff much more noisy than it probably needs to be).

Perhaps there's more justification in the work you're doing, but as it stands I would keep the same function signature for all this code.

Can we rename it back and keep the same args, what's the real change this `toString` that you mentioned yesterday?

src/wrapper-utils.js
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/**
+ * Wrap a method descriptor. Only for function properties. For data properties, use wrapProperty(). For constructors, use wrapConstructor().
+ * @param {any} object - object or string object path within the global object
+ * @param {string} propertyName
+ * @param {(originalFn, ...args) => any } wrapperFn - wrapper function receives the original function as the first argument
+ * @returns {PropertyDescriptor|undefined} original property descriptor, or undefined if it's not found
+ */
+export function wrapMethod (object, propertyName, wrapperFn) {
```
jonathanKingston, line 112: This method and wrapProperty all dead code as it stands. I would keep it in the PR that needs it.

README.md
```diff
@@ -67,11 +67,40 @@ In the built output you will see these dramatic differences in the bundled code
 
 ### Features scope injection utilities
 
-To handle the difference in scope injection we expose multiple utilities which behave differently per browser in src/utils.js. for Firefox the code exposed handles [xrays correctly](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Sharing_objects_with_page_scripts) without needing the features to be authored differently.
```
jonathanKingston, line 70: We should perhaps move all the DDGProxy stuff also then we can remove utils.js from this comment. (Can be a follow up, just thinking outloud)

src/wrapper-utils.js
```diff
@@ -0,0 +1,191 @@
...
+    newFn.toString = function () {
+        if (this === newFn) {
+            return functionToString.call(origFn)
+        } else {
+            return functionToString.call(this)
+        }
+    }
+}
+
+function getObjectFromPath (objPath, base = globalObj) {
```
jonathanKingston, line 52: I think something like this might already exist, perhaps in runtimeChecks? (I feel like I've wrote this 50 times in the last month, so maybe we should move it over if there is one)

src/wrapper-utils.js
```diff
@@ -0,0 +1,191 @@
...
+/**
+ * add a fake toString() method to a wrapper function to resemble the original function
+ * @param {*} newFn
+ * @param {*} origFn
+ */
+function wrapToString (newFn, origFn) {
+    if (!newFn || !origFn) {
+        return
+    }
+    newFn.toString = function () {
```
jonathanKingston, line 43: I think we had tests for this with one disabled for `toString.toString()` can you check if we can get that enabled?

jonathanKingston (dismissed):
I think this looks good overall, I've not had chance to look deeply into it.

Ping me over the other changes, I think some of this patch probably belongs there so it can be reviewed with it (wrapConstructor etc)

Also I think we could continue to use the overrideProperty signature as it is and perhaps support a different code path. That would reduce the risk of this change perhaps; we can always refactor it away later.

src/features/fingerprinting-hardware.js
muodov: The idea was to replace the existing `overrideProperty()` with something more generic. The idea is that we should use `wrapProperty()` every time when we augment an existing descriptor. `defineProperty()` should be only used when we want to define a new property (for example in webcompat scripts).

The new signature allows to use `wrapProperty()` in many more cases. For example, in cookie protection we're overriding `document.cookie` using `defineProperty()` because `overrideProperty()` only supports static values. On top of that, we should actually be overriding `Document.prototype.cookie` instead. And `wrapProperty()` would make sure that you're not changing the type of the descriptor as well.

`overrideProperty()` and `wrapProperty()` have very similar purpose, so it doesn't make sense to keep them both. The latter is more flexible, and the signature is closer to that of `defineProperty()`.

If you're worried about conflicts with the parallel branches, I can include these changes completely in https://github.com/duckduckgo/content-scope-scripts/pull/527, since you're the reviewer there anyways. Would this help?

src/wrapper-utils.js
muodov: The closest I could find is https://github.com/duckduckgo/content-scope-scripts/blob/main/src/features/runtime-checks/script-overload.js#L216, but it's not quite the same. Is there another function I can reuse?

src/wrapper-utils.js
jonathanKingston: I think this will also need to be exported for mozilla code too.

src/wrapper-utils.js
jonathanKingston: Yeah maybe not then, ignore :)

src/wrapper-utils.js
muodov: I tested it on Firefox, and it seems like it ignores custom `toString` on descriptor functions anyway, so it works either way.

src/wrapper-utils.js
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * add a fake toString() method to a wrapper function to resemble the original function
+ * @param {*} newFn
+ * @param {*} origFn
+ */
+function wrapToString (newFn, origFn) {
+    if (typeof newFn !== 'function' || typeof origFn !== 'function') {
+        return
+    }
+    newFn.toString = function () {
```
jonathanKingston, line 68: I think `x.toString.toString()` will fail, we've had reports for this in the past.

We should probably raise an issue to track.

src/wrapper-utils.js
```diff
@@ -0,0 +1,131 @@
...
+ * @param {Partial<PropertyDescriptor>} descriptor
+ * @returns {PropertyDescriptor|undefined} original property descriptor, or undefined if it's not found
+ */
+export function wrapProperty (object, propertyName, descriptor) {
+    if (!object) {
+        return
+    }
+    if (typeof object === 'string') {
+        object = getObjectFromPath(object)
+    }
```
jonathanKingston, line 107: I think this is dead code even in your https://github.com/duckduckgo/content-scope-scripts/pull/527 patch, can we remove it along with the method too?

src/wrapper-utils.js
```diff
@@ -0,0 +1,131 @@
+/* global cloneInto, exportFunction, mozProxies */
+// Tests don't define this variable so fallback to behave like chrome
+const hasMozProxies = typeof mozProxies !== 'undefined' ? mozProxies : false
+const globalObj = typeof window === 'undefined' ? globalThis : window
+const getOwnPropertyDescriptor = Object.getOwnPropertyDescriptor
+const functionToString = Function.prototype.toString
+const objectKeys = Object.keys
```
jonathanKingston, line 7: These probably belong more in src/captured-globals.js which just landed (I assume you're capturing them to prevent extension).

You might need to capture the descriptors there if you **need** them. However I think it's OK to not be perfect if it's only ever internally used like we are here.

We'd just have:
```suggestion
import { Object, Function } from './captured-globals.js'
```

src/wrapper-utils.js
```diff
@@ -0,0 +1,131 @@
...
+const NativeProxy = Proxy
```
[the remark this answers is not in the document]
muodov: good catch, I removed it

src/wrapper-utils.js
muodov: As discussed, the current implementation in captured-globals has some limitations, so we'll address it in a separate task.

src/wrapper-utils.js
muodov: 👍 I removed it

jonathanKingston (approved):

PR #539, opened 2023-05-25
Fix CTL default width

franfaccin:


kzar (commented):
I was wondering:
1. Could we just use `Math.min(attrSettings.default, window.innerWidth)` for the width?
2. How come we only need the `parentInnerWidth` logic if the default width is too large? Isn't the new logic better in all situations?

franfaccin:
> I was wondering:
> 
> 1. Could we just use `Math.min(attrSettings.default, window.innerWidth)` for the width?
> 2. How come we only need the `parentInnerWidth` logic if the default width is too large? Isn't the new logic better in all situations?

@kzar:
1. There are 3 possible values to be used: default, window width and parent width. Parent width is the most tricky since it can be an inline element and cause issues. So I can't compare all of them straight away with `Math.min()`. I had to go by process of elimination to find the best value to be used.
2. Thinking on desktop size, unless the parent has a defined width, we wouldn't want to set the Facebook iframe to be the whole width of the page by default. So having the default size is still a valid choice. Perhaps what could be changed is that we first try to use the the parent inner width size, and if that's not defined, we do the `Math.min(attrSettings.default, window.innerWidth)` like you suggested. WDTY?

src/features/click-to-load.js
```diff
@@ -110,7 +110,30 @@ class DuckWidget {
...
+                    // should not be bigger than the space available in the parent element. There is no straightforward way of
+                    // doing this. We need to get the parent's .clientWidth and remove the paddings size from it.
+                    const parentInnerWidth = parentElement && parentStyles && parentStyles.display !== 'inline'
+                        ? parentElement.clientWidth - parseFloat(parentStyles.paddingLeft) - parseFloat(parentStyles.paddingRight)
+                        : null
+
+                    if (parentInnerWidth && parentInnerWidth < windowWidth) {
+                        value = parentInnerWidth.toString()
+                    } else {
+                        value = window.innerWidth.toString()
```
kzar, line 132: Nit:
```suggestion
                        value = windowWidth.toString()
```

src/features/click-to-load.js
```diff
@@ -110,7 +110,30 @@ class DuckWidget {
                     // with a light version, replace with full version.
                     this.clickAction.type = 'allowFull'
                 }
-                value = attrSettings.default
+                // Our default value for width is often greater than the window size of smaller
+                // screens (ie mobile). When it is, use either the parent width (if reliable)
+                // or the window width instead.
+                if (attrName === 'data-width' && attrSettings.default > window.innerWidth) {
+                    const windowWidth = window.innerWidth
+                    const parentElement = this.originalElement.parentElement
```
kzar, line 118: Nit:
```suggestion
                    const { parentElement } = this.originalElement
```

src/features/click-to-load.js
```diff
@@ -110,7 +110,30 @@ class DuckWidget {
...
+                    const parentStyles = parentElement
+                        ? window.getComputedStyle(parentElement)
+                        : null
+                    // We want to calculate the inner width of the parent element as the iframe, when added back,
+                    // should not be bigger than the space available in the parent element. There is no straightforward way of
+                    // doing this. We need to get the parent's .clientWidth and remove the paddings size from it.
+                    const parentInnerWidth = parentElement && parentStyles && parentStyles.display !== 'inline'
```
kzar, line 125: I recommend against mixing ternary operator with logical and/or. Sometimes the results can be surprising. If you want to stick with this, I recommend at least adding parenthesis around the condition.
```suggestion
                    const parentInnerWidth = (parentElement && parentStyles && parentStyles.display !== 'inline')
```

kzar (approved):
LGTM

PR #546, opened 2023-05-31
Add new Messaging format to Click to Load

franfaccin:
Remove sendMessage calls from click-to-load.js and replace it with new messaging layer wrapper


src/features/click-to-load.js
```diff
@@ -48,9 +52,18 @@ const readyToDisplayPlaceholders = new Promise(resolve => {
 let afterPageLoadResolver
 const afterPageLoad = new Promise(resolve => { afterPageLoadResolver = resolve })
 
-// Used to choose between extension/desktop flow or mobile apps flow.
-// Updated on ClickToLoad.init()
-let isMobileApp
+// Messaging layer for Click to Load
+/** @type {import("@duckduckgo/messaging").Messaging} */
+let _messagingModuleScope
+const lazy = {
```
kzar, line 58: I'm not a huge fan of this, I figure it would be better to use the new messaging abstraction directly?

I've not looked at that, but I assume it would require making more things async, or at least thinking through more carefully about when the messaging system is ready for use. But I think that's better than pretending piercing the abstraction to pretend the messaging system is always ready.

src/features/click-to-load.js
```diff
@@ -842,9 +855,10 @@ async function replaceClickToLoadElements (targetElement) {
  * the page.
  * @param {unblockClickToLoadContentRequest} message
  * @see {@link ddg-ctp-unblockClickToLoadContent-complete} for the response handler.
+ * @return {Promise<any>}
```
kzar, line 858: Think it's `Promise<None>`?

src/features/click-to-load.js
```diff
@@ -1699,48 +1707,34 @@ function createYouTubePreview (originalElement, widget) {
...
-    getYouTubeVideoDetails (response) {
-        if (response?.status && typeof response.videoURL === 'string') {
-            originalWindowDispatchEvent(
-                createCustomEvent(
-                    'ddg-ctp-youTubeVideoDetails',
-                    { detail: response }
-                )
-            )
+export default class ClickToLoad extends ContentFeature {
+    // Messaging layer between Click to Load and the Platform
```
kzar, line 1711: Does the new messaging abstraction not take of this stuff for us? Seems bad that every feature needs to implement this, checking for platform support of messaging etc?

src/features/click-to-load.js
```diff
@@ -1806,7 +1800,20 @@ export default class ClickToLoad extends ContentFeature {
         // Request the current state of Click to Load from the platform.
         // Note: When the response is received, the response handler resolves
         //       the readyToDisplayPlaceholders Promise.
-        sendMessage('getClickToLoadState')
+        const clickToLoadState = await this.messaging.request('getClickToLoadState')
+        this.onClickToLoadState(clickToLoadState)
```
kzar, line 1804: Could we use `.then()` with an arrow function here instead? I figure we can inline the `onClickToLoadState` function here as an arrow function.

kzar (commented):
I took a pass of the click-to-load.js changes @franfaccin. Some suggestions but mostly they look pretty good :+1: . I will leave the sendMessageWrapper.js to Shane.

src/features/click-to-load.js
jonathanKingston: We could move this method to `ContentFeature`, it has access to the featureName as `this.name`

src/features/click-to-load.js
jonathanKingston: I agree but instead of making this code async we should just make is sync but only ever called after messaging is possible.

If that requires too much rewrite this is still acceptable to me so long as we never can get to the thrown error.

src/features/click-to-load.js
jonathanKingston: Does TS understand the type return? We could omit entirely if it does.

src/features/click-to-load.js
jonathanKingston: None of the features init calls block on each other if that's the concern?
Otherwise the current code is more readable imo.

src/features/click-to-load.js
```diff
@@ -1699,48 +1707,34 @@ function createYouTubePreview (originalElement, widget) {
...
-            originalWindowDispatchEvent(
-                createCustomEvent(
-                    'ddg-ctp-youTubeVideoDetails',
-                    { detail: response }
-                )
-            )
+export default class ClickToLoad extends ContentFeature {
+    // Messaging layer between Click to Load and the Platform
+    get messaging () {
+        if (this._messaging) return this._messaging
```
jonathanKingston, line 1713:
```suggestion
        if (this.#messaging) return this.#messaging
```

jonathanKingston: If we move this to `ContentFeature` we can change this to a private property like above. (Not really a requirement at all :) )

packages/messaging/lib/sendMessageWrapper.js
```diff
@@ -0,0 +1,144 @@
+import { sendMessage } from '../../../src/utils'
+
+export class SendMessageMessagingConfig {}
+
+/**
+ * An temporary implementation of {@link MessagingTransport} to communicate with Android and Extension.
```
jonathanKingston, line 6: If this is temporary, I suggest marking it as such by class name.

I think you can mark it as `@deprecated` immediately too.

src/features/click-to-load.js
jonathanKingston: I guess because of the `SendMessageMessagingConfig` being temporary we might not want to yet? Happy either way, `ContentFeature` is where it should live eventually for sure.

src/features/click-to-load.js
```diff
@@ -1699,48 +1707,34 @@ function createYouTubePreview (originalElement, widget) {
...
+
+        if (this.platform.name === 'android' || this.platform.name === 'extension') {
+            const messagingContext = new MessagingContext({
+                context: 'contentScopeScripts',
+                featureName: 'click-to-load',
+                env: this.isDebug ? 'development' : 'production'
+            })
+            const messagingConfig = new SendMessageMessagingConfig()
+            this._messaging = new Messaging(messagingContext, messagingConfig)
+            _messagingModuleScope = this._messaging
```
franfaccin, line 1723: I should actually move this to inside `init()`, right after we check for`this.messaging`, to ensure it is initialised right away.

jonathanKingston: I think this should be sync so shouldn't need to be triggered. I thought the getter was smart for this tbh.

src/features/click-to-load.js
kzar: I've not looked at ContentFeature, but if you think that's a better home for the logic then it sounds good to me. It seems kind of hacky having it in the Click to Load code.

src/features/click-to-load.js
kzar: Fair enough, also I just remembered this message will sometimes be sent unilaterally in the future so Fran's approach here is better too.

src/features/click-to-load.js
shakyShane: The issue here is all the module-level state. Something to be avoided where possible going forward.

These problems arise when you have free-standing functions that actually need access to state (_state_ here meaning the instance of Messaging)

Not for this project, but we should eventually resolve with dependency injection, either:

a) wrapping all the stateful things in more classes, converting to methods instead of free-standing functions
b) or, just wrap everything in a big function and use the closure to access 'stateful' things that we make available via the functions params

When does either, we should also split this file to separate the framework initialisation (boundary) with the actual CTL functionality. That will help to clearly define what is stateful, and what isn't :)

src/features/click-to-load.js
shakyShane: @kzar yep, it's a work around for now since not all platforms are implemented (yet!) - once we've worked around the patterns for CTL extension/android we can back-port to the messaging lib and make this more seamless.

On a call with @franfaccin today we moved the most boiler-plately parts into ContentFeature already :)

https://github.com/duckduckgo/content-scope-scripts/compare/shane/pr-review?expand=1#diff-cb668ba12d63994b1012861678e5aefba65ddcbe151a3e38a67962d4ab1f7861R46-R55

packages/messaging/lib/sendMessageWrapper.js
shakyShane: On a call with @franfaccin - I showed how to achieve this without modifying the underlying Messaging framework 💪🏻 

see: https://github.com/duckduckgo/content-scope-scripts/compare/shane/pr-review?expand=1#diff-bad7f5b7c7b37884f33af05ead612b430d72337fa089568d10e0cb7127e84516

The short version is that we now make it clear this is a CTL extension to messaging, and we handle stateful things with composition instead of ad-hoc changes to the messaging library :)

packages/messaging/index.js
```diff
@@ -212,6 +213,9 @@ function getTransport (config, messagingContext) {
     if (config instanceof TestTransportConfig) {
         return new TestTransport(config, messagingContext)
     }
+    if (config instanceof SendMessageMessagingConfig) {
+        return new SendMessageMessagingTransport(config, messagingContext)
+    }
```
shakyShane, line 218: As mentioned in our call, we can achieve this without touching the messaging framework :)

packages/messaging/lib/sendMessageWrapper.js
```diff
@@ -0,0 +1,144 @@
...
+ * It wraps the current messaging system that calls `sendMessage`
+ *
+ * @implements {MessagingTransport}
+ */
+export class SendMessageMessagingTransport {
```
shakyShane, line 11: This entire construct can be moved into the CTL folder and we keep it very specific to the migration period of extensions/android.

src/features/click-to-load.js
```diff
@@ -615,14 +628,14 @@ function createPlaceholderElementAndReplace (widget, trackingElement) {
 
     // YouTube
     if (widget.replaceSettings.type === 'youtube-video') {
-        sendMessage('updateYouTubeCTLAddedFlag', true)
+        lazy.messaging.notify('updateYouTubeCTLAddedFlag', true)
```
shakyShane, line 631: You cannot send primitive values like bools/string etc in the new system. It must be a `key: value` pair. This restriction helps with documentation, but more importantly it gives the value a name/label that native sides can use to refer to it.

With all _new_ features, you'd describe the message payloads in a way that can contribute to the auto-generated documentation, so it's a good habit to get into. To fix for your PR, all calls to `notify` or `request` would change to look like

```js
lazy.messaging.notify('updateYouTubeCTLAddedFlag', { addedFlag: true })

// or a generic 'payload'

lazy.messaging.notify('updateYouTubeCTLAddedFlag', { payload: true })
```

You can handle this in the custom transport if you don't want the Extension or Android to incur the cost of changing it - we don't need native changes. It's just the API restrictions are there to enforce certain practices :)

src/features/click-to-load.js
```diff
@@ -1526,7 +1537,7 @@ function createYouTubeBlockingDialog (trackingElement, widget) {
     )
     previewToggle.addEventListener(
         'click',
-        () => makeModal(widget.entity, () => sendMessage('setYoutubePreviewsEnabled', true), widget.entity)
+        () => makeModal(widget.entity, () => lazy.messaging.notify('setYoutubePreviewsEnabled', true), widget.entity)
```
shakyShane, line 1540: Same point as above, needs converting to some kind of key:value pair

src/features/click-to-load.js
```diff
@@ -856,6 +870,7 @@ function unblockClickToLoadContent (message) {
 function runLogin (entity) {
     const action = entity === 'Youtube' ? 'block-ctl-yt' : 'block-ctl-fb'
     unblockClickToLoadContent({ entity, action, isLogin: true })
```
shakyShane, line 872: Here, it's using `messaging.request('unblockClickToLoadContent', message)`, but you are not actually using the return value.

This can indicate an area to improve the design. For example, if it's truly a fire-and-forget message that you don't care about waiting for, then create a new message to serve that purpose (and use `.notify()` instead)

Or, it might be the case that you want to wait for the response, but don't need the value - in that case just use `await` on this and it might solve other synchronisation bugs too.

A bigger point to make here is that the distinction between `notify` and `request` is very deliberate - it's there to get you thinking deeply about the program flow whether or not things can/should be concurrent, serial or anything else.

shakyShane (changes requested):
@franfaccin this is great! I've tried to reflect what we spoke about in our call - let me know if you think I missed anything.

Here's the branch I made in our call for reference: https://github.com/duckduckgo/content-scope-scripts/compare/shane/pr-review

src/features/click-to-load.js
franfaccin: Thank you for the feedback.
I think that in this case, we can await the Platform response before firing the event to the surrogate. In my tests, it seems it was still working without calling this in sync, but I agree it might be a good practice to wait the rule to be unblocked before notifying the surrogate we're ready to login.
What do you think, @kzar?

src/features/click-to-load.js
```diff
@@ -1806,7 +1800,20 @@ export default class ClickToLoad extends ContentFeature {
...
+        // eslint-disable-next-line promise/prefer-await-to-then
+        this.messaging.subscribe(
+            'displayClickToLoadPlaceholders',
+            // TODO: Pass `message.options.ruleAction` through, that way only
+            //       content corresponding to the entity for that ruleAction need to
+            //       be replaced with a placeholder.
+            (_) => replaceClickToLoadElements()
+        )
+
         await readyToDisplayPlaceholders
```
franfaccin, line 1817: @kzar Considering that `await this.messaging.request('getClickToLoadState')` is now pausing this function, and that the `clickToLoadState()` that is called right after is what resolves `readyToDisplayPlaceholders`, I can remove this line, right? 
I noticed we still need this for `replaceClickToLoadElements()`, so I moved its subscribe up so its ready before we pause execution here, in case we have a race condition and get messages from the platform before we got the CTL state response.

franfaccin (commented):
I've replied to the comments and updated the PR accordingly. It's ready for another review :)

shakyShane (dismissed):

shakyShane (approved):

shakyShane:
@franfaccin - I've approved this now - thanks for working through everyone's feedback ❤️!

If you consider this complete, please merge when you are ready 💪🏻 

Just a reminder too that we're attempting to keep `main` in an 'always ready to release' state, so if there was any last-minute testing, or small changes you wanted to make, please do them here before merging :)

PR #547, opened 2023-05-31
Change to throw away iframe for capturing globals

jonathanKingston:


src/captured-globals.js
```diff
@@ -1,16 +1,23 @@
...
+        dummyWindow = document.createElement('iframe')
+        injectionElement.appendChild(dummyWindow)
     }
-    return newGlobal
 }
+// Capture prototype to prevent overloading
+function captureGlobal (globalName) {
+    // if we were unable to create a dummy window, return the global
+    // this still has the advantage of preventing aliasing of the global through shawdowing
+    return dummyWindow?.contentWindow?.[globalName] || globalThis[globalName]
```
muodov, line 16: If it really works, it's really neat!
- will the returned values be unaffected by prototype modifications in the current context?
- are there any globals that cannot be copied this way? (i'm thinking `Document`, `Window`, etc)
- can CSP be a problem?
- will removing of the iframe not invalidate the returned objects?

jonathanKingston: 1. Yeah it's completely isolated
2. Possibly, I've not tried them all. The globals won't match if that's your concern.
3. Shouldn't be, the URL is about:blank and will be required for basic iframe use. Worst case scenario it's prevented and we fall back to the globalThis.
4. Didn't in my testing, the reference keeps it from being gc'd even once the document goes.

shakyShane (dismissed):

shakyShane:
@jonathanKingston  what about testing steps - is it just extension + C-S-S, do I need any config changes?

shakyShane:
@jonathanKingston - agreed, good approach :)

jonathanKingston:
I actually found the source of the crazy breakage (remember we've broke this before anyway so clearly we need a test... somehow). I re-opened here: https://github.com/duckduckgo/content-scope-scripts/pull/554

PR #548, opened 2023-05-31
Fix duplicating customElement definition

franfaccin:
Fix duplicating customElement definition by first checking if it was previously defined.

### DIFF
--- a/src/features/click-to-load/components/ctl-placeholder-blocked.js
+++ b/src/features/click-to-load/components/ctl-placeholder-blocked.js
@@ -299,4 +299,6 @@ export class DDGCtlPlaceholderBlockedElement extends HTMLElement {
     }
 }
 
-customElements.define(DDGCtlPlaceholderBlockedElement.CUSTOM_TAG_NAME, DDGCtlPlaceholderBlockedElement)
+if (!customElements.get(DDGCtlPlaceholderBlockedElement.CUSTOM_TAG_NAME)) {
+    customElements.define(DDGCtlPlaceholderBlockedElement.CUSTOM_TAG_NAME, DDGCtlPlaceholderBlockedElement)
+}


src/features/click-to-load/components/ctl-placeholder-blocked.js
```diff
@@ -299,4 +299,6 @@ export class DDGCtlPlaceholderBlockedElement extends HTMLElement {
     }
 }
 
-customElements.define(DDGCtlPlaceholderBlockedElement.CUSTOM_TAG_NAME, DDGCtlPlaceholderBlockedElement)
```
jonathanKingston, line 302: Sorry just seen this PR.

Can we move the call site into `init()` that way it can be disabled with config.

shakyShane: taking @jonathanKingston 's advise, we'll stick this in a separate file within a function. Then if we need remote config later we can easily use it, something like 

```js
import { DDGCtlPlaceholderBlockedElement } from './ctl-placeholder-blocked.js'

/**
 * Wrapping in a function to allow remote-config later if needed
 */
export function registerCustomElements () {
    if (!customElements.get(DDGCtlPlaceholderBlockedElement.CUSTOM_TAG_NAME)) {
        customElements.define(DDGCtlPlaceholderBlockedElement.CUSTOM_TAG_NAME, DDGCtlPlaceholderBlockedElement)
    }
}
```

it can be called from `init()` as suggested, and config can be passed in later if needed :)

franfaccin: Thank you both for your suggestions! I agree with them and have updated the PR with it.

Let me know if there are any follow-up comments.

shakyShane (approved):

PR #549, opened 2023-06-01
Overriden taint methods

jonathanKingston:
https://app.asana.com/0/0/1203922126796182/f


src/features/runtime-checks.js
```diff
@@ -573,4 +576,211 @@ export default class RuntimeChecks extends ContentFeature {
...
+        const storage = new MemoryStorage()
+        this.overrideStorage(config, key, storage)
+    }
+
+    overloadStorageWithSession (key, config) {
+        const storage = globalThis.sessionStorage
+        this.overrideStorage(config, key, storage)
+    }
+
+    overrideStorage (config, key, storage) {
```
shakyShane, line 702: Does this handle the following situations? 

```js
localStorage.foo = "bar"

console.log(localStorage.getItem('foo'))
// 'bar'
```

src/features/runtime-checks.js
```diff
@@ -573,4 +576,211 @@ export default class RuntimeChecks extends ContentFeature {
...
+                if (getTaintFromScope(this, arguments, config.stackCheck)) {
+                    return returnVal
+                }
+                return defaultVal
+            }
+        })
+    }
+}
+
+function findClosestBreakpoint (breakpoints, screenSize) {
```
shakyShane, line 771: is there a simple unit-test we can add to this? seems like low-hanging fruit and would help to explain the math :)

shakyShane, line 771: Can we add a return-type jsdocs here, that'll ensure we can't assign incorrectly in the loop

src/features/runtime-checks.js
```diff
@@ -66,12 +66,20 @@ const jsMimeTypes = [
...
 ]
 
+function getTaintFromScope (scope, args, shouldStackCheck = false) {
+    try {
+        scope = args.callee.caller
+    } catch {}
+    return hasTaintedMethod(scope, shouldStackCheck)
+}
+
 class DDGRuntimeChecks extends HTMLElement {
```
shakyShane, line 76: not directly related to this PR, but with more functionality being added we should be discussing when it makes sense to split this file up.

shakyShane (commented):
Added a couple of comments and 1 question around localStorage/sessionStorage

src/features/runtime-checks.js
jonathanKingston: Nope but it does now with the power of proxies and a test. Thanks for the catch here!

src/features/runtime-checks.js
```diff
@@ -573,4 +577,204 @@ export default class RuntimeChecks extends ContentFeature {
...
+    overloadStorageWithMemory (config, key) {
+        class MemoryStorage {
+            #data = {}
+
+            setItem (id, val) {
+                this.#data[id] = String(val)
+            }
+
+            getItem (id) {
+                return Object.prototype.hasOwnProperty.call(this.#data, id) ? this.#data[id] : undefined
```
shakyShane, line 673: I think this should be `null` and not undefined - to ensure `===` checks by the caller work as expected

src/features/runtime-checks.js
```diff
@@ -573,4 +577,204 @@ export default class RuntimeChecks extends ContentFeature {
...
+            }
+        })
+    }
+
+    overloadStorageWithMemory (config, key) {
+        class MemoryStorage {
+            #data = {}
+
+            setItem (id, val) {
+                this.#data[id] = String(val)
```
shakyShane, line 669: how closely do we need to match APIs here? because `localStorage.setItem('foo')`  (missing the value) would throw in the browser, but this implementation will set the key with `undefined` as the value

src/features/runtime-checks.js
```diff
@@ -573,4 +577,204 @@ export default class RuntimeChecks extends ContentFeature {
...
+            }
+
+            getItem (id) {
+                return Object.prototype.hasOwnProperty.call(this.#data, id) ? this.#data[id] : undefined
+            }
+
+            removeItem (id) {
+                return delete this.#data[id]
```
shakyShane, line 677: will this return `true` to the caller? if so, it doesn't match the browser API (should be undefined/void)

shakyShane (commented):
Couple more about storage :)

shakyShane:
Approving on the basis that we can disable easily - it's something that needs verifying in the wild now :)

shakyShane (approved):

PR #556, opened 2023-06-02
Add search integration test to verify results

jonathanKingston:


shakyShane:
There were some console errors in the integration test output - I don't think we should load the platform-specific features into 'integration' -> testing here https://github.com/duckduckgo/content-scope-scripts/pull/557

shakyShane (approved):

PR #571, opened 2023-06-13
Make duck player page more resilient to race conditions on macOS

shakyShane:
When implementing Duck Player on macOS - I encountered a race-condition where our script in the page is calling for a request/response before the special page has been marked as 'safe to receive messages' by the native side.

I looked into fixing it on their end, but it's heavily related to the fact that user-scripts are global and cannot be directly tied to a particular hostname.

So, I'm fixing it here by attempting (10 times) to make an 'initial connection' - once that's established we continue as normal.

This is a nice way of ensuring we can proceed across all platforms. So whilst it's technically a macOS only fix right now, it really helps everywhere by not giving up on first error.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -69,9 +69,15 @@ jobs:
       if: always()
       with:
         name: playwright-report
-        path: playwright-report/
+        path: test-results
         retention-days: 5
     - run: npm run test-int-x
+    - uses: actions/upload-artifact@v2
+      if: always()
+      with:
+        name: playwright-report-pages
+        path: packages/special-pages/test-results
+        retention-days: 5
     - name: Build docs
       run: npm run docs
 
--- a/packages/special-pages/pages/duckplayer/src/js/index.js
+++ b/packages/special-pages/pages/duckplayer/src/js/index.js
@@ -31,7 +31,12 @@
  *
  * Please see {@link DuckPlayerPageMessages} for the up-to-date list
  */
-import { createDuckPlayerPageMessaging, DuckPlayerPageMessages, UserValues } from './messages'
+import {
+    callWithRetry,
+    createDuckPlayerPageMessaging,
+    DuckPlayerPageMessages,
+    UserValues
+} from './messages'
 import { html } from '../../../../../../src/dom-utils'
 import { initStorage } from './storage'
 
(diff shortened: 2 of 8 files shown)


shakyShane:
* main
  * **PR #571** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/571" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/571?utm_source=stack-comment).

packages/special-pages/pages/duckplayer/src/js/index.js
```diff
@@ -325,23 +330,31 @@ const Comms = {
...
-        Comms.messaging.getUserValues().then((value) => {
-            if ('enabled' in value.privatePlayerMode) {
-                Setting.setState(true)
-            } else {
-                Setting.setState(false)
-            }
+    init: async (opts) => {
+        const messaging = createDuckPlayerPageMessaging(opts)
+        // try to make communication with the native side.
+        const result = await callWithRetry(() => {
```
shakyShane, line 336: When implementing on macOS - I encountered a race-condition where our script in the page is calling for a request/response _before_ the special page has been marked as 'safe to receive messages' by the native side.

I looked into fixing it on their end, but it's heavily related to the fact that user-scripts are global and cannot be directly tied to a particular hostname.

So, I'm fixing it here by attempting (10 times) to make an 'initial connection' - once that's established we continue as normal.

This is a nice way of ensuring we can proceed across all platforms. So whilst it's technically a macOS only fix right now, it really helps everywhere by not giving up on first error.

.github/workflows/tests.yml
```diff
@@ -69,9 +69,15 @@ jobs:
...
         name: playwright-report
-        path: playwright-report/
+        path: test-results
         retention-days: 5
     - run: npm run test-int-x
+    - uses: actions/upload-artifact@v2
+      if: always()
+      with:
+        name: playwright-report-pages
+        path: packages/special-pages/test-results
```
shakyShane, line 79: This just ensures that upload playwright traces when things go wrong

packages/special-pages/tests/duckplayer.spec.js
```diff
@@ -98,7 +98,7 @@ test.describe('duckplayer settings', () => {
         await duckplayer.toggleAlwaysOpenSetting()
         await duckplayer.sentUpdatedSettings()
     })
-    test.skip('when a new value arrives via subscription', async ({ page }, workerInfo) => {
+    test('when a new value arrives via subscription', async ({ page }, workerInfo) => {
```
shakyShane, line 101: these were skipped, but now run 💪🏻

PR #573, opened 2023-06-14
Fix issue to get accurate iframe video src

franfaccin:
Sometimes, YT iframes won't have a set `src` attribute, and instead are using `data-src=*`. This caused breakage since we were relying on `src` to store the original video src when replacing it.

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -523,7 +523,7 @@ function replaceTrackingElement (widget, trackingElement, placeholderElement) {
     // from the DOM after they are collapsed. As a workaround, have the iframe
     // load a benign data URI, so that it's uncollapsed, before removing it from
     // the DOM. See https://crbug.com/1428971
-    const originalSrc = elementToReplace.src
+    const originalSrc = elementToReplace.src || elementToReplace.getAttribute('data-src')
     elementToReplace.src =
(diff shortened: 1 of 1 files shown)


src/features/click-to-load.js
```diff
@@ -523,7 +523,7 @@ function replaceTrackingElement (widget, trackingElement, placeholderElement) {
     // from the DOM after they are collapsed. As a workaround, have the iframe
     // load a benign data URI, so that it's uncollapsed, before removing it from
     // the DOM. See https://crbug.com/1428971
-    const originalSrc = elementToReplace.src
+    const originalSrc = elementToReplace.src || elementToReplace.getAttribute('data-src')
```
shakyShane, line 526: are there any situations where both could be absent? `x.getAttribute('...')` can return `null` you see.

franfaccin: Idk what could be a better fallback in case both are empty. I'm taking suggestions. Although, now that I went to check it, we're only selecting elements that have a video in either of these 2 fields (https://github.com/duckduckgo/content-scope-scripts/blob/main/src/features/click-to-load/ctl-config.js#L988), so I think it would be safe to assume that at least on of them will have the video url.

shakyShane: Sounds good!

shakyShane (approved):

kzar (approved):

PR #574, opened 2023-06-14
Prevent debugging serialization to cause page to break

jonathanKingston:
Fixes page load issues on Amazon, https://www.citi.com/ and many others where we're serializing symbols that aren't serializable.

This appears to be any page with react so makes debugging very annoying.


dharb (dismissed):
Thanks! 👍

PR #586, opened 2023-06-20
Click to Load - Facebook login - mobile

franfaccin:


src/features/click-to-load.js
```diff
@@ -801,43 +858,6 @@ function hideInfoTextIfNarrowPlaceholder (shadowRoot, placeholder, narrowWidth)
...
 }
 
-/**
- * Replace the blocked CTL elements on the page with placeholders.
- * @param {HTMLElement} [targetElement]
- *   If specified, only this element will be replaced (assuming it matches
- *   one of the expected CSS selectors). If omitted, all matching elements
- *   in the document will be replaced instead.
- */
-async function replaceClickToLoadElements (targetElement) {
```
franfaccin, line 811: I moved this function inside `ClickToLoad` class to have access to the its context and `this.platform` when creating a new  `DuckWidget`

src/features/click-to-load/assets/variables.css
franfaccin, line 1: This was renamed to `shared.css` to also include shared CSS styles, like Buttons and Links.

src/features/click-to-load/components/ctl-login-button.js
```diff
@@ -0,0 +1,169 @@
...
+import css from '../assets/ctl-login-button.css'
+import { logoImg } from '../ctl-assets'
+
+/**
+ * @typedef LearnMoreParams - "Learn More" link params
+ * @property {string} readAbout - "Learn More" aria-label text
+ * @property {string} learnMore - "Learn More" link text
+ */
+
+export class DDGCtlLoginButton extends HTMLElement {
```
franfaccin, line 12: Since I had this practically done as part of my [POC](https://github.com/duckduckgo/content-scope-scripts/compare/main...franfaccin:content-scope-scripts:ffaccin/ctl-fe-refactor), I cleaned it up and reused what was the same from the previous designs and only updated the button styles to match what we need for mobile.

src/features/click-to-load.js
```diff
@@ -80,8 +84,10 @@ class DuckWidget {
      *   The original tracking element to replace with a placeholder.
      * @param {string} entity
      *   The entity behind the tracking element (e.g. "Facebook, Inc.").
+     * @param {import('../utils').Platform} platform
+     *   The platform where Click to Load and the Duck Widget is running on.
```
kzar, line 88: Nit: Maybe explain what a platform is in this context? E.g. "The browser platform, e.g. "android"."

src/features/click-to-load.js
```diff
@@ -27,9 +28,12 @@ const titleID = 'DuckDuckGoPrivacyEssentialsCTLElementTitle'
...
 let styles = null
-// Used to choose between extension/desktop flow or mobile apps flow.
-// Updated on ClickToLoad.init()
-let isMobileApp
+
+/**
+ * List of platforms where we can skip showing a Web Modal from C-S-S.
+ * It is generally expected that the platform will show a native modal instead.
+ * @type {import('../utils').Platform["name"][]} */
+const platformsToSkipCSSModal = ['android', 'ios']
```
kzar, line 36: Nits:

- Maybe `platformsWithNavitveModalSupport`?
- It would be nice to have the similar arrays (or ideally sets) for mobile platforms and custom element support here too.
- Maybe add the checking method for this one to DDGWidget too, or maybe remove the checking methods for the other ones? I don't mind either way, it just seems kind of inconsistent at the moment.

src/features/click-to-load.js
```diff
@@ -473,23 +482,53 @@ class DuckWidget {
                     if (onError) {
                         fbElement.addEventListener('error', onError, { once: true })
                     }
+                // eslint-disable-next-line promise/prefer-await-to-then
+                }).catch((err) => {
+                    if (err === 'ddg-ctp-user-cancel') {
```
kzar, line 487: I wonder if we should re-throw the exception if it's not the user cancelling one?

Also, where is this being thrown? I can't find `ddg-ctp-user-cancel` elsewhere. I was looking for it because I wondered if the exception is _really_ a string, I figured it might be something like `new Error("ddg-ctp-user-cancel")` instead.

src/features/click-to-load/components/ctl-login-button.js
```diff
@@ -0,0 +1,169 @@
...
+    createLoginButton = () => {
+        const { label, hoverText, logoIcon, learnMore } = this.params
+
+        const { popoverStyle, arrowStyle } = this.calculatePopoverPosition()
+
+        const container = document.createElement('div')
+        // Add our own styles and inherit any local class styles on the button
+        container.classList.add('ddg-fb-login-container')
+
+        container.innerHTML = html`
```
kzar, line 78: Assigning to innerHTML like this makes me a little nervous, especially since we're putting translated strings in there. Has potential to be a security vulnerability if a translation sneaks in with some malicious HTML for example. This might be fine and there might not be a better way, just thought I should raise it.

kzar (commented):
Looks good Fran. Just a few nits/small comments.

src/features/click-to-load/components/ctl-login-button.js
shakyShane: Actually the `html` tagged template function used here will handle the escaping of any content injected via `${...}`

shakyShane (changes requested):

src/features/click-to-load/components/ctl-login-button.js
kzar: Cool OK, so for this one we don't need to use any localised strings that are supposed to contain HTML?

franfaccin: That's a good point. If the localization has HTML, I guess it wouldn't work. So far, the translations for the elements in these web components didn't have an HTML embedded in them. I'm aware that the YouTube Preview has one, although the mobile layouts and copy planned for it I believe don't have it.

Should we make an effort to avoid HTML in the translations? Perhaps either break down the string when possible or give preference to a placeholder to extract what needs to be wrapped in a HTML tag.

src/features/click-to-load/ctl-messaging-transport.js
```diff
@@ -77,20 +77,25 @@ export class ClickToLoadMessagingTransport {
...
 
         sendMessage(req.method, params)
 
-        return new this.globals.Promise((resolve) => {
+        return new this.globals.Promise((resolve, reject) => {
             this._subscribe(comparator, (msgRes, unsubscribe) => {
                 unsubscribe()
 
-                resolve(msgRes.response)
+                if ('error' in msgRes) {
```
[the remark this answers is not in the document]
franfaccin: I've replied to this better in the thread above https://github.com/duckduckgo/content-scope-scripts/pull/586/files#r1238471284 .  But in summary, you're right this not an exception, but a different type of response, and I'm treating it as such now.

src/features/click-to-load/components/ctl-login-button.js
shakyShane: @franfaccin sounds likes a good approach for now. If it becomes unavoidable then we can make the `html` tag work - but no need right now if we can avoid :)

shakyShane (approved):

src/features/click-to-load/components/ctl-login-button.js
kzar: > Should we make an effort to avoid HTML in the translations?

Well, in the past I've seen translation systems that allow translated strings to contain a subset of HTML tags (e.g. <b>, <i>, <em>, <a>). All attributes were stripped from those tags in the translation and the link targets etc were populated by the system at runtime. I think sometimes you do need HTML in translations, to handle the "Learn more" case properly.

I guess this is off-topic for this review though, so feel free to resolve. I guess we can just punt this, but it would be good if our translation / HTML templating systems could handle it in the future.

src/features/click-to-load.js
kzar: > I can't add this check to a method in the DDGWidget too because there isn't always a widget...

Ah right, got it.

>  I'd like to start avoiding having a lot of config floating in the module scope.

I don't feel too strongly. To me, these are constants that are not really a configuration, but a enumeration of the code's capabilities. In the future if more platforms were added to support native dialogs, the constant could be updated with that change. That said, I don't feel strongly and I'd rather not  bike shed it. So feel free to put those constants where you think makes most sense!

kzar (approved):
LGTM :1st_place_medal:

src/features/click-to-load/components/ctl-login-button.js
franfaccin: I like the idea of a subset of tags without any properties. One of the issues of having HTML in the translated copies, besides possible injections, is maintaining the HTML. We might need to change the CSS, add/remove attributes, change href, etc, and that becomes hard to do when they're all embedded in the translated string. So having a placeholder in the string to extract what needs to wrapped in HTML or some subset of simple tags that can be repopulated with the necessary attributes sound like good ideas to consider.

But like you said, this is not our case in the project yet 😅 But I agree we will need to figure out how to make it work best once that starts happening.

PR #610, opened 2023-07-06
Set Facebook Click to Load breakage flags

kzar:
Take care to set the new Facebook Click to Load breakage flags, to
give us a better idea how often the feature causes website breakage
for our users.

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -503,6 +503,12 @@ class DuckWidget {
         // If this is a login button, show modal if needed
         if (this.replaceSettings.type === 'loginButton' && entityData[this.entity].shouldShowLoginModal) {
             return e => {
+                // Even if the user cancels the login attempt, consider Facebook Click to
+                // Load to have been active on the page if the user reports the page as broken.
+                if (this.entity === 'Facebook, Inc.') {
+                    notifyFacebookLogin()
+                }
+
                 handleUnblockConfirmation(
                     this.platform.name, this.entity, handleClick, e
                 )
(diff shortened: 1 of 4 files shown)


src/features/click-to-load.js
```diff
@@ -910,6 +911,12 @@ function unblockClickToLoadContent (message) {
  *   The parameters passed to acceptFunction when it is called.
  */
 function handleUnblockConfirmation (platformName, entity, acceptFunction, ...acceptFunctionParams) {
+    // Even if the user cancels the login attempt, consider Facebook Click to
+    // Load to have been active on the page if the user reports the page as broken.
+    if (entity === 'Facebook, Inc.') {
+        ctl.messaging.notify('updateFacebookCTLBreakageFlags', { ctlFacebookLogin: true })
+    }
```
franfaccin, line 918: `handleUnblockConfirmation` might not only be used for login in the future. Might be best to set the flag either only in the `runLogin` or prior this message being called.

kzar: Ah right, that's a good point. Thing is I still want to set the flag if the user clicks a login button, but then cancels the modal. I don't think `runLogin` is called until the user accepts the modal. Is there a better place to put this? I wasn't sure where was best to be honest.

franfaccin: Yes, you're right, `runLogin` might be too late then.
I'd suggest adding it right before this function is called then. It would be [here](https://github.com/duckduckgo/content-scope-scripts/blob/main/src/features/click-to-load.js#L506) and [here](https://github.com/duckduckgo/content-scope-scripts/blob/main/src/features/click-to-load.js#L1848).

kzar: Got it, thanks. Done.

franfaccin (approved):
great work! ✅

shakyShane (approved):

PR #617, opened 2023-07-10
Replace private method support with _

joshliebe:
Asana task: https://app.asana.com/0/488551667048375/1204986151711455/f

Fixes an issue with private method support on older Android `WebView` versions.


src/features/click-to-load/components/ctl-login-button.js
```diff
@@ -20,7 +20,7 @@ export class DDGCtlLoginButton {
      * Placeholder container element for blocked login button
      * @type {HTMLDivElement}
      */
-    #element
```
jonathanKingston, line 23: I don't think we need this at least, properties landed much before methods.

jonathanKingston (approved):

PR #619, opened 2023-07-10
duckplayer fixes

shakyShane:
https://app.asana.com/0/0/1205002113935587/f

This fixes 2 things:

- [x] 1) A previous fix introduced a regression where we always called `e.preventDefault()` when we shouldn't
- [x] 2) YT upgrades the markup on the elements for shorts, so now we double-check on click whether or not it's a valid duck player url before we prevent default and try to navigate.

### DIFF
--- a/integration-test/playwright/duckplayer.e2e.spec.js
+++ b/integration-test/playwright/duckplayer.e2e.spec.js
@@ -19,4 +19,24 @@ test.describe('e2e: Duck Player Thumbnail Overlays on YouTube.com', () => {
         await overlays.hoverShort()
         await overlays.overlaysDontShow()
     })
+    test('control (without our script): clicking on a short loads correctly', async ({ page }, workerInfo) => {
+        // @ts-expect-error - TS doesn't know about the "use.e2e" property
+        workerInfo.skip(!workerInfo.project.use?.e2e)
+        const overlays = DuckplayerOverlays.create(page, workerInfo)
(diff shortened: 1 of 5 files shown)


shakyShane:
* main
  * **PR #619** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/619" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/619?utm_source=stack-comment).

integration-test/playwright/windows-permissions.spec.js
```diff
@@ -31,11 +31,42 @@ export class WindowsPermissionsSpec {
...
     }
 
+    /**
+     * In CI, the global objects such as USB might not be installed on the
+     * version of chromium running there.
+     */
+    async installPolyfills () {
+        await this.page.addInitScript(() => {
+            // @ts-expect-error - testing
+            if (typeof Bluetooth === 'undefined') {
```
jonathanKingston, line 47: Why is this in the patch?

shakyShane: I think we need to revert the Harmful API merge - I'll open it as a separate PR after I've checked with @muodov

jonathanKingston (approved):
As discussed this is more robust than the existing code as we're not caching the href.

We're removing the hover code as this was a YT experiment.

PR #645, opened 2023-07-21
Remove unused logic from referrer trimming

sammacbeth:
https://app.asana.com/0/892838074342800/1205066264803671/f

We previously had a code path in the referrer trimming code to take a referrer value sent by the background script. Since we removed eTLD+1 truncation, this isn't needed anymore (this script has enough context to apply the trimming itself), so we can remove that code path.


jonathanKingston (approved):
Perfect thank you!

PR #648, opened 2023-07-24
Breakage debug flags

muodov:

This PR adds an MVP infrastructure for breakage debug flags. Currently this is only passed in web extensions (see https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/2138).
In this PR, flags are sent by `fingerprinting-battery` and `fingerprinting-hardware` features, more will come in the follow-up PR.

---
  - https://app.asana.com/0/0/1205073793816810

### DIFF
--- a/build-output.eslintrc
+++ b/build-output.eslintrc
@@ -1,6 +1,6 @@
 {
     "root": true,
-    "plugins": ["promise", "@typescript-eslint"],
+    "plugins": ["promise", "@typescript-eslint", "eslint-plugin-n"],
     "parserOptions": {
       "ecmaVersion": "latest",
     },
--- a/inject/chrome.js
+++ b/inject/chrome.js
@@ -13,6 +13,7 @@ const allowedMessages = [
     'getClickToLoadState',
     'getYouTubeVideoDetails',
     'openShareFeedbackPage',
+    'addDebugFlag',
     'setYoutubePreviewsEnabled',
     'unblockClickToLoadContent',
     'updateYouTubeCTLAddedFlag',
--- a/src/content-feature.js
+++ b/src/content-feature.js
@@ -31,8 +31,8 @@ export default class ContentFeature {
     #messagingContext
     /** @type {import('../packages/messaging').Messaging} */
     #debugMessaging
-    /** @type {Set<string>} */
-    #debugFlags
+    /** @type {boolean} */
+    #isDebugFlagSet = false
 
     /** @type {{ debug?: boolean, featureSettings?: Record<string, unknown>, assets?: AssetConfig | undefined, site: Site  } | null} */
     #args
@@ -41,7 +41,6 @@ export default class ContentFeature {
         this.name = featureName
         this.#args = null
         this.monitor = new PerformanceMonitor()
-        this.#debugFlags = new Set()
     }
 
     get isDebug () {
@@ -248,15 +247,13 @@ export default class ContentFeature {
     }
 
     /**
-     * Register a string flag that will be added to page breakage reports
-     * @param {string} flag
+     * Register a flag that will be added to page breakage reports
      */
-    addDebugFlag (flag = '') {
-        if (this.#debugFlags.has(flag)) return
-        this.#debugFlags.add(flag)
-        const suffix = flag ? `.${flag}` : ''
+    addDebugFlag () {
+        if (this.#isDebugFlagSet) return
+        this.#isDebugFlagSet = true
         this.debugMessaging?.notify('addDebugFlag', {
-            flag: `${this.name}${suffix}`
+            flag: this.name
         })
     }
 }
--- a/src/content-feature.js
+++ b/src/content-feature.js
@@ -110,7 +110,7 @@ export default class ContentFeature {
         if (this.#messaging) return this.#messaging
 
         // TODO: use createMessaging() from create-messaging.js when all platforms are supported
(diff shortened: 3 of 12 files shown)


build-output.eslintrc
```diff
@@ -1,6 +1,6 @@
 {
     "root": true,
-    "plugins": ["promise", "@typescript-eslint"],
+    "plugins": ["promise", "@typescript-eslint", "eslint-plugin-n"],
```
jonathanKingston, line 3: I still dunno why this was needed hah, has there been anything changed you can see that correlates?

jonathanKingston: > error  Definition for rule 'n/no-callback-literal' was not found  n/no-callback-literal

was the error from [this run](https://github.com/duckduckgo/content-scope-scripts/actions/runs/5646734760/job/15295293432?pr=648)

inject/chrome.js
```diff
@@ -13,6 +13,7 @@ const allowedMessages = [
     'getClickToLoadState',
     'getYouTubeVideoDetails',
     'openShareFeedbackPage',
+    'addDebugFlag',
```
jonathanKingston, line 16: Is this still needed, aren't you using the other transport?

src/content-feature.js
```diff
@@ -87,14 +90,32 @@ export default class ContentFeature {
      */
     get messagingContext () {
         if (this.#messagingContext) return this.#messagingContext
+
+        const contextName = import.meta.injectName === 'apple-isolated'
```
jonathanKingston, line 94: We could just land the messaging refactoring + extension support as an isolated PR and discuss the other parts after.

src/content-feature.js
```diff
@@ -222,4 +243,14 @@ export default class ContentFeature {
     // eslint-disable-next-line @typescript-eslint/no-empty-function
     update () {
     }
+
+    /**
+     * Register a string flag that will be added to page breakage reports
+     * @param {string} flag
+     */
+    addDebugFlag (flag = 'fired') {
+        this.debugMessaging?.notify('addDebugFlag', {
```
jonathanKingston, line 252: This method **needs** to memoise and fire only once, to prevent swamping the native messaging with debugging if the site does something like:

```
const pixels = ctx.getImageData(0, 0, canvas.width, canvas.height);
for (let i = 0; i<imageData.data.length; i++) {
  if (navigator.deviceMemory) {
  ...
  }
}
```

jonathanKingston (changes requested):

build-output.eslintrc
muodov: This test was failing since the first push, I'm not sure what's causing it. I can try to bisect

inject/chrome.js
muodov: This is needed for MV2. We haven't implemented a proper messaging transport for the extension.

jonathanKingston: Ok thanks for the heads up :)

build-output.eslintrc
jonathanKingston: And not locally right? So weird. If you get chance it would be worth a bash.

unit-test/content-feature.js
```diff
@@ -49,4 +49,50 @@ describe('Helpers checks', () => {
...
+            feature.addDebugFlag('someflag')
+            expect(spyNotify).toHaveBeenCalledWith(
+                'addDebugFlag',
+                {
+                    flag: 'someFeatureName.someflag'
+                }
+            )
+        })
+
+        it('should not send duplicate flags', () => {
```
jonathanKingston, line 78: Nice thanks for this.

src/content-feature.js
```diff
@@ -222,4 +246,17 @@ export default class ContentFeature {
     // eslint-disable-next-line @typescript-eslint/no-empty-function
     update () {
     }
+
+    /**
+     * Register a string flag that will be added to page breakage reports
+     * @param {string} flag
+     */
+    addDebugFlag (flag = '') {
```
jonathanKingston, line 254: If we're not intending to use this flag code path immediately I would prefer we remove it.

The code instead could set itself to be a noop even (not sure if that's a weird convention and hard to follow) but instead it could check a boolean state instead of a set.

src/features/fingerprinting-hardware.js
```diff
@@ -4,16 +4,25 @@ import { wrapProperty } from '../wrapper-utils'
 export default class FingerprintingHardware extends ContentFeature {
     init () {
         wrapProperty(globalThis.Navigator.prototype, 'keyboard', {
-            // @ts-expect-error - error TS2554: Expected 2 arguments, but got 1.
-            get: () => this.getFeatureAttr('keyboard')
+            get: () => {
+                this.addDebugFlag()
```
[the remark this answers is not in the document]
muodov: I am going to follow up with this change in another PR: https://github.com/duckduckgo/content-scope-scripts/pull/652

build-output.eslintrc
muodov: I'm getting this error locally even on the current `main` branch. Maybe some of the external packages changed?

jonathanKingston (approved):

build-output.eslintrc
jonathanKingston: Not sure (shouldn't do as the packages should be pinned), or the tests were skipped recently. Let's add for now. Thanks for checking.

PR #652, opened 2023-07-26
Wrapper utils as methods

muodov:

This PR is a follow-up on https://github.com/duckduckgo/content-scope-scripts/pull/648. It does the following:
- moves standalone descriptor wrapper functions inside the `ContentFeature` class
- refactors features to use them (by converting some functions into methods)
- adds debug logging in those methods so that features automatically report "triggering"

This PR **does not**:
- add debug logging to all features
- instrument `DDGProxy` with debug logging
These will come in a separate PR

I recommend reviewing commit-by-commit, I tried to make them logical and clean


src/content-feature.js
```diff
@@ -259,4 +265,126 @@ export default class ContentFeature {
...
+     * @param {PropertyDescriptor} descriptor
+     */
+    defineProperty (object, propertyName, descriptor) {
+        // make sure to send a debug flag when the property is used
+        // NOTE: properties passing data in `value` would not be caught by this
+        ['value', 'get', 'set'].forEach((k) => {
+            const descriptorProp = descriptor[k]
+            if (typeof descriptorProp === 'function') {
+                // eslint-disable-next-line @typescript-eslint/no-this-alias
+                const self = this
```
jonathanKingston, line 282: `self` is shadowing `window.self` kinda shocked the eslint rule isn't telling you off for that too. Could we choose a different name here?

Alternatively you could alias the debug method instead perhaps?

```suggestion
                const addDebugFlag = this.addDebugFlag
```

muodov: Right, thanks for catching!

src/features/runtime-checks.js
```diff
@@ -729,7 +729,7 @@ export default class RuntimeChecks extends ContentFeature {
 
     overloadDateGetTimezoneOffset (config) {
         const offset = (new Date()).getTimezoneOffset()
-        defineProperty(globalThis.Date.prototype, 'getTimezoneOffset', {
+        this.defineProperty(globalThis.Date.prototype, 'getTimezoneOffset', {
```
muodov, line 732: @jonathanKingston not sure if Messaging will actually work in injected scripts. The tests pass, but how do I test it manually to verify?

jonathanKingston: I doubt it will work, I don't think we wired it in.
We have a test page that you can run these locally: https://github.com/duckduckgo/content-scope-scripts/blob/main/integration-test/test-pages/runtime-checks/pages/basic-run.html with the config: https://github.com/duckduckgo/content-scope-scripts/blob/main/integration-test/test-pages/runtime-checks/config/basic-run.json

muodov: Ok, I just confirmed that the messages are silently ignored, so we can keep using the wrapper methods, but debug flags won't be delivered from runtime checks at the moment. Which is fine for now.

src/wrapper-utils.js
```diff
@@ -2,10 +2,11 @@
...
 export const hasMozProxies = typeof mozProxies !== 'undefined' ? mozProxies : false
 const globalObj = typeof window === 'undefined' ? globalThis : window
-const getOwnPropertyDescriptor = Object.getOwnPropertyDescriptor
 const functionToString = Function.prototype.toString
-const objectKeys = Object.keys
 
+/**
+ * @deprecated use the ContentFeature method instead
+ */
 export function defineProperty (object, propertyName, descriptor) {
```
jonathanKingston, line 10: How often is this used still?

src/features/runtime-checks.js
jonathanKingston: Thanks for checking :)

src/wrapper-utils.js
```diff
@@ -74,81 +75,3 @@ export function wrapFunction (functionValue, realTarget) {
...
-}
-
-/**
- * Wrap a method descriptor. Only for function properties. For data properties, use wrapProperty(). For constructors, use wrapConstructor().
- * @param {any} object - object whose property we are wrapping (most commonly a prototype)
- * @param {string} propertyName
- * @param {(originalFn, ...args) => any } wrapperFn - wrapper function receives the original function as the first argument
- * @returns {PropertyDescriptor|undefined} original property descriptor, or undefined if it's not found
- */
-export function wrapMethod (object, propertyName, wrapperFn) {
```
jonathanKingston, line 125: I'm assuming these have had as close to near zero changes to move into ContentFeature or should I be checking much more closely they're like for like?

muodov: Yup, no changes except the added addDebugFlag() call

src/wrapper-utils.js
muodov: only by `overloadDescriptor()` inside `DDGProxy`

jonathanKingston: Can you raise a task to make this use the same base code? The one inside contentFeature can use this util or something?

jonathanKingston (approved):

PR #654, opened 2023-07-26
Debug flags all features

muodov:

This PR is a follow-up on https://github.com/duckduckgo/content-scope-scripts/pull/652

It adds debug breakage flags to all remaining features except DuckPlayer (that is currently being refactored in a parallel PR)

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -73,6 +73,8 @@ const afterPageLoad = new Promise(resolve => { afterPageLoadResolver = resolve }
 // we need a module scoped reference.
 /** @type {import("@duckduckgo/messaging").Messaging} */
 let _messagingModuleScope
+/** @type function */
+let _addDebugFlag
 const ctl = {
     /**
      * @return {import("@duckduckgo/messaging").Messaging}
@@ -80,6 +82,11 @@ const ctl = {
     get messaging () {
         if (!_messagingModuleScope) throw new Error('Messaging not initialized')
         return _messagingModuleScope
+    },
+
+    addDebugFlag () {
+        if (!_addDebugFlag) throw new Error('addDebugFlag not initialized')
+        return _addDebugFlag()
     }
 }
 
@@ -662,6 +669,7 @@ function createPlaceholderElementAndReplace (widget, trackingElement) {
 
     // Facebook
     if (widget.replaceSettings.type === 'dialog') {
+        ctl.addDebugFlag()
         ctl.messaging.notify('updateFacebookCTLBreakageFlags', { ctlFacebookPlaceholderShown: true })
         if (widget.shouldUseCustomElement()) {
             /**
@@ -703,6 +711,7 @@ function createPlaceholderElementAndReplace (widget, trackingElement) {
 
     // YouTube
     if (widget.replaceSettings.type === 'youtube-video') {
+        ctl.addDebugFlag()
         ctl.messaging.notify('updateYouTubeCTLAddedFlag', { youTubeCTLAddedFlag: true })
         replaceYouTubeCTL(trackingElement, widget)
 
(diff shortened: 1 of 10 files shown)


src/features/fingerprinting-canvas.js
```diff
@@ -29,8 +29,11 @@ export default class FingerprintingCanvas extends ContentFeature {
             clearCache(canvas)
         }
 
+        // proxy methods are bound to the handler, so we need a closure reference to `this`
+        const addDebugFlag = this.addDebugFlag.bind(this)
         const proxy = new DDGProxy(featureName, HTMLCanvasElement.prototype, 'getContext', {
             apply (target, thisArg, args) {
+                addDebugFlag()
```
muodov, line 36: I didn't instrument `DDGProxy` because chances are that it will be hidden under `.wrapMethod`, and therefore won't need the debug flags

src/features/click-to-load.js
```diff
@@ -662,6 +669,7 @@ function createPlaceholderElementAndReplace (widget, trackingElement) {
 
     // Facebook
     if (widget.replaceSettings.type === 'dialog') {
+        ctl.addDebugFlag()
```
muodov, line 672: I consulted with @kzar and we agreed to make this flag just match the existing breakage flags.

src/features/fingerprinting-canvas.js
jonathanKingston: This makes all those a second class citizen despite being more mature. I think we should fix them tbh.

src/features/click-to-load.js
jonathanKingston: The intent is to drop the other one eventually right?

muodov: yes, I expect to convert the existing breakage flags to the generic format

src/features/runtime-checks.js
```diff
@@ -497,133 +475,6 @@ function isRuntimeElement (element) {
     return false
 }
 
-function overloadGetOwnPropertyDescriptor () {
```
jonathanKingston, line 500: Moving code around like this really belongs in it's own refactoring PR. The diff here would be much easier to understand passing in the `ContentFeature` `this`.

The rationale of not having these within the class is to prevent usages of `this` given the nested nature of them.

I'd prefer for now we just revert these changes specifically because it's a feature not in active development with perhaps less tests we'd like.

Moving big blocks of code around IMO should be avoided within some other piece of work. The time to review these meaningfully is very costly to ensure there is no shadowing and the code has correctly moved across.

As a quicker fix just set a file scope global once the init/load method is fired and wire through the `ContentFeature` `this`.

src/features/web-compat.js
```diff
@@ -12,96 +11,97 @@ function windowSizingFix () {
...
-            },
-            configurable: true,
-            enumerable: true
-        })
-    } catch {
-        // Ignore exceptions that could be caused by conflicting with other extensions
-    }
-}
-
+// TODO: verify that all descriptor shapes match the original (i.e. we're not overriding value descriptors with get/set descriptors)
```
jonathanKingston, line 14: window.safari stuff is correct, navigator.credentials is wrong.

Change this comment to a TODO more local to the navigator.credentials.

src/features/runtime-checks.js
muodov: I disagree that moving code should be avoided. I personally much prefer these changes to changes to the API signatures and calling mechanisms. Those have longer lasting consequences (limit further code changes), look confusing for first-time readers, and are essentially a tech debt. Git(hub) is a powerful tool, one can always track changes down, no matter how big or old they are.
As for the review costs, I intentionally made this move in a separate commit, so you can exclude it from the diff.
That said, I agree with your argument about using `this` inside nested functions, and since this is the only blocking thing, I just made the change.

src/features/element-hiding.js
```diff
@@ -334,4 +297,43 @@ export default class ElementHiding extends ContentFeature {
...
     }
+
+    /**
+     * Apply relevant hiding rules to page at set intervals
+     * @param {Object[]} rules
+     * @param {string} rules[].selector
+     * @param {string} rules[].type
+     */
+    applyRules (rules) {
+        this.addDebugFlag()
```
jonathanKingston, line 308: This will always fire when we have hide rules.

I would suggest moving it to `collapseDomNode` instead and leaving this method.

jonathanKingston (approved):

PR #655, opened 2023-07-28
feat(webcompat): proxy webkit message handlers

shakyShane:
@jonathanKingston the idea here would be:

- 1) allow a list of known handlers to us ~(we could pass it via config later)~ using remote config
- 2) allow certain handlers to return 'undefined'
- 3) allow other handlers to return a polyfilled `.postMessage`

This will fix a known breakage, but is less-than-ideal in the longer term.

Privacy config PR: https://github.com/duckduckgo/privacy-configuration/pull/1155

Known problems:

- what if a site expects a particular result from postMessage?
- the presence of window.webkit itself is the larger issue

### DIFF
--- a/src/features/web-compat.js
+++ b/src/features/web-compat.js
@@ -83,6 +83,9 @@ export default class WebCompat extends ContentFeature {
         if (this.getFeatureSettingEnabled('safariObject')) {
             this.safariObjectFix()
         }
+        if (this.getFeatureSettingEnabled('messageHandlers')) {
+            this.messageHandlersFix()
+        }
     }
 
     /**
@@ -170,6 +173,48 @@ export default class WebCompat extends ContentFeature {
         if (this.getFeatureSettingEnabled('permissions')) {
             const settings = this.getFeatureSettingEnabled('permissions')
             permissionsFix(settings)
+        }
+    }
+
+    /**
+     * Support for proxying `window.webkit.messageHandlers`
+     */
+    messageHandlersFix () {
+        const settings = this.getFeatureSetting('messageHandlers')
+
+        // Do nothing if `messageHandlers` is absent
+        if (!globalThis.webkit?.messageHandlers) return
+
+        const proxy = new Proxy(globalThis.webkit.messageHandlers, {
+            get (target, messageName, receiver) {
+                const handlerName = String(messageName)
+
+                // handle known message names, such as DDG webkit messaging
+                if (settings.handlerStrategies.reflect.includes(handlerName)) {
+                    return Reflect.get(target, messageName, receiver)
+                }
+
+                if (settings.handlerStrategies.undefined.includes(handlerName)) {
+                    return undefined
+                }
+
+                if (settings.handlerStrategies.polyfill.includes('*') ||
+                    settings.handlerStrategies.polyfill.includes(handlerName)
+                ) {
+                    return {
+                        postMessage () {
+                            return Promise.resolve({})
+                        }
+                    }
+                }
+                // if we get here, we couldn't handle the message handler name, so we opt for doing nothing.
+                // It's unlikely we'll ever reach here, since `["*"]' should be present
+            }
+        })
+
+        globalThis.webkit = {
+            ...globalThis.webkit,
+            messageHandlers: proxy
         }
     }
 }
--- a/integration-test/playwright/type-helpers.mjs
+++ b/integration-test/playwright/type-helpers.mjs
@@ -58,6 +58,7 @@ export class Build {
     get artifact () {
         const path = this.switch({
             windows: () => 'build/windows/contentScope.js',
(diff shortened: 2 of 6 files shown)


shakyShane:
* main
  * **PR #655** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/655" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/655?utm_source=stack-comment).

src/features/web-compat.js
```diff
@@ -72,6 +72,29 @@ function permissionsFix (settings) {
     }
 }
 
+function messageHandlersFix () {
+    // Do nothing if `messageHandlers` is absent
+    if (!window.webkit?.messageHandlers) return
```
jonathanKingston, line 77: probably should move to using `globalThis` always for node and keep @sammacbeth happy :D

src/features/web-compat.js
```diff
@@ -72,6 +72,29 @@ function permissionsFix (settings) {
...
+
+    // handlers in any page-world user-scripts
+    const knownHandlers = ['trackerDetectedMessage', 'printHandler', 'specialPages']
+
+    window.webkit = {
+        messageHandlers: new Proxy(window.webkit.messageHandlers, {
+            get (target, messageName, receiver) {
+                if (knownHandlers.includes(String(messageName))) {
+                    return Reflect.get(target, messageName, receiver)
```
jonathanKingston, line 86: It seems like there's three potential strategies here we could take (for the short term).

Return undefined, return a blank post message and return the real listener. Is it easy enough to make each three configurable based on name passed in from the config.

shakyShane: @jonathanKingston yes I like the idea of more things configurable, but I'm not exactly sure what you mean - are you talking about when we return for the dynamic name part? 

From testing on that site (and their less-than-ideal js) we need to ensure the entire access path of `window.webkit.messageHandlers.<name>.postMessage` returns a function, otherwise it will continue to throw

jonathanKingston: We can drop the third one I mentioned as you say just having window.webkit is less than ideal anyway that we should fix in itself.

jonathanKingston: Strategies are what we return:

1. undefined
2. `{ postMessage () {...}}`
3. `Reflect.get(target, messageName, receiver)`

Then I was thinking the config would be something like:
```
{
    "webCompat": {
        "settings": {
            "messageHandlers" {
                "state": "enabled",
                "handlerStrategies": {
                    undefined: ['boop'],
                    reflect: ['trackerDetectedMessage', 'printHandler', 'specialPages'],
                    polyfil: ['*']
                }
            }
        }
    }
}
```

src/features/web-compat.js
```diff
@@ -72,6 +72,29 @@ function permissionsFix (settings) {
...
 }
 
+function messageHandlersFix () {
+    // Do nothing if `messageHandlers` is absent
+    if (!window.webkit?.messageHandlers) return
+
+    // handlers in any page-world user-scripts
+    const knownHandlers = ['trackerDetectedMessage', 'printHandler', 'specialPages']
+
+    window.webkit = {
```
jonathanKingston, line 82: Do we need to override the whole object here? If the site for some reason has added stuff we're going to wipe it.

src/features/web-compat.js
```diff
@@ -72,6 +72,29 @@ function permissionsFix (settings) {
...
+        messageHandlers: new Proxy(window.webkit.messageHandlers, {
```
jonathanKingston, line 83: Should this be DDGProxy instead? or defineProperty or another of the wrappers. These handle toString and now debug flags.

src/features/web-compat.js
shakyShane: we can't assign directly to `messageHandler` unfortunately - let me try a couple of other ways

jonathanKingston: `defineProperty` might work?

src/features/web-compat.js
shakyShane: I couldn't get DDGProxy to work - but I'll try the others

src/features/web-compat.js
shakyShane: @jonathanKingston yep that works.

shakyShane:
@jonathanKingston I removed the validation stuff, should be closer now

jonathanKingston:
Looks good, is it still poc or can we merge?

shakyShane:
@jonathanKingston given it's 100% guarded behind config, I think it's good to merge ASAP, if you can R+

jonathanKingston (approved):
LGTM thanks for adding in the flexibility to configure.

PR #667, opened 2023-08-02
Enable max debug count

jonathanKingston:
See: https://app.asana.com/0/892838074342800/1205174622371353/f

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -537,8 +537,8 @@ function numberOfTimesDebugged (feature, message) {
 
 const DEBUG_MAX_TIMES = 1000
 
-export function postDebugMessage (feature, message) {
-    if (!debug) {
+export function postDebugMessage (feature, message, allowNonDebug = false) {
+    if (!debug && !allowNonDebug) {
         return
     }
     if (numberOfTimesDebugged(feature, message) > DEBUG_MAX_TIMES) {
--- a/src/utils.js
+++ b/src/utils.js
@@ -520,7 +520,29 @@ export class DDGProxy {
     }
 }
 
+const maxCounter = new Map()
+function numberOfTimesDebugged (feature, message) {
+    let key
+    try {
+        // Could throw if message is not serializable
+        key = JSON.stringify({ feature, message })
+    } catch {
+        return 0
+    }
+    if (!maxCounter.has(key)) {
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -520,7 +520,29 @@ export class DDGProxy {
     }
 }
 
+const maxCounter = new Map()
+function numberOfTimesDebugged (feature, message) {
+    let key
+    try {
+        // Could throw if message is not serializable
+        key = JSON.stringify({ feature, message })
```
muodov, line 526: I'm guessing `JSON.stringify` in a loop could be taxing too. Besides, any difference in message would create a new key. Could we use a simpler key construction method? Perhaps just `feature`?

src/utils.js
```diff
@@ -520,7 +520,29 @@ export class DDGProxy {
...
+    } else {
+        maxCounter.set(key, maxCounter.get(key) + 1)
+    }
+    return maxCounter.get(key)
+}
+
+const DEBUG_MAX_TIMES = 1000
+
 export function postDebugMessage (feature, message) {
+    if (numberOfTimesDebugged(feature, message) > DEBUG_MAX_TIMES) {
```
muodov, line 544: I guess rate-limiting instead of a max_times would not be helpful for debugging?

muodov (approved):

src/utils.js
jonathanKingston: As in just preventing _n_ messages over _y_ window of time from happening?

I think that would be less clear unless we provided info about the truncation, the alternative is telling the debugger we've rate limited but that involves changes on the extension too, on page load we get a lot of messages sometimes so that threshold initially might be very high.

Ultimately this is just a quick fix to prevent it breaking the debugger we could always revisit it.

src/utils.js
jonathanKingston: Serializing isn't too concerning to me in comparison with swamping the messaging to another process a few times.

Feature seemed a little too coarse, the only concern I had is that the `jsException` is still serialized always (all others are gated with a `debug` flag).

- Maybe just up the limit to 5k and use feature is simpler though.

I actually removed this change but I think we should bring it back. If we're checking `debug` worst case we're just going to reintroduce a similar problem for people debugging with memory exhaustion.

PR #670, opened 2023-08-03
Android webcompat fix

jonathanKingston:
This code was dropped in https://github.com/duckduckgo/content-scope-scripts/commit/f644105b9aaf6f2f0e9dd1c38c85d27ea2c9a06a through a merge issue.

This change adds a test to verify the shape along with changing to use `defineProperty` for `window.Notification`. As we're shimming raw values this isn't possible for the permissions API.

Note locally I'm having cached config issues from Android, I don't know the cause so needs someone else to verify there's not another issue here. (I was getting only GPC enabled setup)

See: https://app.asana.com/0/0/1205196878614274/f

### DIFF
--- a/src/features/web-compat.js
+++ b/src/features/web-compat.js
@@ -51,8 +51,12 @@ export default class WebCompat extends ContentFeature {
         Notification.maxActions = 2
 
         // Expose the API
-        // @ts-expect-error window.Notification isn't assignable
-        window.Notification = Notification
+        this.defineProperty(window, 'Notification', {
+            value: Notification,
+            writable: true,
+            configurable: true,
+            enumerable: false
+        })
     }
 
     /**
--- a/src/features/web-compat.js
+++ b/src/features/web-compat.js
@@ -11,67 +11,6 @@ function windowSizingFix () {
     window.outerWidth = window.innerWidth
 }
 
-/**
- * Notification fix for adding missing API for Android WebView.
- */
-function notificationFix () {
-    if (window.Notification) {
-        return
-    }
-    const Notification = () => {
-        // noop
-    }
-    Notification.requestPermission = () => {
-        return Promise.resolve({ permission: 'denied' })
-    }
-    Notification.permission = 'denied'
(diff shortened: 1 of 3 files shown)


src/features/web-compat.js
```diff
@@ -51,8 +51,12 @@ export default class WebCompat extends ContentFeature {
...
 
         // Expose the API
-        // @ts-expect-error window.Notification isn't assignable
-        window.Notification = Notification
+        this.defineProperty(window, 'Notification', {
+            value: Notification,
+            writable: true,
+            configurable: true,
+            enumerable: false
+        })
```
muodov, line 59: You could use `wrapProperty()` here to get runtime shape checks

jonathanKingston: No we can't it doesn't exist: https://github.com/duckduckgo/content-scope-scripts/blob/11e62c5311f705ad0b461184a614daaaf5f9f32a/src/content-feature.js#L320-L322

src/features/web-compat.js
```diff
@@ -85,7 +84,8 @@ export default class WebCompat extends ContentFeature {
             'midi'
         ]
         const validPermissionNames = settings.validPermissionNames || defaultValidPermissionNames
-        window.navigator.permissions.query = (query) => {
+        permissions.query = (query) => {
```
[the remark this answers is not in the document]
jonathanKingston: I think we're very fine here:

- If `harmfulApis` runs first `wrapMethod` will short circuit as it doesn't exist.
    - `webCompat` then runs and we create the API with the expected behaviour. 
- If `webCompat` runs first this overloads the property and provides `PermissionStatus` which is disabled,
    - `harmfulApis` then runs and we just blindly overwrite the response with a plain object (I think we should fix this to return a class).

Either way it's off and there's no clear intent to turn on globally to include Android anytime soon.

src/features/web-compat.js
muodov: ah, right, sorry

muodov (approved):

PR #683, opened 2023-08-14
Element hiding: wrap selectors in :is(...) to make them forgiving

dharb:
Asana link: https://app.asana.com/0/1203557590179903/1204908558710325/f

We're using one set of rules across several browser engines, which means we can only use selectors that are supported everywhere. This can be a nuisance, as we saw in https://github.com/duckduckgo/privacy-configuration/pull/1062 where we wanted to hide a 'Switch to Chrome' popup appearing in our desktop browsers and Edge using `:has`, but `:has` is yet to be fully implemented in Firefox.

Wrapping all of the selectors we use in `:is(...)` will make it so we don't have to worry about unsupported selectors throwing errors - instead they'll silently fail when they're unsupported.


src/features/element-hiding.js
```diff
@@ -210,7 +212,7 @@ function injectStyleTag (rules) {
         }
     })
 
-    styleTagContents = styleTagContents.concat('{display:none!important;min-height:0!important;height:0!important;}')
+    styleTagContents = styleTagContents.concat('){display:none!important;min-height:0!important;height:0!important;}')
```
jonathanKingston, line 215: This seems like it might be easily break in future. Could we create a variable for the selector specifically something like:

```suggestion
    let selector = ''
    rules.forEach((rule, i) => {
        if (i !== rules.length - 1) {
            selector = selector.concat(rule.selector, ',')
        } else {
            selector = selector.concat(rule.selector)
        }
    })
    const styleTagProperties = 'display:none!important; min-height:0!important; height:0!important;'
    const styleTagContents = `${forgivingSelector(selector)} {${styleTagProperties}}`
    injectGlobalStyles(styleTagContents)
```


```javascript
function forgivingSelector (selector) {
    return `is(${selector})`
}
```

dharb: Good call, thanks.

dharb:
@jonathanKingston mind giving this another look when you have a moment?

jonathanKingston (approved):

PR #682, opened 2023-08-14
Element hiding: move timeouts to remote config

dharb:
Asana Task/Github Issue: https://app.asana.com/0/1203557590179903/1204126774857760/f

Corresponding remote config PR: https://github.com/duckduckgo/privacy-configuration/pull/1211

This change simply moves the array of timeouts to remote config.

PR #696, opened 2023-08-16
Element hiding: implement several new rule types

dharb:
This PR adds three new utility rule types to element hiding:
- `disable-default`
  - Commonly known as `generichide` by popular adblockers, when this rule is added to a domain it disables all generic rules for the domain. This is used most frequently to diffuse adblock walls.
- `modify-attr`
  - This rule allows us to target a specific selector and change attributes on the selector. These rules consist of a selector, the rule type, and an array of properties and values. eg:
```
{
    "selector": "html",
    "type": "modify-attr",
    "values": [
        {
            "property": "data-leaderboard-is-fixed",
            "value": "false"
        }
    ]
}
```
- `modify-style`
  - This rule allows us to target a specific selector and change styles on the selector. These rules consist of a selector, the rule type, and an array of properties and values. eg:
```
{
    "selector": "html",
    "type": "modify-attr",
    "values": [
        {
            "property": "overflow-y",
            "value": "auto"
        }
    ]
}
```


src/features/element-hiding.js
```diff
@@ -343,6 +396,7 @@ export default class ElementHiding extends ContentFeature {
         setTimeout(() => {
             appliedRules = new Set()
             hiddenElements = new WeakMap()
-        }, 3100)
+            modifiedElements = new WeakMap()
+        }, clearCacheTimer)
```
dharb, line 400: This was an oversight on my part, really should have been included in #682.

src/features/element-hiding.js
```diff
@@ -25,8 +26,9 @@ function collapseDomNode (element, rule, previousElement) {
     }
     const type = rule.type
     const alreadyHidden = hiddenElements.has(element)
-
-    if (alreadyHidden) {
+    const alreadyModified = modifiedElements.has(element) && modifiedElements.get(element) === rule.type
```
dharb, line 29: Don't want to accidentally apply the same modifications to the same element multiple times, but also want to be able to apply a `modify-style` rule and a `modify-attr` rule to the same element if needed.

src/features/element-hiding.js
```diff
@@ -318,11 +368,14 @@ export default class ElementHiding extends ContentFeature {
...
     applyRules (rules) {
         const timeoutRules = extractTimeoutRules(rules)
+        // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
+        const clearCacheTimer = unhideTimeouts.at(-1) + 100
 
         // several passes are made to hide & unhide elements. this is necessary because we're not using
         // a mutation observer but we want to hide/unhide elements as soon as possible, and ads
         // frequently take from several hundred milliseconds to several seconds to load
         // check at 0ms, 100ms, 200ms, 300ms, 400ms, 500ms, 1000ms, 1500ms, 2000ms, 2500ms, 3000ms
+        console.log('rules to be applied', timeoutRules)
```
jonathanKingston, line 378: Remember to remove this for merging :)

src/features/element-hiding.js
jonathanKingston: I'm not sure this actually is guaranteed to finish after unhideTimeout -1.

I don't think it matters all that much but it might? We could resolve promises in the timers and await them all if we want to be sure.

dharb: No you're right, it doesn't have to be exact - we just want it to be somewhat close to the upper bound of our checks. Let's say we end up clearing these at 12100ms:
- The last iteration of `collapseDomNode` happens with a 10000ms timeout. Seems unlikely that this would actually happen a full 2100ms later, but in the off chance it did, [this](https://github.com/duckduckgo/content-scope-scripts/blob/5cead52e6b95cfde349279d0e895b1106da31c50/src/features/element-hiding.js#L31) wouldn't return for an element that has already been hidden/modified, so we'd end up checking/hiding/modifying elements that had already been addressed. All of the timing based modifications we're making here are idempotent, so this wouldn't cause any issues.
- The last iteration of `expandNonEmptyDomNode` happens with a 12000ms timeout. If this were to occur after caches have been cleared, it likely wouldn't do anything, since `unhideNode` is only run if `alreadyHidden` is true, and `alreadyHidden` can only be true if `collapseDomNode` was run again after the caches were cleared. In the off chance `alreadyHidden` is actually true here, this would just check the hidden elements and unhide them if they contain visible content.

src/features/element-hiding.js
```diff
@@ -25,13 +27,12 @@ function collapseDomNode (element, rule, previousElement) {
...
     const alreadyHidden = hiddenElements.has(element)
-
-    if (alreadyHidden) {
+    const alreadyModified = modifiedElements.has(element) && modifiedElements.get(element) === rule.type
+    // return if the element has already been hidden, or modified by the same rule type
+    if (alreadyHidden || alreadyModified) {
         return
     }
 
-    featureInstance.addDebugFlag()
```
dharb, line 33: This was firing every time `collapseDomNode` runs, so I moved it into `hideNode` so that it only fires when something is actually hidden; and I added a check so it only fires once.

src/features/element-hiding.js
```diff
@@ -109,6 +116,11 @@ function hideNode (element) {
     element.style.setProperty('min-height', '0px', 'important')
     element.style.setProperty('height', '0px', 'important')
     element.hidden = true
+
+    if (!debugFlagAdded) {
```
jonathanKingston, line 120: This isn't needed, we should only trigger this once.

src/features/element-hiding.js
```diff
@@ -318,11 +370,12 @@ export default class ElementHiding extends ContentFeature {
      */
     applyRules (rules) {
         const timeoutRules = extractTimeoutRules(rules)
+        // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
+        const clearCacheTimer = unhideTimeouts.at(-1) + 100
```
jonathanKingston, line 374: In case we change the ordering or which has the larger number we should consider joining all timers and finding the max. Something like:

```suggestion
        const clearCacheTimer = unhideTimeouts.concat(hideTimeouts).reduce((a,b) => Math.max(a,b), 0) + 100
```

dharb:
Thanks for the review @jonathanKingston, concerns addressed.

jonathanKingston (approved):

PR #719, opened 2023-09-05
Capture globals generically

jonathanKingston:
See: https://app.asana.com/0/72649045549333/1204999114961040/f

Also resolves the core issue of: https://app.asana.com/0/892838074342800/1205340475534185/f
- To test this part you have to revert the fix @sammacbeth made and do the testing steps there including using a VPN.


src/content-scope-features.js
```diff
@@ -89,7 +89,7 @@ function injectFeatures (args) {
         ${codeFeatures.join('\n')}
     })();`
     script.src = 'data:text/javascript;base64,' + btoa(code)
-    getInjectionElement().appendChild(script)
+    getInjectionElement()?.appendChild(script)
```
shakyShane, line 92: Was this just noticed during testing and fixed on the fly?

shakyShane (commented):
@jonathanKingston I know we've been over this problem a few times and this implementation looks fine from a code POV.

The only reason I didn't give an explicit approval is based on not knowing the scope of testing before merging? For instance, is this something you intend to test cross-platform *before* we merge, or did you want to cut a release and test as each platforms updates?

I don't have any preferences here, just curious what the plan is there (if you have one yet :))

src/content-scope-features.js
jonathanKingston: I changed the method signature to return null also.

The reason this was added is the utils.js is used within the extension testing which doesn't have a DOM, so the code falls back to just using the existing global (under a different alias so still some slight improvements).

jonathanKingston:
> The only reason I didn't give an explicit approval is based on not knowing the scope of testing before merging? For instance, is this something you intend to test cross-platform _before_ we merge, or did you want to cut a release and test as each platforms updates?


There's been a fair amount of testing on each platform locally, some changes have happened since though so happy to just verify as we roll out.
I don't think there's significant risk on other platforms other than Firefox, which I've just retested and spotted issues again :/

jonathanKingston:
Marking as on draft as on hold right now.

PR #724, opened 2023-09-07
Perform the document type check earlier

kzar:
Before loading the content-scope-script features for a document, we
check if the document is HTML/XHTML. That way, we can avoid attempt to
load the features for other types of documents. Unfortunately, it
turns out that some websites mess with  `document.__proto__` which
breaks tests like `document instanceof HTMLDocument`.

Let's perform those checks earlier therefore. For now, let's run the
checks before the load() function (instead of during the load(),
init() and update() calls). In the future, we might consider running
the checks even earlier, so that they run in the "isolated
world" (instead of "main world") for the Chrome extension. That would
be preferable, but requires more plumbing.

### DIFF
--- a/src/content-scope-features.js
+++ b/src/content-scope-features.js
@@ -5,24 +5,24 @@ import { PerformanceMonitor } from './performance'
 import injectedFeaturesCode from 'ddg:runtimeInjects'
 import platformFeatures from 'ddg:platformFeatures'
 
-function shouldRun () {
-    // don't inject into non-HTML documents (such as XML documents)
-    // but do inject into XHTML documents
-    // Should check HTMLDocument as Document is an alias for XMLDocument also.
-    if (document instanceof HTMLDocument === false && (
-        document instanceof XMLDocument === false ||
-        document.createElement('div') instanceof HTMLDivElement === false
-    )) {
-        return false
-    }
-    return true
-}
-
 let initArgs = null
 const updates = []
 const features = []
 const alwaysInitFeatures = new Set(['cookie'])
 const performanceMonitor = new PerformanceMonitor()
+
+// It's important to avoid enabling the features for non-HTML documents (such as
+// XML documents that aren't XHTML). Note that it's necessary to check the
+// document type in advance, to minimise the risk of a website breaking the
+// checks by altering document.__proto__. In the future, it might be worth
+// running the checks even earlier (and in the "isolated world" for the Chrome
+// extension), to further reduce that risk.
+const isHTMLDocument = (
+    document instanceof HTMLDocument || (
+        document instanceof XMLDocument &&
+            document.createElement('div') instanceof HTMLDivElement
+    )
+)
 
 /**
  * @typedef {object} LoadArgs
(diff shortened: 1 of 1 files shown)


src/content-scope-features.js
```diff
@@ -5,30 +5,23 @@ import { PerformanceMonitor } from './performance'
...
 let initArgs = null
 const updates = []
 const features = []
 const alwaysInitFeatures = new Set(['cookie'])
 const performanceMonitor = new PerformanceMonitor()
 
+// The document type is checked as soon as possible, then provided as a load
+// argument. That way, it minimises the risk of a website breaking our
+// instanceof checks by changing document.__proto__.
+let isHTMLDocument = null
```
jonathanKingston, line 17: Why not just import the var here than wire it through args adding a bunch of complexity?

jonathanKingston: Equally the simplest change is:
```suggestion
const shouldRun = shouldRun()
```

And just change the code to use the variable below:
`if (!shouldRun) {`

jonathanKingston: (equally once tested we should keep your refactor of the method as it's way more readable I've just left it in how it was for so long)

kzar: Done

src/content-scope-features.js
```diff
@@ -5,25 +5,24 @@ import { PerformanceMonitor } from './performance'
...
 const features = []
 const alwaysInitFeatures = new Set(['cookie'])
 const performanceMonitor = new PerformanceMonitor()
 
+// Check the document type in advance. That way, it minimises the risk of a
+// website breaking the `document instanceof ...` checks by changing
+// document.__proto__. In the future, it might be worth moving this to run even
+// earlier (and in the "isolated world" for the Chrome extension), to further
+// reduce that risk.
+const isHTMLDocument = (
```
jonathanKingston, line 19: Can you leave in the comments about not wanting to inject into XML, that's super important to explain to newcomers why we don't run. You can reword it etc, just explain why we're doing the check at all.

- Maybe you've removed it as it's now just a variable that doesn't represent `shouldRun`?
   - If so I would consider renaming it that and making the code a function as you had it before?

I'd reduce this new comment wording a little tbh; I don't really think the 'in the future' bit is needed; if you really wish raise an issue for that.

kzar: I've added the part of the comment back that you wanted and cut the word count down a bit.

jonathanKingston: Can you just remove the last sentence and raise an issue? Nice to haves aren't really needed in the comments imo.
This check doesn't relate to 'privacy protections', can you reword somehow?

kzar: >  This check doesn't relate to 'privacy protections', can you reword somehow?

Well, this variable is checked before calling load(), init() and update() for the content-scope-script features, and those features are primarily privacy protections. I don't mind rewording that again if you'd like, but would you like to make a suggestion?

> Can you just remove the last sentence and raise an issue?

I would prefer to keep that, I think it's useful. If in the future this issue comes up again, the comment will give us more of a clue.

kzar: Done

jonathanKingston (approved):

PR #725, opened 2023-09-08
post-process apple bundle

shakyShane:
https://app.asana.com/0/1201614831475344/1205448470130004/f

### DIFF
--- a/packages/messaging/lib/webkit.js
+++ b/packages/messaging/lib/webkit.js
@@ -59,11 +59,6 @@ import { isResponseFor, isSubscriptionEventFor } from '../schema.js'
  * @implements {MessagingTransport}
  */
 export class WebkitMessagingTransport {
-    /** @type {WebkitMessagingConfig} */
-    config
-    /** @internal */
-    globals
-
     /**
      * @param {WebkitMessagingConfig} config
      * @param {import('../index.js').MessagingContext} messagingContext
--- a/packages/special-pages/package.json
+++ b/packages/special-pages/package.json
@@ -23,7 +23,7 @@
     "@playwright/test": "^1.36.2",
     "http-server": "^14.1.1",
     "@duckduckgo/messaging": "*",
-    "esbuild": "^0.18.19",
+    "esbuild": "^0.19.2",
     "web-resource-inliner": "^6.0.1"
   }
 }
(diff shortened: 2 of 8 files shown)


shakyShane:
* main
  * **PR #725** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/725" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/725?utm_source=stack-comment).

packages/messaging/lib/webkit.js
```diff
@@ -59,11 +59,6 @@ import { isResponseFor, isSubscriptionEventFor } from '../schema.js'
  * @implements {MessagingTransport}
  */
 export class WebkitMessagingTransport {
-    /** @type {WebkitMessagingConfig} */
-    config
```
jonathanKingston, line 63: Do we need these changes still?

shakyShane: These were always assigned in the constructor, so currently served no purpose.

jonathanKingston: I guess, just seems unrelated but it's fine.

packages/special-pages/package.json
```diff
@@ -23,7 +23,7 @@
     "@playwright/test": "^1.36.2",
     "http-server": "^14.1.1",
     "@duckduckgo/messaging": "*",
-    "esbuild": "^0.18.19",
```
jonathanKingston, line 26: Can this just inherit instead?

shakyShane: yeah I think we can move to top-level and then workspaces will make it work for special pages too :)

packages/messaging/lib/webkit.js
shakyShane: it's not unrelated - these were the first errors I encountered (until I dug deeper), so when evaluating them and realising they were not needed, made sense to remove as part of this work :)

jonathanKingston (approved):

shakyShane:
Thanks @jonathanKingston just going through last-min testing on Catalina - found something else problematic, but hopefully not much delay here

PR #731, opened 2023-09-13
Update russian translation

Blueberryy:


ladamski (approved):
Approved per discussion with @GioSensation

PR #782, opened 2023-10-17
Shim navigator.mediaSession and navigator.presentation APIs for Android

ladamski:
https://app.asana.com/0/0/1205678675729282/f

update: privacy config PR: https://github.com/duckduckgo/privacy-configuration/pull/1410

### DIFF
--- a/src/features/web-compat.js
+++ b/src/features/web-compat.js
@@ -32,6 +32,12 @@ export default class WebCompat extends ContentFeature {
             const settings = this.getFeatureSettingEnabled('permissions')
             this.permissionsFix(settings)
         }
+
+        // TODO ANDROID if (this.getFeatureSettingEnabled('mediaSession')) {
+        this.mediaSessionFix()
+
+        // TODO ANDROID if (this.getFeatureSettingEnabled('presentation')) {
+        this.presentationFix()
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/features/web-compat.js
```diff
@@ -184,6 +192,87 @@ export default class WebCompat extends ContentFeature {
...
+                enumerable: true
+            })
+        } catch {
+            // Ignore exceptions that could be caused by conflicting with other extensions
+        }
+    }
+
+    presentationFix () {
+        try {
+            // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
```
jonathanKingston, line 247: I'm not sure what the errors are here but we should fix them or explain why they're not relevant.

https://app.asana.com/0/1201614831475344/1203979574128023/f is a placeholder so we could rapidly adopt TS it shouldn't be used on new code.

ladamski: Error for posterity:
```
src/features/web-compat.js:245:34 - error TS2339: Property 'presentation' does not exist on type 'Navigator'.
```

jonathanKingston: > Those were already present throughout web-compat.js

Yeah they're temporary, this message is a TODO when we adopted TS.

> but I get the impression that TS isn't aware of some of these new APIs so it considers them to be undefined objects

Sure it only has what is cross platform and mature.

I would just change the comment to explain that the interface is missing:
```suggestion
            // @ts-expect-error navigator.presentation isn't defined in TS
```

jonathanKingston (approved):

PR #810, opened 2023-11-02
Fix passing of permission settings

jonathanKingston:
See: https://app.asana.com/0/1200277586140538/1205860617092627/f


dharb (approved):
👍

PR #818, opened 2023-11-06
Separate messaging and clean

jonathanKingston:
https://app.asana.com/0/1201614831475344/1205868341160053/f

### DIFF
--- a/inject/android.js
+++ b/inject/android.js
@@ -5,7 +5,7 @@
 import { load, init } from '../src/content-scope-features.js'
 import { processConfig, isGloballyDisabled } from './../src/utils'
 import { isTrackerOrigin } from '../src/trackers'
-import { AndroidMessagingConfig } from '../packages/messaging/index.js'
+import { AndroidMessagingConfig } from '@duckduckgo/messaging'
 
 function initCode () {
     // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
--- a/inject/apple.js
+++ b/inject/apple.js
@@ -5,7 +5,7 @@
 import { load, init } from '../src/content-scope-features.js'
 import { processConfig, isGloballyDisabled } from './../src/utils'
 import { isTrackerOrigin } from '../src/trackers'
-import { WebkitMessagingConfig, TestTransportConfig } from '../packages/messaging/index.js'
+import { WebkitMessagingConfig, TestTransportConfig } from '@duckduckgo/messaging'
 
 function initCode () {
     // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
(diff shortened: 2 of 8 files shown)


inject/apple.js
```diff
@@ -18,7 +19,15 @@ function initCode () {
         trackerLookup: processedConfig.trackerLookup,
         documentOriginIsTracker: isTrackerOrigin(processedConfig.trackerLookup),
         site: processedConfig.site,
-        bundledConfig: processedConfig.bundledConfig
+        bundledConfig: processedConfig.bundledConfig,
+        constructMessagingConfig: (context) => {
+            if (import.meta.injectName !== 'apple-isolated') throw new Error('Only apple-isolated messaging supported')
+            return new WebkitMessagingConfig({
+                webkitMessageHandlerNames: [context.context],
```
shakyShane, line 26: in C-S-S, this is a known value, so I'd remove the wiring of context and just hard-code it here, or get it from `import.meta` or something. I only re-used the context name because I could (at the time) but if it's awkward let's remove.

jonathanKingston: We actually don't need this to be special anywhere 'debug' isn't used by the extension so we can simplify the context creation too.

jonathanKingston: Removed it being a method entirely due to the removal of the context too.

src/content-feature.js
```diff
@@ -94,33 +92,38 @@ export default class ContentFeature {
...
         })
-        return this.#messagingContext
     }
 
-    // Messaging layer between the content feature and the Platform
-    get debugMessaging () {
-        if (this.#debugMessaging) return this.#debugMessaging
-
-        if (this.platform?.name === 'extension' && typeof import.meta.injectName !== 'undefined') {
-            this.#debugMessaging = createMessaging({ name: 'debug', isDebug: this.isDebug }, import.meta.injectName)
```
muodov, line 119: Note that debug messaging was separated because it had a separate 'feature' name (`'debug'` instead of `this.name`). This was supposed to allow the native platform side write a single message handler for breakage flags instead of duplicating the same one per every C-S-S feature (even those that don't really use messaging otherwise).
The only consumer at the moment is the extension though, where this is not a big deal, but on other platforms it could be important

jonathanKingston: I think we can approach that when we come to it. I don't really know each platforms integrations but that sounds a little fragile if they're implemented that way.

Either way we can create a singleton and pass it in but we'd risk a tonne of duplication again.

Let's get @shakyShane to confirm, I'd rather we don't do this for now.

muodov (dismissed):

inject/android.js
```diff
@@ -2,19 +2,10 @@
...
-const allowedMessages = [
-    'getClickToLoadState',
-    'getYouTubeVideoDetails',
-    'openShareFeedbackPage',
-    'setYoutubePreviewsEnabled',
-    'unblockClickToLoadContent',
-    'updateYouTubeCTLAddedFlag',
-    'updateFacebookCTLBreakageFlags'
-]
+import { AndroidMessagingConfig } from '../packages/messaging/index.js'
```
muodov, line 8: should this be `import { AndroidMessagingConfig } from '@duckduckgo/messaging'`?

jonathanKingston: Sure, done

shakyShane (approved):

PR #825, opened 2023-11-09
Shim Web Share API for Android WebView

muodov:
https://app.asana.com/0/1163321984198618/1205583865502213/f

---
  - https://app.asana.com/0/0/1205583865502217
  - https://app.asana.com/0/0/1205907591966293


src/features/web-compat.js
```diff
@@ -43,6 +67,64 @@ export default class WebCompat extends ContentFeature {
...
+                    return Promise.reject(new DOMException('Share must be initiated by a user gesture', 'InvalidStateError'))
+                }
+
+                const dataToSend = {}
+                for (const key of ['title', 'text', 'url']) {
+                    if (key in data) dataToSend[key] = data[key]
+                }
+                if ('url' in data) dataToSend.url = new URL(data.url) // clean url
+
+                // eslint-disable-next-line promise/prefer-await-to-then
```
jonathanKingston, line 106: Why not use await?

muodov: good question. I wanted the return value be the same as #activeShare, but not sure why.. I'll change

muodov: actually on the second thought I do like it this way because it makes a encapsulates #activeShare cleanup in itself. And error handling is kinda nicer then a try..catch, so I'd rather keep it, but let me know if you think it's better another way.

jonathanKingston: One of many variants:

```javascript
                this.#activeShare = this.messaging.request(MSG_WEB_SHARE, dataToSend)
                try {
                      const resp = await this.#activeShare
                      this.#activeShare = null
                      if (resp.failure) {
                          switch (resp.failure.name) {
                          case 'AbortError':
                          case 'NotAllowedError':
                          case 'DataError':
                              throw new DOMException(resp.failure.message, resp.failure.name)
                          default:
                              throw new DOMException(resp.failure.message, 'DataError')
                          }
                      }
                      return resp
                } catch (err) {
                        this.#activeShare = null
                        throw new DOMException(err.message, 'DataError')
                }
```

Or:
```javascript
                this.#activeShare = this.messaging.request(MSG_WEB_SHARE, dataToSend)
                let resp
                try {
                        resp = await this.#activeShare
                } catch (err) {
                        this.#activeShare = null
                        throw new DOMException(err.message, 'DataError')
                }
                this.#activeShare = null
                if (resp.failure) {
                    switch (resp.failure.name) {
                    case 'AbortError':
                    case 'NotAllowedError':
                    case 'DataError':
                        throw new DOMException(resp.failure.message, resp.failure.name)
                    default:
                        throw new DOMException(resp.failure.message, 'DataError')
                    }
                }
                return resp
```

Or if you want the encapsulation (I think you're on about):

```javascript
            async function createWebShare (dataToSend) {
                try {
                      const resp = await this.messaging.request(MSG_WEB_SHARE, dataToSend)
                      if (resp.failure) {
                          switch (resp.failure.name) {
                          case 'AbortError':
                          case 'NotAllowedError':
                          case 'DataError':
                              throw new DOMException(resp.failure.message, resp.failure.name)
                          default:
                              throw new DOMException(resp.failure.message, 'DataError')
                          }
                      }
                      return resp
                } catch (err) {
                        throw new DOMException(err.message, 'DataError')
                }
            }
```

And then in this method:
```javascript
try {
    this.#activeShare = await createWebShare(dataToSend)
    this.#activeShare = null
} catch {
    this.#activeShare = null
}
```

src/features/web-compat.js
```diff
@@ -43,6 +87,59 @@ export default class WebCompat extends ContentFeature {
...
         }
+
+        if (this.getFeatureSettingEnabled('webShare')) {
+            this.shimWebShare()
+        }
+    }
+
+    /** Shim Web Share API in Android WebView */
+    shimWebShare () {
+        if (typeof navigator.canShare === 'function') return
```
jonathanKingston, line 98: Should the if checks be combined?

EG:
```suggestion
        if (typeof navigator.canShare === 'function' || typeof navigator.share === 'function') return
```

Seems like it might be unexpected to just override one.

muodov: agreed, good idea

src/features/web-compat.js
```diff
@@ -11,7 +11,51 @@ function windowSizingFix () {
...
+        // combine everything into text
+        dataToSend.text = `${dataToSend.text} ${dataToSend.url}`
+        delete dataToSend.url
+    }
+    if (!('url' in dataToSend) && !('text' in dataToSend)) {
+        // if there's only title, create a dummy empty text
+        dataToSend.text = ''
+    }
+    return dataToSend
+}
```
jonathanKingston, line 53: Take it or leave it, but this is a little simpler to read (personally):
```suggestion
/**
 * @returns {ShareRequestData}
 */
function cleanShareData (data) {
    const { title, text, url } = data;
    if (url) url = (new URL(url, location.href)).href // clean url
    if (url && text) {
        // combine everything into text
        text = `${text} ${url}`
        url = undefined
    }
    if (!url && !text) {
        // if there's only title, create a dummy empty text
        text = ''
    }
    return { title, text, url }
}
```

src/features/web-compat.js
```diff
@@ -11,7 +11,51 @@ function windowSizingFix () {
...
+        } catch (err) {
+            return false
+        }
+    }
+    if (window !== window.top) return false // Not supported in iframes
+    return true
+}
+
+/** @returns {ShareRequestData} */
+function cleanShareData (data) {
```
jonathanKingston, line 36: Can we explain why we're doing this? Is it to match what Android can support?

src/features/web-compat.js
muodov: There are some caveats with empty strings and undefined values, but I'll take some of your ideas here :)

jonathanKingston: If url / text is empty do we really want to do anything special? Maybe the final fallback to `text=''` is enough?

src/features/web-compat.js
muodov: Yeah, this just minimizes checks on the android side. Also some data manipulation is needed to make it more similar to the Chrome behaviour. For example, concatenating url and text if both are present. It's still not 100% the same as the sharesheet UI differs per device.

src/features/web-compat.js
```diff
@@ -11,7 +11,51 @@ function windowSizingFix () {
...
+    if ('url' in dataToSend && 'text' in dataToSend) {
+        // combine everything into text
+        dataToSend.text = `${dataToSend.text} ${dataToSend.url}`
+        delete dataToSend.url
+    }
+    if (!('url' in dataToSend) && !('text' in dataToSend)) {
+        // if there's only title, create a dummy empty text
+        dataToSend.text = ''
+    }
+    return dataToSend
```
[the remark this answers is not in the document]
muodov: Correct, we don't support files. `canShare()` returns `false` if you request file sharing though, so the API should behave per spec. `navigator.share()` will raise a `DataError`

src/features/web-compat.js
muodov: There are subtle differences. For example, if url is an empty string, the effective value is the current page address.
The internal convention with Android we came up with is to pass a cleaned `url` or `text`, but never both at the same time.

muodov: As discussed, both versions may hide some caveats, but I also think it's a not a too complicated function, so I just added comments in the existing version.

jonathanKingston (approved):
LGTM thanks for all the changes!

PR #843, opened 2023-11-21
DBP: first action

shakyShane:
https://app.asana.com/0/0/1206000767975136/f

**Feature Summary**

Native platforms will send one of the following instructions, one at a time:

- `navigate`
- `extract`
- `click` <- included in this PR
- `expectation`
- `fillForm`
- `getCaptchaInfo`
- `solveCaptcha`

Each of them has a corresponding function that will be executed, and will return either a SuccessResponse or ErrorResponse to native.

The flow looks like this: (taken from the markdown file in the PR)

image

### DIFF
--- a/src/features.js
+++ b/src/features.js
@@ -20,7 +20,8 @@ const otherFeatures = /** @type {const} */([
     'duckPlayer',
     'harmfulApis',
     'webCompat',
-    'windowsPermissionUsage'
+    'windowsPermissionUsage',
+    'brokerProtection'
 ])
 
 /** @typedef {baseFeatures[number]|otherFeatures[number]} FeatureName */
@@ -31,7 +32,8 @@ export const platformSupport = {
         ...baseFeatures
     ],
     'apple-isolated': [
-        'duckPlayer'
+        'duckPlayer',
+        'brokerProtection'
     ],
     android: [
         ...baseFeatures,
@@ -42,7 +44,8 @@ export const platformSupport = {
         'cookie',
         ...baseFeatures,
         'windowsPermissionUsage',
-        'duckPlayer'
+        'duckPlayer',
+        'brokerProtection'
     ],
     firefox: [
         'cookie',
--- a/src/features/broker-protection.js
+++ b/src/features/broker-protection.js
@@ -1,0 +1,21 @@
+import ContentFeature from '../content-feature.js'
+import { execute } from './broker-protection/execute.js'
+
+export default class BrokerProtection extends ContentFeature {
+    init () {
+        this.messaging.subscribe('onActionReceived', (/** @type {any} */params) => {
+            try {
+                const action = params.state.action
+                const data = params.state.data
+                if (!action) {
+                    return this.messaging.notify('actionError', { error: 'No action found.' })
+                }
+                const result = execute(action, data)
+                this.messaging.notify('actionCompleted', { result })
+            } catch (e) {
+                console.log('unhandled exception: ', e)
+                this.messaging.notify('actionError', { error: e.toString() })
+            }
+        })
+    }
+}
--- a/src/features/broker-protection/click.js
+++ b/src/features/broker-protection/click.js
@@ -1,0 +1,19 @@
+import { getElement } from './utils.js'
+import { ErrorResponse, SuccessResponse } from './types.js'
+
+/**
+ * @param action // TODO: get type based on actionType
+ * @return {import('./types.js').ActionResponse}
+ */
+export function click (action) {
+    // there can be multiple elements provided by the action
+    for (const element of action.elements) {
+        const elem = getElement(document, element.selector)
+        if (!elem) {
+            return new ErrorResponse({ actionID: action.id, message: `could not find element to click with selector '${element.selector}'!` })
+        }
+        elem.click()
+    }
+
+    return new SuccessResponse({ actionID: action.id, actionType: action.actionType, response: null })
+}
--- a/src/features/broker-protection/execute.js
+++ b/src/features/broker-protection/execute.js
@@ -1,0 +1,31 @@
+import { click } from './click.js'
+import { ErrorResponse } from './types.js'
+
+/**
+ * @param {object} action
+ * @param {string} action.id
+ * @param {"extract" | "fillForm" | "click" | "expectation" | "getCaptchaInfo" | "solveCaptcha" | "navigate"} action.actionType
+ * @param {any} data
+ * @return {import('./types.js').ActionResponse}
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function execute (action, data) {
+    try {
+        switch (action.actionType) {
+        case 'click':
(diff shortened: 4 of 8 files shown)


shakyShane:
* main
  * **PR #843** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/843?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈
    * **PR #844** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/844?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #845** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/845?utm_source=stack-comment-icon" target="_blank">Graphite</a> 

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/843?utm_source=stack-comment).

src/features/broker-protection.js
```diff
@@ -0,0 +1,21 @@
...
+    init () {
+        this.messaging.subscribe('onActionReceived', (/** @type {any} */params) => {
+            try {
+                const action = params.state.action
+                const data = params.state.data
+                if (!action) {
+                    return this.messaging.notify('actionError', { error: 'No action found.' })
+                }
+                const result = execute(action, data)
+                this.messaging.notify('actionCompleted', { result })
```
muodov, line 14: does the return value matter in the `subscribe()` callback? The line 11 returns something, but this does not

src/features.js
```diff
@@ -31,7 +32,8 @@ export const platformSupport = {
         ...baseFeatures
     ],
     'apple-isolated': [
-        'duckPlayer'
+        'duckPlayer',
+        'brokerProtection'
```
muodov, line 36: does this affect iOS too?

src/features/broker-protection/click.js
```diff
@@ -0,0 +1,19 @@
+import { getElement } from './utils.js'
+import { ErrorResponse, SuccessResponse } from './types.js'
+
+/**
+ * @param action // TODO: get type based on actionType
+ * @return {import('./types.js').ActionResponse}
+ */
+export function click (action) {
```
muodov, line 8: heh, this is so similar to what we have in [autoconsent](https://github.com/duckduckgo/autoconsent/blob/main/lib/rule-executors.ts#L142). Just a note

src/features/broker-protection/utils.js
```diff
@@ -0,0 +1,150 @@
...
+        if (element && 'querySelectorAll' in element) {
+            return Array.from(element?.querySelectorAll?.(selector))
+        }
+        return null
+    } catch (e) {
+        return null
+    }
+}
+/**
+ * @param {HTMLElement|Element|Node|Document} element
```
muodov, line 93: wouldn't just `Node` be enough? Also for other functions here

src/features/broker-protection/execute.js
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @param {object} action
+ * @param {string} action.id
+ * @param {"extract" | "fillForm" | "click" | "expectation" | "getCaptchaInfo" | "solveCaptcha" | "navigate"} action.actionType
+ * @param {any} data
+ * @return {import('./types.js').ActionResponse}
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function execute (action, data) {
```
muodov, line 12: should this return a Promise?

muodov (commented):
looks good to me overall

src/features/broker-protection.js
shakyShane: No, just an early exit.

src/features/broker-protection.js
```diff
@@ -0,0 +1,21 @@
+import ContentFeature from '../content-feature.js'
+import { execute } from './broker-protection/execute.js'
+
+export default class BrokerProtection extends ContentFeature {
+    init () {
+        this.messaging.subscribe('onActionReceived', (/** @type {any} */params) => {
```
muodov, line 6: We shouldn't be worried about a possible race condition here, right?
Say, if Native decides to requests an action, but the page navigates, and the action is delivered to the new page that Native did not intend to ask.
I'm not familiar with the DBP use-case, but this was a big issue for autoconsent last year

src/features.js
shakyShane: no, currently iOS is not loading apple-isolated, but even if it were, we obviously wouldn't enable there (via config)

src/features/broker-protection/execute.js
shakyShane: no - originally all actions were async, but it created a lot of noise and in the end I realised none were currently doing anything async.

So, I've made them all sync as a default, and we can always add `async` ones when needed :)

src/features/broker-protection.js
shakyShane: It's a great question, something I raised at the beginning of the project. 

The way the scheduler is implemented on the native side means we were able to shrink it down to just this implementation.

src/features.js
muodov: Will the feature code be bundled when iOS starts loading apple-isolated?

src/features/broker-protection/utils.js
shakyShane: I'll try - I think it might be related to splitting the PRs up, I'm sure I widened this to fix something.

shakyShane:
@muodov  I narrowed some of the types as you suggested - but I don't think there's anything else blocking here?

src/features/broker-protection/utils.js
shakyShane: Done!

muodov:
@shakyShane no blockers, it looks good. I understand that these scripts will only be run on select pages, right? If so, we can be less worried about webcompat issues in the wild.

src/features/broker-protection/utils.js
```diff
@@ -0,0 +1,150 @@
+/**
+ * Get a single element.
+ *
+ * @param {Node} doc
+ * @param {string} selector
+ * @return {HTMLElement | null}
+ */
+export function getElement (doc = document, selector) {
+    if (isXpath(selector)) {
+        return safeQuerySelectorXPath(doc, selector)
```
jonathanKingston, line 10: What does the 'safe' prefix represent?

muodov (approved):

brianhall (approved):

src/features/broker-protection/utils.js
shakyShane: just that each of the calls is wrapped in a try-catch with a `null` return. Mostly because of throwing methods like querySelector.

shakyShane: resolving the conversation for the sake of merging, but we can come back and re-name these if you think there'a a better alternative :)

PR #846, opened 2023-11-21
Add support for schema-based message definitions

shakyShane:
<details>
<summary><strike>Feature entry pointbefore</strike></summary>

Features can opt-in when ready. Each feature would contain 3 top-level keys

- `notifications`
- `requests`
- `subscriptions`

These are designed to match the messaging API directly, and will provide automatic documentation for
native teams as they integrate.

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "title": "WebCompatMessages",
  "description": "Requests, Notifications and Subscriptions from the Web Compat feature",
  "additionalProperties": false,
  "properties": {
    "notifications": {},
    "requests": {
      "oneOf": [
        {
          "$ref": "./web-compat/web-share.json"
        }
      ]
    },
    "subscriptions": {}
  },
  "required": [
    "requests",
    "notifications",
    "subscriptions"
  ]
}
```
</details>

<details>
<summary>Feature entry point <b>NEW</b></summary>

Place a folder inside `src/messages` with at least 1 notification, request or subscription

```
📁 src/messages
  - 📁 duck-player
     - setUserValues.notify.json
```

When you have multiple messages, will look more like

```
📁 src/messages
  - 📁 duck-player
     - setUserValues.notify.json
     - getUserValues.request.json
     - getUserValues.response.json
     - onUserValuesUpdated.subscribe.json
```

`src/messages`
</details>

<details>
<summary><strike>Individual Message</strike></summary>

The important parts of a message are:

- `method` this is what native teams will map to their handlers
- `params` (optional) this is the payload we'll send in the outgoing message
- `result` (optional) this is the payload the native side will return in a response to a request.

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "title": "WebShare",
  "additionalProperties": false,
  "description": "todo: add description for `webShare` message",
  "required": ["method", "params", "result"],
  "properties": {
    "method": {
      "const": "webShare"
    },
    "params": {
      "title": "WebShareParams",
      "type": "object",
      "additionalProperties": false,
      "properties": {
        "title": {
          "description": "todo: add description for 'title' field",
          "type": "string"
        },
        "url": {
          "description": "todo: add description for 'url' field",
          "type": "string"
        },
        "text": {
          "description": "todo: add description for 'text' field",
          "type": "string"
        }
      }
    },
    "result": {
      "description": "todo: add return type here"
    }
  }
}
```
</details>

<details>
<summary>Individual Message <b>NEW</b></summary>

### Notifications

Notifications have 2 elements - `method` and `params`.

- `method`: this is represented in the filename, everything before the first `.` will be used
- `params`: in your json schema file, add the object type and properties as seen below

**notification without params**

```json
{ "$schema": "http://json-schema.org/draft-07/schema#" }
```
Given the filename `debug.notity.json`, this will generate

```typescript
export interface DebugNotification {
  method: "debug";
}
```

**notification _with_ params**

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "additionalProperties": false,
  "required": ["a"],
  "properties": {
    "a": {
      "type": "string"
    }
  }
}
```

Given the filename `debug.notity.json`, this will generate

```typescript
export interface SetValueNotification {
  method: "setValue";
  params: SetValueParams;
}
export interface SetValueParams {
  a: string;
}
```

### requests & responses

Requests work the same, except they use the file format `debug.request.json`.

- method: as above, it's inferred from the filename
- params: as above

Requests differ from Notifications in that they have an optional return type.
To create a return type, create a new file `debug.response.json`.

As above, the response can contain an object with properties.

For example, it might be common to not have any params, but to expect a return types (for example, when fetching data). 

To create that situation, create the following 2 file

- `fetchData.request.json`
- `fetchData.response.json`

```json
{ "$schema": "http://json-schema.org/draft-07/schema#" }
```

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "MyData",
  "type": "object",
  "properties": {
    "a": {
      "type": "string"
    }
  }
}
```

These 2 files would create the following: 

```ts
export interface FetchDataRequest {
  method: "setValue";
  result: MyData;
}
export interface MyData {
  a: string
}
```

### subscriptions

Subscriptions use the file format `onChange.subscribe.json`.

- subscriptionEvent: `onChange` from the filename
- params: as above, anything specified under the top level type will be used

</details>

<details>
<summary>Generated Types</summary>

From those JSON files, we get these types that can be imported and used in each feature

```ts
/**
 * Requests, Notifications and Subscriptions from the Web Compat feature
 */
export interface WebCompatMessages {
  requests: WebShare;
}
/**
 * todo: add description for `webShare` message
 */
export interface WebShare {
  method: "webShare";
  params: WebShareParams;
  /**
   * todo: add return type here
   */
  result: {
    [k: string]: unknown;
  };
}
export interface WebShareParams {
  /**
   * todo: add description for 'title' field
   */
  title?: string;
  /**
   * todo: add description for 'url' field
   */
  url?: string;
  /**
   * todo: add description for 'text' field
   */
  text?: string;
}
```
</details>

<details>
<summary>Javascript Usage</summary>

Once we have the types, we can proxy calls to `notify`, `request` and `subscribe` and provide
type-checking.

For example, from the types above we know that this feature can only make a `request` for `webShare`.

```js
export class WebCompat extends ContentFeature {
  init() {
    this.request('webShare', dataToSend)
      .then(result => console.log(result))
      .catch(console.error)
  }
}
```
</details>

image


shakyShane:
* main
  * **PR #846** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/846?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/846?utm_source=stack-comment).

src/messages/web-compat.json
```diff
@@ -0,0 +1,23 @@
...
+  "properties": {
+    "notifications": {},
+    "requests": {
+      "oneOf": [
+        {
+          "$ref": "./web-compat/web-share.json"
+        }
+      ]
+    },
+    "subscriptions": {}
```
shakyShane, line 16: for `notifications`, `requests` and `subscriptions`, we can use `oneOf` in JSON schema, which gives the following output when there is multiple items

```
interface Messages {
   requests: Msg1 | Msg2 | Msg3,
}
```

src/messages/web-compat/web-share.json
```diff
@@ -0,0 +1,35 @@
...
+          "description": "todo: add description for 'text' field",
+          "type": "string"
+        }
+      }
+    },
+    "result": {
+      "description": "todo: add return type here"
+    }
+  }
+}
```
shakyShane, line 35: This is the proposed format - the naming of `method`, `params` + `result` are based on the documented keys in https://duckduckgo.github.io/content-scope-scripts/modules/Messaging.html

src/types/web-compat.ts
```diff
@@ -0,0 +1,46 @@
...
+  title?: string;
+  /**
+   * todo: add description for 'url' field
+   */
+  url?: string;
+  /**
+   * todo: add description for 'text' field
+   */
+  text?: string;
+}
```
shakyShane, line 46: This is an example where in the JSDoc we gave an object a `title` -> That forces the the code generator to make a separate interface and allows you to import it into JS files

src/features/web-compat.js
```diff
@@ -64,6 +64,9 @@ function cleanShareData (data) {
     return dataToSend
 }
 
+/**
+ * @extends {ContentFeature<import('../types/web-compat.js').WebCompatMessages>}
```
jonathanKingston, line 68: I really want us to avoid this if possible.

src/features/web-compat.js
```diff
@@ -35,10 +35,10 @@ function canShare (data) {
 
 /**
  * Clean data before sending to the Android side
- * @returns {ShareRequestData}
+ * @returns {import('../types/web-compat.js').WebShareParams}
```
jonathanKingston, line 38: I wonder if we could alias this to the feature class itself?
Something like:`WebCompat.Messages.WebShareParams`

I know this isn't core to the design itself just would be nice to not require feature authors to alias into the type dir.

src/types/web-compat.ts
```diff
@@ -0,0 +1,46 @@
+// eslint-disable
+/**
+ * @module Webcompat Messages Schema
+ * @description
+ * These types are auto-generated from schema files.
```
jonathanKingston, line 5: Could these belong in the `build` dir to add clarity they're somewhat temporal based on other files?
I'm guessing that might be an issue due to files might need to be referenced prior to the final build.
My other suggestion would be perhaps to 'generated-types' or similar.

muodov: +1, would be nice to separate auto-generated files in the folder structure. `build` may be confusing (we probably want to commit these files so that code is traversable in Github), but a separate `generated-types` sgtm

src/features/web-compat.js
muodov: +1 this feels confusing because the you're supposed to think of what exactly would be generated.
I guess for new features we can auto-generate the whole boilerplate ContentFeature subclass, but I like Jonathan's suggestion more

src/messages/web-compat/web-share.json
shakyShane: Sorry, the link was incorrect, it's more like these: https://duckduckgo.github.io/content-scope-scripts/classes/Messaging_Schema.NotificationMessage.html

in general, we followed the JSON RPC spec for the naming scheme in the message format that actually get's sent/received so this is just a way of being consistent when a feature author creates individual messages.

For example, `method`, `params`, & `result` all come from https://duckduckgo.github.io/content-scope-scripts/classes/Messaging_Schema.RequestMessage.html

src/features/web-compat.js
shakyShane: Removed - none of this is needed in the new design!

src/types/web-compat.ts
shakyShane: right now it's inside `src/types`, because that's where other auto-generated types live. 

I don't imagine this is a blocker - we can talk about moving them all within another PR :)

src/types/web-compat.ts
```diff
@@ -0,0 +1,46 @@
...
+ * scripts/build-types.mjs is responsible for type generation.
```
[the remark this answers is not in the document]
shakyShane: I understand this sentiment, but I prefer to impose the JSON schema restriction for a few reasons:

- 1) It's a standard format that other platforms can consume if they so desire (and we should encourage after going this far)
- 2) The format over the wire is JSON, which is what JSON schema represents. Having boundary data defined in Typescript leaves us open to the ability to define types that are not JSON serializable - not to mention the temptation to use other advanced typescript features. By using JSON schema you're starting and ending with the exact feature set representable - with Typescript you'd be starting with additional features and would have to ensure you restrict each developer to only use a subset (via linting, PR review etc)
- 3) Consistency with other important projects like the privacy configuration, autofill (already using json schema there)
- 4) There are many tools out there (not to mention AI) that can very easily knock-up a JSON schema for you based on TS types if you need help authoring them - I almost always do that.

In short, JSON Schema is a PIA to author - but you tend to touch it very rarely once you have it in place.

shakyShane:
What's changed in the update:

- Removed the need to specify any types in the feature file 🎉
  - The original had a requirement where the feature author had to add the following JSDOC comment within each feature: `@extends {ContentFeature<import('../types/web-compat.js').WebCompatMessages>}`
  - The new version completely removes this - instead we import and augment the module inside a types files that the feature author never touches

shakyShane:
@jonathanKingston I need to add a couple of unit tests for the code-gen part, but this is ready to re-review now :)

shakyShane:
CC: @muodov

scripts/build-types.mjs
```diff
@@ -45,7 +56,30 @@ export async function buildTypes(mapping = defaultMapping) {
...
+            const template = `
+declare module ${JSON.stringify(relativePath)} {
+  export interface ${className} {
+    ${lines.join(',   ')}
+  }
+}
+
+`
+            content = content + template;
+        }
```
shakyShane, line 82: I'll be adding some tests for this code-gen block. It's not very complicated, but since it uses a convention-over-configuration approach (inferring things like class name) I'd like to have a few test cases in place

src/content-feature.js
```diff
@@ -235,6 +235,24 @@ export default class ContentFeature {
...
+
+    /**
+     * This is a wrapper to ensure all calls are typed via schemas.
+     * @type {(...args: any[]) => Promise<any>}
+     */
+    request (...args) {
+        const [name, params] = args
+        return this.messaging.request(name, params)
+    }
+
```
shakyShane, line 255: nobody will ever call these un-typed - they are here purely to proxy the params through to the messaging library.

This means features authors no longer do

```js
this.messaging.request("abc", {def: ""})
```

instead, they just call this directly, which will only allow messages which have been defined via the schema
```
this.request("abc", {def: ""})
```

src/features/web-compat.js
```diff
@@ -67,7 +67,7 @@ function cleanShareData (data) {
     return dataToSend
 }
 
-export default class WebCompat extends ContentFeature {
+export class WebCompat extends ContentFeature {
     /** @type {Promise<any> | null} */
```
shakyShane, line 71: because of the module augmentation, we need to expose this class separately as well as default

src/globals.d.ts
```diff
@@ -46,3 +46,24 @@ declare module 'ddg:runtimeInjects' {
...
+    notify<
+      Method extends T['notifications']['method'],
+      Msg = Extract<T['notifications'], { method: Method }>,
+    >(...args: Msg extends { params: infer Params } ? [Method, Params]: [Method]): void
+    request<
+      Method extends T['requests']['method'],
+      Msg = Extract<T['requests'], { method: Method }>,
+      Return = Msg extends { result: infer Result } ? Result : void
+    >(...args: Msg extends { params: infer Params } ? [Method, Params]: [Method]): Promise<Return>
+}
```
shakyShane, line 69: I wouldn't normally advocate for these complex types, but in here they exist ONLY to provide types in JS files - nobody other than the 'library' author will need to interact with these.

src/types/web-compat.ts
```diff
@@ -0,0 +1,51 @@
...
+   */
+  text?: string;
+}
+
+declare module "../features/web-compat.js" {
+  export interface WebCompat {
+    request: MessagingBase<WebCompatMessages>['request']
+  }
+}
+
```
shakyShane, line 51: this is the 'magic' 🪄 - here we are appending the `request` method to `WebCompat` in a type-safe way. Then, because `request` in the base class is typed in a loose manner, this one 'wins' and the type safety is achieved in the feature-file.

shakyShane: (same will apply for notifications and subscriptions too)

src/globals.d.ts
shakyShane: Also, I need to add support for subscriptions too, I'll do that when I add the unit tests

src/features/web-compat.js
jonathanKingston: What happens if a feature author does this wrong?

src/types/web-compat.ts
jonathanKingston: I see, so if we had messages that aren't feature specific we might need to do some further work to support that?

shakyShane: I'm looking now before we merge if there's any tweaks to make it easier to support none-features

src/features/web-compat.js
shakyShane: I think we should follow with a unit-test for this (as you suggested) - just in the interest of this PR not getting any larger.

shakyShane:
@jonathanKingston I updated the description with the new simpler model.

shakyShane:
@jonathanKingston  can you take another look through this 🙏🏻

src/content-feature.js
```diff
@@ -235,6 +235,33 @@ export default class ContentFeature {
...
+        this.messaging.notify(name, params)
+    }
+
+    /**
+     * This is a wrapper to ensure all calls are typed via schemas.
+     * @type {(...args: any[]) => Promise<any>}
+     */
+    request (...args) {
+        const [name, params] = args
+        return this.messaging.request(name, params)
```
jonathanKingston, line 253: Can we fast follow to make `this.messaging `this internal either by making it private # or at least have prefix notation.

unit-test/scripts/build-types.js
```diff
@@ -0,0 +1,126 @@
...
+
+describe('createMessagingTypes', () => {
+    /**
+     * a helper to avoid newlines/spaces breaking otherwise good tests
+     * @param {string} a
+     * @param {string} b
+     */
+    function compare (a, b) {
+        const lines = a.trim().split('\n').map(x => x.trim()).join('\n')
+        const lines3 = b.trim().split('\n').map(x => x.trim()).join('\n')
```
jonathanKingston, line 11: lines3 lol?

src/content-feature.js
```diff
@@ -235,6 +235,33 @@ export default class ContentFeature {
...
+     * This is a wrapper to ensure all calls are typed via schemas.
+     * @type {(...args: any[]) => any}
+     */
+    notify (...args) {
+        const [name, params] = args
+        this.messaging.notify(name, params)
+    }
+
+    /**
+     * This is a wrapper to ensure all calls are typed via schemas.
```
jonathanKingston, line 248: I think explain that these are overloaded by specific types in the types/ dir. (we're gonna forget how this wires into the code in a week or so)

scripts/build-types.mjs
```diff
@@ -59,3 +122,5 @@ if (isLaunchFile(import.meta.url)) {
             process.exit(1)
         })
 }
+
+console.log('shane');
```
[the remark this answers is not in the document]
shakyShane: 😱 🤣🤣

scripts/build-types.mjs
```diff
@@ -42,48 +43,75 @@ const defaultMapping = {
 export async function buildTypes(mapping = defaultMapping) {
     for (let [featureName, manifest] of Object.entries(mapping)) {
         if (manifest.exclude) continue;
-        const typescript = await compileFromFile(manifest.schema, {
-            bannerComment: `
-            // eslint-disable
+        if (manifest.kind === 'messages' || manifest.kind === 'settings') {
```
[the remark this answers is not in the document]
shakyShane: Ahh good catch, I need to change this back to just be called 'messages'

src/content-feature.js
shakyShane: I think we should mark this.messaging as @deprecated in this PR. That way nothing changes but people start to see info in their editors about it.

jonathanKingston (approved):
Thanks for getting this over the line! Really great work!

PR #856, opened 2023-11-29
Fix WebView viewport quirks and desktop mode

muodov:
https://app.asana.com/0/72649045549333/1206040541729852/f

### DIFF
--- a/src/features/web-compat.js
+++ b/src/features/web-compat.js
@@ -104,7 +104,7 @@ export default class WebCompat extends ContentFeature {
             this.shimWebShare()
         }
 
-        if (this.getFeatureSettingEnabled('viewportWidth')) {
+        if (this.getFeatureSettingEnabled('viewportWidth') || true) {
             this.viewportWidthFix()
         }
     }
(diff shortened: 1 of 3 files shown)


src/features/web-compat.js
```diff
@@ -104,7 +104,7 @@ export default class WebCompat extends ContentFeature {
             this.shimWebShare()
         }
 
-        if (this.getFeatureSettingEnabled('viewportWidth')) {
+        if (this.getFeatureSettingEnabled('viewportWidth') || true) {
```
muodov, line 107: Will remove this this before merging

src/features/web-compat.js
```diff
@@ -512,15 +512,41 @@ export default class WebCompat extends ContentFeature {
...
+        const viewportContentParts = viewportContent?.split(/,|;/)
+        if (!viewportTag || this.desktopModeEnabled) {
+            // force wide viewport width
+            if (!viewportTag) {
+                viewportTag = document.createElement('meta')
+                viewportTag.setAttribute('name', 'viewport')
+            }
+            const forcedWidth = screen.width >= 1280 ? 1280 : 980
+            // Race condition: depending on the loading state of the page, initial scale may or may not be respected, so the page may look zoomed-in after applying this hack.
+            // Usually this is just an annoyance, but it may be a bigger issue if user-scalable=no is set, so we remove it too.
```
muodov, line 528: To reproduce this: go to reddit.com, enable desktop mode, and sometimes the page would be zoomed in.
I tried to delay based on `document.loadingState`, but that didn't work unfortunately. Suggestions welcome.

src/features/web-compat.js
```diff
@@ -512,15 +512,42 @@ export default class WebCompat extends ContentFeature {
...
+            const forcedInitialScale = (screen.width / forcedWidth).toFixed(3)
+            const newContentParts = [`width=${forcedWidth},initial-scale=${forcedInitialScale}`]
+            viewportContentParts?.forEach((part) => {
+                if (!part.includes('width') && !part.includes('initial-scale') && !part.includes('user-scalable')) {
```
jonathanKingston, line 532: Should we care about the value here do you think?
What about preserving ordering of the width, initial scale too?

```suggestion
                const [key, value] = part.split('=').map(a => a.trim())
                if (!['width', 'initial-scale', 'user-scalable'].includes(key)) {
```

src/features/web-compat.js
jonathanKingston: Without digging into the specific case I don't know if I can think of anything.

For Android I don't think we can run any earlier even if we moved the code into `load` phase rather than `init`.

src/features/web-compat.js
```diff
@@ -512,15 +512,42 @@ export default class WebCompat extends ContentFeature {
...
+                    newContentParts.push(part)
+                }
+            })
+            viewportTag.setAttribute('content', newContentParts.join(','))
+            document.head.appendChild(viewportTag)
```
jonathanKingston, line 537: I think we should manipulate the last tag as we do for the other code arm. Adding tags has a chance of breakage for sites not expecting additional nodes / ordering.

muodov: This should only add a new tag if it was not there before. Existing tag would be moved

jonathanKingston: Can we prevent it moving in the existing case?

src/features/web-compat.js
muodov: The order should not matter since we remove all original occurrences, so it's guaranteed that there will only be one occurrence in the final value.
Searching within the key only makes sense. And it should be case-insensitive btw, I'll make the change

src/features/web-compat.js
```diff
@@ -512,15 +512,46 @@ export default class WebCompat extends ContentFeature {
...
+                    newContentParts.push(part)
+                }
+            })
+            viewportTag.setAttribute('content', newContentParts.join(','))
+            if (!viewportTagExists) {
+                document.head.appendChild(viewportTag)
+            }
+        } else { // mobile mode with a viewport tag
+            // fix an edge case where WebView forces the wide viewport
+            const widthPart = viewportContentParts?.find((part) => part.includes('width'))
```
jonathanKingston, line 544: Should we lower and trim these too? Sorry for follow up here.

muodov: thanks for catching this, it slipped my attention!

src/features/web-compat.js
```diff
@@ -528,27 +532,25 @@ export default class WebCompat extends ContentFeature {
...
             // Usually this is just an annoyance, but it may be a bigger issue if user-scalable=no is set, so we remove it too.
             const forcedInitialScale = (screen.width / forcedWidth).toFixed(3)
-            const newContentParts = [`width=${forcedWidth},initial-scale=${forcedInitialScale}`]
-            viewportContentParts?.forEach((part) => {
-                const key = part.split('=')[0].trim().toLowerCase()
+            let newContent = `width=${forcedWidth}, initial-scale=${forcedInitialScale}`
+            parsedViewportContent.forEach(([key], idx) => {
                 if (!['width', 'initial-scale', 'user-scalable'].includes(key)) {
-                    newContentParts.push(part)
+                    newContent = newContent.concat(`,${viewportContentParts[idx]}`) // reuse the original values, not the parsed ones
```
jonathanKingston, line 538: Took me a minute to understand but 🔥 thanks!

jonathanKingston (approved):
Great work thank you!

PR #870, opened 2023-12-14
Fix viewport edge cases

muodov:
Fix https://github.com/duckduckgo/privacy-configuration/issues/1633

### DIFF
--- a/src/features/web-compat.js
+++ b/src/features/web-compat.js
@@ -511,23 +511,42 @@ export default class WebCompat extends ContentFeature {
         }
     }
 
+    /**
+     * create or update a viewport tag with the given content
+     * @param {HTMLMetaElement|null} viewportTag
+     * @param {string} forcedValue
+     */
+    forceViewportTag (viewportTag, forcedValue) {
+        const viewportTagExists = Boolean(viewportTag)
+        if (!viewportTag) {
+            viewportTag = document.createElement('meta')
+            viewportTag.setAttribute('name', 'viewport')
+        }
+        viewportTag.setAttribute('content', forcedValue)
+        if (!viewportTagExists) {
+            document.head.appendChild(viewportTag)
+        }
+    }
+
     viewportWidthFixInner () {
+        /** @type {NodeListOf<HTMLMetaElement>} **/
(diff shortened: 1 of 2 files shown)


src/features/web-compat.js
```diff
@@ -503,6 +503,11 @@ export default class WebCompat extends ContentFeature {
     }
 
     viewportWidthFix () {
+        if (document.readyState === 'loading') {
+            // if the document is not ready, we may miss the original viewport tag
+            document.addEventListener('DOMContentLoaded', () => this.viewportWidthFix())
```
jonathanKingston, line 508: I'd rename the rest of the method `viewportWidthFixInner` or similar and have the remainder of this function and this listener call that.

So this function would become:
```javascript
    viewportWidthFix() {
       if (document.readyState === 'loading') {
            // if the document is not ready, we may miss the original viewport tag
            document.addEventListener('DOMContentLoaded', () => this.viewportWidthFixInner())
            return
       }
       this.viewportWidthFixInner()
    }
```

This reduces the risk of edge case re-entrancy problems in future.

Alternatively you could solve it at the current `viewportWidthFix` call site but I think that isn't quite as clean.

src/features/web-compat.js
```diff
@@ -511,23 +511,42 @@ export default class WebCompat extends ContentFeature {
...
             const [key, value] = part.split('=').map(p => p.trim().toLowerCase())
             return [key, value]
         })
-        if (!viewportTag || this.desktopModeEnabled) {
+
+        const { forcedDesktopValue, forcedMobileValue } = this.getFeatureSetting('viewportWidth')
+        if (typeof forcedDesktopValue === 'string' && this.desktopModeEnabled) {
+            this.forceViewportTag(viewportTag, forcedDesktopValue)
+        } else if (typeof forcedMobileValue === 'string' && !this.desktopModeEnabled) {
+            this.forceViewportTag(viewportTag, forcedMobileValue)
```
jonathanKingston, line 547: Shouldn't these return after?

jonathanKingston (approved):
Cool I think the helper method makes the code a little simpler to follow now too :)

PR #875, opened 2023-12-19
Add support for screen.lock/unlock

nshuba:
Task/Issue URL: https://app.asana.com/0/0/1206163779688838/f

### Description
Support [screen.lock](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/lock) and [screen.unlock](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/unlock) APIs

### Steps to test this PR
See https://github.com/duckduckgo/Android/pull/4027

### DIFF
--- a/src/features/web-compat.js
+++ b/src/features/web-compat.js
@@ -14,6 +14,7 @@ function windowSizingFix () {
 
 const MSG_WEB_SHARE = 'webShare'
 const MSG_PERMISSIONS_QUERY = 'permissionsQuery'
+const MSG_SCREEN_LOCK = 'screenLock'
 
 function canShare (data) {
     if (typeof data !== 'object') return false
@@ -108,6 +109,7 @@ export default class WebCompat extends ContentFeature {
         if (this.getFeatureSettingEnabled('viewportWidth')) {
             this.viewportWidthFix()
         }
+        this.screenLockFix()
     }
 
     /** Shim Web Share API in Android WebView */
@@ -230,6 +232,23 @@ export default class WebCompat extends ContentFeature {
                     }
                 }
                 return Reflect.apply(target, thisArg, args)
+            }
+        })
+    }
+
+    screenLockFix() {
+        window.ScreenOrientation.prototype.lock = new Proxy(async (requestedOrientation) => {
+            this.addDebugFlag()
+            // TODO handle error cases
+
+            const response = await this.messaging.request(MSG_SCREEN_LOCK, { orientation : requestedOrientation })
+            console.log("xxx Got Response xxx")
+            console.log(response)
+
+            return Promise.reject(new Error("DOMException xxx: The page needs to be fullscreen in order to call screen.orientation.lock()."))
+        }, {
+            get (target, name) {
+                return Reflect.get(target, name)
             }
         })
     }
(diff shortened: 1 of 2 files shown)


src/features/web-compat.js
```diff
@@ -234,6 +240,46 @@ export default class WebCompat extends ContentFeature {
         })
     }
 
+    screenLockFix () {
+        // @ts-expect-error Property 'lock' does not exist on type 'ScreenOrientation'
+        window.ScreenOrientation.prototype.lock = new Proxy(async (requestedOrientation) => {
```
nshuba, line 245: Asked this in Asana, but asking here as well: is this the correct way to override? Also, how should we check if we need to override since lock/unlock are actually defined?

jonathanKingston: Answered in Asana for the check part: https://app.asana.com/0/0/1206174768741464/1206228892953010/f

As this is a propertyDescriptor I think you want to use `this.wrapProperty` and match the properties returned by:

```
Object.getOwnPropertyDescriptors(window.ScreenOrientation.prototype).lock
> Object { value: lock(), writable: true, enumerable: true, configurable: true }
```

I also scoped a project to make this simpler to understand: https://app.asana.com/0/1163321984198618/1206229471617467/f

nshuba: Got it, thanks a lot for updating the docs!

src/features/web-compat.js
```diff
@@ -283,6 +292,68 @@ export default class WebCompat extends ContentFeature {
...
+            'portrait-primary',
+            'portrait-secondary',
+            'landscape-primary',
+            'landscape-secondary',
+            'unsupported'
+        ]
+
+        this.wrapProperty(globalThis.ScreenOrientation.prototype, 'lock', {
+            value: async (requestedOrientation) => {
+                this.addDebugFlag()
```
jonathanKingston, line 313: One to validate, but I think we should be taking care of this now: https://github.com/duckduckgo/content-scope-scripts/blob/012b3925944a3cd5338aef51a42736448d254937/src/content-feature.js#L287

nshuba: Tested - looks like it's indeed not needed. Fixed, thanks!

src/features/web-compat.js
```diff
@@ -283,6 +292,66 @@ export default class WebCompat extends ContentFeature {
...
+                    return Promise.reject(new TypeError("Failed to execute 'lock' on 'ScreenOrientation': 1 argument required, but only 0 present."))
+                }
+                if (!validOrientations.includes(requestedOrientation)) {
+                    return Promise.reject(new TypeError(`Failed to execute 'lock' on 'ScreenOrientation': The provided value '${requestedOrientation}' is not a valid enum value of type OrientationLockType.`))
+                }
+                if (this.#activeScreenLockRequest) {
+                    return Promise.reject(new DOMException('Screen lock already in progress', 'AbortError'))
+                }
+
+                this.#activeScreenLockRequest = await this.messaging.request(MSG_SCREEN_LOCK, { orientation: requestedOrientation })
```
jonathanKingston, line 323:
```suggestion
                this.#activeScreenLockRequest = this.messaging.request(MSG_SCREEN_LOCK, { orientation: requestedOrientation })
```

nshuba: oops, nice catch!

jonathanKingston (approved):
Looks good to me, I've not tested this but I suspect we will as part of the release.

PR #877, opened 2023-12-19
Support forced zoom in webcompat feature

muodov:
Moves the forced zoom implementation on Adnroid into C-S-S
See https://app.asana.com/0/0/1206193165431943/f

---
  - https://app.asana.com/0/0/1206205692287339

### DIFF
--- a/src/features/web-compat.js
+++ b/src/features/web-compat.js
@@ -603,7 +603,7 @@ export default class WebCompat extends ContentFeature {
         // Chrome respects only the last viewport tag
         const viewportTag = viewportTags.length === 0 ? null : viewportTags[viewportTags.length - 1]
         const viewportContent = viewportTag?.getAttribute('content') || ''
-        const viewportContentParts = viewportContent ? viewportContent.split(/\s*,\s*|\s*;\s*/) : []
+        const viewportContentParts = viewportContent ? viewportContent.split(/,|;/) : []
         const parsedViewportContent = viewportContentParts.map((part) => {
             const [key, value] = part.split('=').map(p => p.trim().toLowerCase())
             return [key, value]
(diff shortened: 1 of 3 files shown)


src/features/web-compat.js
```diff
@@ -534,30 +534,38 @@ export default class WebCompat extends ContentFeature {
...
+        }
+
+        // otherwise, check for special cases
+        const forcedValues = {}
+
+        if (this.forcedZoomEnabled) {
+            forcedValues['initial-scale'] = 1
+            forcedValues['user-scalable'] = 'yes'
+            forcedValues['maximum-scale'] = 10
+        }
```
muodov, line 560: Theoretically, it should be enough just to remove `user-scalable=no` if it exists, but I want to mimic the existing implementation which explicitly sets these values

src/features/web-compat.js
```diff
@@ -603,30 +603,38 @@ export default class WebCompat extends ContentFeature {
         // Chrome respects only the last viewport tag
         const viewportTag = viewportTags.length === 0 ? null : viewportTags[viewportTags.length - 1]
         const viewportContent = viewportTag?.getAttribute('content') || ''
-        const viewportContentParts = viewportContent ? viewportContent.split(/,|;/) : []
+        const viewportContentParts = viewportContent ? viewportContent.split(/\s*,\s*|\s*;\s*/) : []
```
jonathanKingston, line 606: Does the trim not take care of this?

muodov: It does, but the trimmed results end up in `parsedViewportContent`, and we reuse `viewportContentParts` to construct the final value

src/features/web-compat.js
```diff
@@ -635,10 +643,25 @@ export default class WebCompat extends ContentFeature {
...
+
+        const newContent = []
+        'width' in forcedValues && newContent.push(`width=${forcedValues.width}`)
+        'initial-scale' in forcedValues && newContent.push(`initial-scale=${forcedValues['initial-scale']}`)
+        'user-scalable' in forcedValues && newContent.push(`user-scalable=${forcedValues['user-scalable']}`)
+        'maximum-scale' in forcedValues && newContent.push(`maximum-scale=${forcedValues['maximum-scale']}`)
+        if (newContent.length > 0) { // need to override at least one viewport component
+            parsedViewportContent.forEach(([key], idx) => {
+                // @ts-expect-error hasOwn does exist
+                if (!(Object.hasOwn(forcedValues, key))) {
```
jonathanKingston, line 659: We 'own' forcedValues why not just use `in`?

src/features/web-compat.js
jonathanKingston: I think I understand, to keep the whitespace around for later. Can you add this as a comment please :)

src/features/web-compat.js
```diff
@@ -635,10 +643,25 @@ export default class WebCompat extends ContentFeature {
...
+                    newContent.push(`${viewportContentParts[idx]}`) // reuse the original values, not the parsed ones
```
jonathanKingston, line 660:
```suggestion
                    newContent.push(viewportContentParts[idx]) // reuse the original values, not the parsed ones
```

Maybe run `trim()` here if you wanted to remove the whitespace rather than making the regex more complex?

src/features/web-compat.js
jonathanKingston: I think trimming there would be clearer and simpler personally.

src/features/web-compat.js
jonathanKingston, line 660: There is a risk that `idx` gets outdated if `viewportContentParts` or `parsedViewportContent` changes item sorting or counts, maybe a comment at their declaration might be helpful?

muodov: I've marked those variables as readonly, that should address this concern?

jonathanKingston (approved):
Cheers for this!

PR #879, opened 2023-12-20
Let extractProfiles support a top-level xpath selector

brianhall:
I was getting an error with a json file that was using an xpath selector in the "profile" section, since we already have a good util to query with either css or xpath selectors, I've swapped `querySelectorAll` out for `getElements`.


brianhall:
@shakyShane Could I get your 👀 on this? I think it should be a pretty uncontroversial change. I'm still doing some testing, but seems to work as expected.

shakyShane (approved):

PR #889, opened 2024-01-05
set screenY and screenTop to 0 when equal to screen height

ladamski:
Background: https://app.asana.com/0/276630244458377/1206165160907347/f
See also https://developer.mozilla.org/en-US/docs/Web/API/Window/screenY

For Mac where we are incorrectly providing the screen height instead of `A number equal to the number of CSS pixels from the top edge of the browser viewport to the top edge of the screen.`, which should be the actual value or 0.

### DIFF
--- a/src/features/web-compat.js
+++ b/src/features/web-compat.js
@@ -5,6 +5,11 @@ import { URL } from '../captured-globals.js'
  * Fixes incorrect sizing value for outerHeight and outerWidth
  */
 function windowSizingFix () {
+    // macOS browser incorrectly reports window height for window.screenY / window.screenTop
+    if (window.screenY === window.screen.height) {
+        window.screenY = window.screenTop = 0
+    }
+
     if (window.outerHeight !== 0 && window.outerWidth !== 0) {
         return
     }
--- a/src/features/web-compat.js
+++ b/src/features/web-compat.js
@@ -4,10 +4,17 @@ import { URL } from '../captured-globals.js'
 /**
  * Fixes incorrect sizing value for outerHeight and outerWidth
  */
-function windowSizingFix () {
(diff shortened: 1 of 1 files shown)


src/features/web-compat.js
```diff
@@ -5,6 +5,11 @@ import { URL } from '../captured-globals.js'
  * Fixes incorrect sizing value for outerHeight and outerWidth
  */
 function windowSizingFix () {
+    // macOS browser incorrectly reports window height for window.screenY / window.screenTop
+    if (window.screenY === window.screen.height) {
+        window.screenY = window.screenTop = 0
```
jonathanKingston, line 10: `0` isn't a valid value here, it should be the position of the content within the screen. So for me Safari is at it's minimum `38` due to the browser chrome being in the way.

I would suggest we either make this value remotely configurable to adjust the geometry of all of these values.

`outerHeight` is `0` despite it being a size that should always be bigger than innerHeight. This value should actually be the same offset when the browser is fullscreen as `screenTop`/`Y`

ladamski: As mentioned in the Asana task, our extension sets `screenY/screenTop` to `0`, just verified it again in Chrome and Firefox.  

`0` is actually a valid value here, when you are in full-screen mode.  If we want to make it configurable we can but it can vary per system anyway (`screenY` is `24` for me in Chrome ATM), so `0` seemed like a reasonable default.

ladamski: Re `outerHeight` are you saying we should change this existing behavior?  I can update that if so.

ladamski: Updated to use the `windowSizing.screenYOverride` setting if defined, or `screen.availTop` if not.

shakyShane:
@ladamski hey! if this is still required, please re-open as needed :)

PR #909, opened 2024-01-31
HTML Onboarding

shakyShane:
Added a new Onboarding application (to be shared across all browsers eventually)

- Windows PR that uses this https://github.com/duckduckgo/windows-browser/pull/1228
- Asana task (ship review) https://app.asana.com/0/0/1206490374198520/f

**Added:** 
- [x] the full application built with preact + regular CSS
- [x] messaging documentation
- [x] integration tests

**Changed**
- [x] small tweaks to the build script, to allow more external files to be referenced

**How to Test**

The application has been extensively tested in-place during development and during the ship review, but if you want to run it:

- Within the Windows App: 
  - https://github.com/duckduckgo/windows-browser/pull/1228
- Manual:
  - `npm run build` in the root of the project. 
  - Then open the `build/integration/pages/onboarding` directory with a local web server
- Automated
  - Run the playwright tests:
    - `cd packages/special-pages`
    - `npm test`


greyivy (dismissed):
Code LGTM! I wasn't super thorough, but nothing big stuck out and it functions well in my testing.

Flagging the failing test.

packages/special-pages/index.mjs
```diff
@@ -113,15 +118,26 @@ for (const copyJob of copyJobs) {
...
+            // external: ['../assets/img/*'],
+            sourcemap: NODE_ENV === 'development',
+            // publicPath: '/js',
+            loader: {
+                '.js': 'jsx',
+                '.module.css': 'local-css',
+                '.svg': 'file',
+                '.jpg': 'file',
+                '.png': 'file',
+                '.riv': 'file'
```
shakyShane, line 139: `jsx` -> for the preact components
`local-css` -> for the scoped css
`file` -> just uses the files as-is without processing

packages/special-pages/index.mjs
```diff
@@ -67,20 +72,20 @@ for (const [pageName, injectNames] of Object.entries(support)) {
...
-                const jsSrc = join(pageSrc, 'js', 'index.js')
-                if (!existsSync(jsSrc)) {
-                    errors.push(`${jsSrc} does not exist`)
-                    continue
-                }
-                const jsDest = join(pageOutputDirectory, 'js', 'index.js')
+                const entryPoints = [
+                    join(pageSrc, 'js', 'index.js'),
+                    join(pageSrc, 'js', 'inline.js')
+                ].filter(pathname => existsSync(pathname));
```
shakyShane, line 83: this is allowing an additional entry point, called `inline.js` which can be used as a blocking request in HTML documents, useful for setting up environment stuff (since it has access to `import.meta` vars). The idea is that this would just be a few lines for now, but could be used to tree-shake more stuff in the future like messaging implementations etc.

shakyShane:
@jonathanKingston merging for now to unblock the experiment - if we want to tweak anything in terms of the build step changes we can follow up :)

PR #911, opened 2024-02-06
update ClickToLoad to use shared C-S-S (Mac)

ladamski:
Supporting work for https://app.asana.com/0/72649045549333/1205105078450227/f

Specifically, it includes fixes to:
- handle missing `window.crypto.subtle` in `captureGlobals` for HTTP pages (https://app.asana.com/0/0/1206422390411022/1206518726679048/f)
- enabled Mac webkit messaging for CTL

### DIFF
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -2025,7 +2025,7 @@ export default class ClickToLoad extends ContentFeature {
             const config = new WebkitMessagingConfig({
                 secret: '',
                 hasModernWebkitAPI: true,
-                webkitMessageHandlerNames: ['contentScopeScripts']
+                webkitMessageHandlerNames: ['contentScopeScriptsIsolated'] // contentScopeScriptsIsolated
             })
             this._messaging = new Messaging(this.messagingContext, config)
             return this._messaging
--- a/src/features/click-to-load.js
+++ b/src/features/click-to-load.js
@@ -2025,7 +2025,7 @@ export default class ClickToLoad extends ContentFeature {
             const config = new WebkitMessagingConfig({
                 secret: '',
                 hasModernWebkitAPI: true,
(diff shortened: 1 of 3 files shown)


src/features/click-to-load.js
```diff
@@ -2016,16 +2016,16 @@ export default class ClickToLoad extends ContentFeature {
...
             const config = new TestTransportConfig(this._clickToLoadMessagingTransport)
             this._messaging = new Messaging(this.messagingContext, config)
             return this._messaging
-        } else if (this.platform.name === 'ios') {
+        } else if (this.platform.name === 'ios' || this.platform.name === 'macos') {
             const config = new WebkitMessagingConfig({
                 secret: '',
                 hasModernWebkitAPI: true,
-                webkitMessageHandlerNames: ['contentScopeScripts']
+                webkitMessageHandlerNames: ['contentScopeScriptsIsolated'] // contentScopeScriptsIsolated
```
jonathanKingston, line 2028: We're not going to be in the isolated context are we? IMO this should be changed back.

```suggestion
                webkitMessageHandlerNames: ['contentScopeScripts']
```

ladamski: I meant to remove that comment, but we do require `contentScopeScriptsIsolated`.  The issue is that the `subscribe` messaging pattern didn't work otherwise, @shakyShane  and I spent some time looking at that but this appeared to be the best path forward.  CTL works fine in isolated FWIW, and seems like the safer path anyway?

jonathanKingston: Yeah cool just didn't think it did work, just remove the comment instead and we're golden. Thanks for the explanation :)

ladamski: @jonathanKingston oh yep, I removed the comment already so resolving this thread.

jonathanKingston (approved):

jonathanKingston (approved):

PR #912, opened 2024-02-07
feat: add retry logic

shakyShane:
Asana: https://app.asana.com/0/0/1206543393124878/f

This allows actions to be defined with a 'retry' config.

The 'environment: web' part allows this to be interpreted by native at a later point (to consolidate logic that's being added on each native side)

```
retry: {
   environment: 'web',
   maxAttempts: 10,
   interval: { ms: 1000 }
}
```

### DIFF
--- a/integration-test/playwright/broker-protection.spec.js
+++ b/integration-test/playwright/broker-protection.spec.js
@@ -352,4 +352,60 @@ test.describe('Broker Protection communications', () => {
             dbp.isSuccessMessage(response)
         })
     })
+
+    test.describe('retrying', () => {
+        test('retrying a click', async ({ page }, workerInfo) => {
+            const dbp = BrokerProtectionPage.create(page, workerInfo)
+            await dbp.enabled()
+            await dbp.navigatesTo('retry.html')
+
+            await dbp.simulateSubscriptionMessage('onActionReceived', {
+                state: {
+                    action: {
+                        actionType: 'click',
+                        id: '5',
+                        retry: {
+                            environment: 'web',
+                            maxAttempts: 10,
+                            interval: { ms: 1000 }
+                        },
+                        elements: [
+                            {
+                                type: 'button',
+                                selector: 'button'
+                            }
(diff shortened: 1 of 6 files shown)


shakyShane:
* `main`
  * **#912** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/912?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

src/timer-utils.js
```diff
@@ -0,0 +1,36 @@
...
+    const exceptions = []
+    for (let i = 0; i < config.maxAttempts; i++) {
+        try {
+            lastResult = fn()
+        } catch (e) {
+            exceptions.push(e.toString())
+        }
+
+        // stop when there's a good result to return
+        if (lastResult && 'success' in lastResult) break
```
brianhall, line 27: Maybe a silly question, but where do we get 'success' from?

integration-test/playwright/broker-protection.spec.js
```diff
@@ -270,4 +270,60 @@ test.describe('Broker Protection communications', () => {
...
+            await dbp.navigatesTo('retry.html')
+
+            await dbp.simulateSubscriptionMessage('onActionReceived', {
+                state: {
+                    action: {
+                        actionType: 'click',
+                        id: '5',
+                        elements: [
+                            {
+                                type: 'button',
```
brianhall, line 317: I think the primary use-case here is buttons, which totally makes sense. Would this work for links as well? Standard links and links with only event handlers?

shakyShane: this doesn't currently limit the actual execution to just 'button's - it's currently only used to indicate 'anything that can be clicked'. So I think it'll just support <a> by default :)

src/timer-utils.js
shakyShane: Added comments :)

PR #913, opened 2024-02-07
support matching weights

shakyShane:
https://app.asana.com/0/0/1206543393124882/f

previous click actions looked like this: 

```json
{
  "actionType": "click",
  "elements": [
    {
      "type": "button",
      "selector": ".opt-out"
    }
  ]
}
```

when used, the `document` is used as the 'root element'. Meaning that the CSS selector `.opt-out` here is the equivalent of

```js
document.querySelector('.opt-out')
```

This change allows us to declare a 'parent' element to replace 'document' as the root element where the search should occur. 

The first supported 'parent' selector is a profile match, therefore the following definition is the same as saying

- first select the parent, based on profile matching logic
- use that as the root element for the the subsequent match

Untitled (8)

### DIFF
--- a/integration-test/playwright/page-objects/broker-protection.js
+++ b/integration-test/playwright/page-objects/broker-protection.js
@@ -103,7 +103,7 @@ export class BrokerProtectionPage {
     /**
      * Simulate the native-side pushing an action into the client-side JS
      *
-     * @param {'extract.json' | 'extract2.json' | 'extract3.json' | 'extract4.json' | 'extract5.json' | 'extract-irregular1.json' | 'extract-irregular2.json' | 'extract-irregular3.json' | 'results2.json' | 'navigate.json' | 'fill-form.json' | 'click.json' | 'expectation.json' | 'get-captcha.json' | 'solve-captcha.json' | 'action-not-found.json'} action - add more action types here
+     * @param {string} action
      * @return {Promise<void>}
      */
     async receivesAction (action) {
--- a/integration-test/playwright/page-objects/broker-protection.js
+++ b/integration-test/playwright/page-objects/broker-protection.js
@@ -60,20 +60,7 @@ export class BrokerProtectionPage {
      * @return {void}
      */
     isExtractMatch (response, person) {
-        if (person.name) { expect(response[0]?.name).toBe(person.name) }
-        if (person.alternativeNames) { expect(response[0]?.alternativeNames).toStrictEqual(person.alternativeNames) }
-        if (person.age) { expect(response[0]?.age).toBe(person.age) }
-        if (person.addresses) { expect(response[0]?.addresses).toStrictEqual(person.addresses) }
-        if (person.relatives) { expect(response[0]?.relatives).toStrictEqual(person.relatives) }
-        if (person.phoneNumbers) { expect(response[0]?.phoneNumbers).toStrictEqual(person.phoneNumbers) }
-        if (person.profileUrl) { expect(response[0]?.profileUrl).toContain(person.profileUrl) }
-    }
-
-    /**
-     * @return {void}
-     */
-    isMultiple (response) {
-        expect(response.length).toBeGreaterThan(1)
+        expect(person).toMatchObject(response)
     }
 
     /**
--- a/integration-test/test-pages/broker-protection/actions/click-parent.json
+++ b/integration-test/test-pages/broker-protection/actions/click-parent.json
@@ -1,0 +1,42 @@
+{
+  "state": {
+    "action": {
+      "actionType": "click",
+      "elements": [
+        {
+          "type": "button",
+          "selector": ".",
+          "parent": {
+            "profileMatch": {
+              "selector": ".record",
+              "profile": {
+                "name": {
+                  "selector": ".name"
+                },
+                "age": {
+                  "selector": ".age"
+                },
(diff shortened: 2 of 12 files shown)


shakyShane:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/913?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* `main`
  * **#912** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/912?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **#913** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/913?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

src/features/broker-protection/actions/extract.js
```diff
@@ -27,18 +27,44 @@ import { matchAddressFromAddressListCityState } from '../comparisons/address.js'
...
+            const scrapedData = createProfile(element, action.profile)
+            const { result, score } = scrapedDataMatchesUserData(userData, scrapedData)
+            return {
+                scrapedData,
+                result,
+                score,
+                element
+            }
+        })
+    }
```
shakyShane, line 67: this change is all about returning all elements, to allow different consumers to choose which element to use. 

For example, a scan might be really loose and just select multiple matches, but other places like a search situation you might want to consider the match's score

src/features/broker-protection/comparisons/address.js
```diff
@@ -23,7 +23,7 @@ export function matchAddressFromAddressListCityState (userAddresses, foundAddres
             stateFound = possibleLocation.state
 
             if (isSameAddressCityState(userCity, userState, cityFound, stateFound)) {
-                return { cityFound, stateFound }
+                return true
             }
```
shakyShane, line 27: I realised this method only ever considers a truthy result - so I made that explicit

unit-test/broker-protection.js
```diff
@@ -189,11 +189,11 @@ describe('Actions', () => {
 
             describe('matchAddressFromAddressListCityState', () => {
                 it('should match when city/state is present', () => {
-                    expect(matchAddressFromAddressListCityState(userData.addresses, ['chicago, il', 'schaumburg, il']))
+                    expect(matchAddressFromAddressListCityState(userData.addresses, [{ city: 'chicago', state: 'il' }])).toBe(true)
                 })
 
                 it('should not match when city/state is not present', () => {
-                    expect(matchAddressFromAddressListCityState(userData.addresses, ['los angeles, ca', 'portland, or']))
+                    expect(matchAddressFromAddressListCityState(userData.addresses, [{ city: 'los angeles', state: 'ca' }])).toBe(false)
```
shakyShane, line 196: these were bugs.

integration-test/playwright/page-objects/broker-protection.js
```diff
@@ -116,7 +103,7 @@ export class BrokerProtectionPage {
     /**
      * Simulate the native-side pushing an action into the client-side JS
      *
-     * @param {'extract.json' | 'extract2.json' | 'extract3.json' | 'extract4.json' | 'extract5.json' | 'extract-irregular1.json' | 'extract-irregular2.json' | 'extract-irregular3.json' | 'results2.json' | 'navigate.json' | 'fill-form.json' | 'click.json' | 'expectation.json' | 'get-captcha.json' | 'solve-captcha.json' | 'action-not-found.json'} action - add more action types here
+     * @param {string} action
```
shakyShane, line 106: just relaxing the type here - it'll fail during the test with a descriptive 'not found' error.

integration-test/test-pages/broker-protection/actions/click-parent.json
```diff
@@ -0,0 +1,42 @@
...
+                  "selector": ".age"
+                },
+                "addressCityState": {
+                  "selector": ".address"
+                }
+              }
+            }
+          }
+        }
+      ]
```
shakyShane, line 26: This is the example where the element matches, and the click should occur on the element - that's why it's `"selector": "."` on line 8

integration-test/test-pages/broker-protection/actions/click-weighted.json
```diff
@@ -0,0 +1,42 @@
...
+                "age": {
+                  "selector": ".age"
+                },
+                "addressCityState": {
+                  "selector": ".address"
+                }
+              }
+            }
+          }
+        }
```
shakyShane, line 25: and this example shows how to match an outer element first, and then perform a sub-query inside.

src/features/broker-protection/actions/extract.js
brianhall: This is well thought-through, nicely done. I love the flexibility of letting the consumer decide how to use the results.

brianhall (approved):
This looks great, @shakyShane. I can see a world where we expand some of the scoring methodology in a debug scenario (like giving a list of which data matches actually caused the score to increment), but for now this should work great.

PR #915, opened 2024-02-12
adding support for file upload

shakyShane:


src/features/broker-protection/actions/fill-form.js
```diff
@@ -17,12 +22,26 @@ export function fillForm (action, userData) {
         // let inputElem = form.elements[element.selector]
         // find the correct userData to put in the form
         if (inputElem) {
-            // @ts-expect-error - double check if this is strict enough
-            // todo: determine if this requires any events to be dispatched also
-            setValueForInput(inputElem, userData[element.type])
+            if (element.type === '$file_id$') {
```
brianhall, line 25: Why does the id look like this (vs. 'file')?

brianhall (approved):

src/features/broker-protection/actions/fill-form.js
```diff
@@ -54,5 +73,38 @@ function setValueForInput (el, val) {
...
+        for (let i = 0; i < length; i++) {
+            bytes[i] = binaryString.charCodeAt(i)
+        }
+
+        // Create the Blob from the Typed Array
+        const blob = new Blob([bytes], { type: 'image/png' })
+
+        // Create a DataTransfer object and append the Blob
+        const dataTransfer = new DataTransfer()
+        dataTransfer.items.add(new File([blob], 'id.png', { type: 'image/png' }));
```
brianhall, line 101: I don't think it's necessary now, but we may want to consider randomizing the filename at some point in the future.

brianhall (dismissed):

src/features/broker-protection/actions/fill-form.js
shakyShane: because all other 'types' here represent a key in the data object - but this is the first instance we have where the 'type' indicates a none-data-lookup. so, the `$`'s are there to make the difference clear - not sure if this is a good/bad idea yet, I propose we merge & work with it and the re-address if we come across a better solution :)

src/features/broker-protection/actions/fill-form.js
shakyShane: agreed - let's follow up as needed.

PR #916, opened 2024-02-14
Extract identifier from profile url

brianhall:
Asana: https://app.asana.com/0/0/1205912576739062/f
Accompanying broker JSON PR: https://github.com/duckduckgo/data-broker-protection-json/pull/40

Additional context: https://app.asana.com/0/608920331025329/1205560036177758

## Description

We're [adding](https://github.com/duckduckgo/data-broker-protection-json/pull/40) a profile identifier to the broker JSON that should allow us to extract a unique id from a profile URL (vs. using the entire URL, which is what we're doing currently). This prevents us from having duplicate information in the database due to url params being in a different order, etc.

### DIFF
--- a/src/features/broker-protection/actions/extract.js
+++ b/src/features/broker-protection/actions/extract.js
@@ -191,7 +191,7 @@ export function aggregateFields (profile) {
         addresses,
         phoneNumbers,
         relatives: profile.relativesList,
-	...profile.profileUrl
+        ...profile.profileUrl
     }
 }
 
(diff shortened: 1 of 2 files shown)


Bunn:
Just closing the loop on our conversation about MM.
It would be beneficial for the `identifier` to always be returned, either with the ID or the profileURL as a fallback. This way, the native apps only have to check a single property.
Thanks.

src/features/broker-protection/actions/extract.js
```diff
@@ -191,7 +194,7 @@ export function aggregateFields (profile) {
         addresses,
         phoneNumbers,
         relatives: profile.relativesList,
-        profileUrl: profile.profileUrl
+        ...profile.profileUrl
```
jonathanKingston, line 197: Identifier here seems a little ambiguous as a field name. I don't really understand the code if that matters but it could be confusing further down the line.

brianhall: Thanks @jonathanKingston, I think it's ok in the context of the object returned (which is a profile-specific object). Unless @Bunn disagrees, I think it's ok to keep it as `identifier` for now.

Bunn: yeah, it's an identifier of the object in the context. LGTM

jonathanKingston (approved):

PR #918, opened 2024-02-20
Strip newlines from name matches

brianhall:
Asana: https://app.asana.com/0/0/1206646445020571/f

Sometimes when the parts of the name (first/middle/last) are on different lines, C-S-S returns the string with newlines in between the parts. This minor tweak strips those newlines out.


shakyShane (approved):

shakyShane:
@brianhall as mentioned in our chat today, I think we can address this kind of problem more broadly later, for now we can accept this localized change to unblock :)

PR #920, opened 2024-02-25
Broker Protection: Extract action returns unmatched records for testing

gtfiorentino:


shakyShane:
@gtfiorentino we've made progress here already in #913 - that now returns a score + the element which we can build on to create debug features :)

shakyShane:
@gtfiorentino in particular, see this comment https://github.com/duckduckgo/content-scope-scripts/pull/913/files#r1481796012

src/features/broker-protection/actions/extract.js
```diff
@@ -141,38 +146,33 @@ function findFromElement (profileElement, dataKey, extractField) {
...
-    }
-
-    if (scrapedData.phone) {
-        if (userData.phone === scrapedData.phone) { return true }
+        matches.address = !!matchesFullAddress(userData.addresses, scrapedData.addressFull)
     }
 
-    // if phone number matches
-    return false
+    return matches.name && matches.age && matches.address;
```
shakyShane, line 175: this doesn't match the existing logic - previously you could match just a `name` + `address`, without age (if the scraped data didn't contain it).

regardless, we can now build upon what I started in #913 and return any amount of additional data for debug/analysis :)

PR #925, opened 2024-02-29
Expand Name Alternatives

brianhall:
Asana: https://app.asana.com/0/0/1206446019903282/f

When a user enters a name, we look to see if there are any nicknames (i.e. find John given Johnathan). Our existing list was very limited, and didn't do the reverse lookup (i.e. find Johnathan given John). This PR greatly expands the nicknames list and implements the reverse full-name lookup.

The source for the new list is https://github.com/carltonnorthern/nicknames. I haven't vetted it super closely and think there are some seemingly uncommon names on there (i.e. epaphroditius?!) but it's harmless to keep in.


src/features/broker-protection/comparisons/is-same-name.js
```diff
@@ -20,16 +20,26 @@ export function isSameName (fullNameExtracted, userFirstName, userMiddleName, us
...
         }
 
+        // If the user didn't supply a middle name, then try to match names extracted names that
+        // might include a middle name.
+        if (!userMiddleName) {
+            const matchesFirstAndLast = new RegExp(`^${firstName}\\s?.*\\s${userLastName}$`)
+
+            if (matchesFirstAndLast.test(fullNameExtracted)) {
+                return true
+            }
```
shakyShane, line 40: Let's make this driven by array comparisons instead of regex. Mostly because I think regex should be a last resort, but more importantly because the next iteration of matching names (and everything else) can be driven in this data-first, declarative style too (instead of how it's implemented now with string interpolation etc)

so, to remove this regex here + be inline with future changes, we could do something like this:

```suggestion
            const elements = fullNameExtracted.split(' ')
            // create the subject, in this case first + last elements in the string
            const subject = [elements[0], elements[elements.length - 1]]
            // target is current name/nickname + user entered last name
            const target = [firstName, userLastName]
            // compare
            const matchesFirstAndLast = subject.join(' ') === target.join(' ')

            if (matchesFirstAndLast) {
                return true
            }
```

src/features/broker-protection/comparisons/is-same-name.js
```diff
@@ -133,11 +143,58 @@ export function isSameName (fullNameExtracted, userFirstName, userMiddleName, us
     return false
 }
 
+/**
+ * Given the user's provided name, look for nicknames or full names and return a list
+ *
+ * @param {string | null} name
+ * @return {Set.<string>|Set<*>}
```
shakyShane, line 150: IDEs tend to create overly wide types like this, but we should clean it up and just have `Set<string>`

src/features/broker-protection/comparisons/is-same-name.js
```diff
@@ -133,11 +143,58 @@ export function isSameName (fullNameExtracted, userFirstName, userMiddleName, us
...
+    name = name.toLowerCase()
+
+    return new Set([name, ...getNicknames(name), ...getFullNames(name)])
+}
+
+/**
+ * Given a full name, get a list of nicknames, e.g. Gregory -> Greg
+ *
+ * @param {string | null} name
+ * @return {Set.<string>|Set<*>}
```
shakyShane, line 164: same thing with the types, I'll omit this from other similar ones :)

src/features/broker-protection/comparisons/is-same-name.js
```diff
@@ -133,11 +143,58 @@ export function isSameName (fullNameExtracted, userFirstName, userMiddleName, us
...
+ */
 export function getNicknames (name) {
-    if (name == null || name.trim() === '') { return [] }
+    const emptySet = new Set()
+
+    if (name == null || name.trim() === '') { return emptySet }
```
shakyShane, line 169: this same comparison keeps popping up, let's add a helper for it

```js
/**
 * This will handle all none-string types like null / undefined too
 * @param {any} [input]
 * @return {input is string}
 */
function noneEmptyString (input) {
    if (typeof input !== 'string') return false
    return input.trim().length > 0
}
```

then for each place:

```js
if (!noneEmptyString(name)) return new Set()
```

etc..

src/features/broker-protection/comparisons/constants.js
```diff
@@ -1,89 +1,1476 @@
...
-    tim: ['tim', 'timothy'],
-    timothy: ['timothy', 'tim'],
-    tom: ['tom', 'thomas', 'tommy'],
-    tommy: ['tommy', 'tom', 'thomas'],
-    tony: ['tony', 'anthony'],
-    will: ['will', 'william'],
-    william: ['william', 'will'],
-    zach: ['zach', 'zachary'],
-    zachary: ['zachary', 'zach']
+    aaron: ['erin', 'ronnie', 'ron'],
```
shakyShane, line 3: this list is huge now (which is great!) but we might want to defer the evaluation until we know it's needed. For example, after bundling, this particular module might get hoisted into the first execution context, so things like duck player (living in the same bundle) would be incurring the memory cost of loading this.

shakyShane (changes requested):

src/features/broker-protection/comparisons/constants.js
brianhall: Yep, that's a good idea. Will split this list out and dynamically import.

src/features/broker-protection/comparisons/is-same-name.js
brianhall: @shakyShane I considered doing this in my first approach, but I'm worried that this won't reliably get the first and last name (for instance, when someone has a last name with a space in it), which is why I went the regex route. 

Since the full extracted name string is the source of truth, it seemed best to try and not mess with the formatting too much since we don't know exactly what we're going to get. What do you think?

shakyShane: This conversation is proof we don't have enough test data yet - because if either of us thinks an approach is 'better', then we should be able to encode that proof into the tests. 

so yeah, if you wanted to keep the regex (I appreciate how you have this problem in your head better than I do, so I trust your judgement), could we at least add another test case that would fail with my example? that way we're building up more verifications for when we refactor

src/features/broker-protection/comparisons/constants.js
shakyShane: it will still have to be in the same bundle (we can't load files dynamically) - it's more about the variable that holds this list being at the top-level of the module/file that it's in. that could cause it to be evaluated for everyone. I can check this before you make the change

src/features/broker-protection/comparisons/is-same-name.js
brianhall: Yeah, that's fair enough. I will do. I think we'll have enough test data soon to switch approaches (or alter the proposed approach to fit).

src/features/broker-protection/comparisons/constants.js
brianhall: Ah, ok. Yeah, that'd be great. Thank you.

src/features/broker-protection/comparisons/is-same-name.js
gtfiorentino: I think the regex approach is fine since we want to tend to overmatch rather than undermatch. Agree we should have a "false positive" test case to make sure we don't go too far in the other direction.

shakyShane: with the following test, you can see that interpolating user data into regexes is overly risky

```js
it('property testing isSameName (seed 1)', () => {
    // this seeded example showed how we needed to escape the user-values in the regex
    fc.assert(fc.property(
        fc.string(),
        fc.string(),
        fc.option(fc.string()),
        fc.string(),
        fc.option(fc.string()),
        (fullNameExtracted, userFirstName, userMiddleName, userLastName, userSuffix) => {
            const result = isSameName(fullNameExtracted, userFirstName, userMiddleName, userLastName, userSuffix)
            expect(typeof result).toBe('boolean')
        }
    ), { seed: 493629956, path: '50:1:0:0:1:2:1:1:6:7:7', endOnFailure: true })
})
```

with that seed, you get:

Screenshot 2024-02-29 at 1 25 50 PM

Looking again at the problem, we're just saying:
- a string that begins with `x` 
- AND ends in `y`
- AND has letters/spaces inbetween

therefor, this would work

```js
        if (!userMiddleName) {
            const combinedLength = firstName.length + userLastName.length
            const matchesFirstAndLast = fullNameExtracted.startsWith(firstName) &&
                fullNameExtracted.endsWith(userLastName) &&
                fullNameExtracted.length > combinedLength
            if (matchesFirstAndLast) {
                return true
            }
        }
```

brianhall: I’m happy with that approach. 👍

brianhall:
This LGTM @shakyShane, thanks again.

gtfiorentino (approved):
In my testing of over 300 records this changes drops the false positive rate from 100% to 46%. Remaining issues seem to be with Address matching but we should ship this and start getting feedback on it.

brianhall:
false positive -> false negative, right @gtfiorentino?

gtfiorentino:
Yes thank you, edited my last comment so it's not confusing

PR #927, opened 2024-03-01
Add the ability to generate a random phone number when filling out forms

brianhall:
Asana: https://app.asana.com/0/0/1206496978629598/f

## Description
Some brokers require phone numbers in order to proceed with opt out requests. We don't ask for user's phone numbers (nor would we want to send real numbers to brokers) so this feature allows us to generate a random phone number on the fly.

In order to prevent sending real phone numbers, I referenced the following resources:
* [Wikipedia article](https://en.wikipedia.org/wiki/555_(telephone_number)#Fictional_usage) on US phone numbers, revealing that 555-0100 and 555-0199 are for fictional use.
* [This stackexchange article](https://math.stackexchange.com/questions/920972/how-many-different-phone-numbers-are-possible-within-an-area-code/1115411#1115411) outlining the safe space for area codes.

### DIFF
--- a/integration-test/test-pages/broker-protection/actions/fill-form.json
+++ b/integration-test/test-pages/broker-protection/actions/fill-form.json
@@ -12,6 +12,10 @@
         {
           "type": "lastName",
           "selector": "#user_last_name"
+        },
+        {
+          "type": "phone",
+          "selector": "#user_phone"
         }
       ]
     },
--- a/src/features/broker-protection/actions/fill-form.js
+++ b/src/features/broker-protection/actions/fill-form.js
@@ -1,4 +1,4 @@
-import { getElement } from '../utils.js'
+import { getElement, generateRandomInt } from '../utils.js'
 import { ErrorResponse, SuccessResponse } from '../types.js'
 
 /**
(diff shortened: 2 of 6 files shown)


integration-test/test-pages/broker-protection/actions/fill-form.json
```diff
@@ -12,6 +12,10 @@
         {
           "type": "lastName",
           "selector": "#user_last_name"
+        },
+        {
+          "type": "phone",
```
shakyShane, line 17: I used a `$` prefix + suffix in a previous PR to differentiate these generated fields from data lookups - we should do the same here, just make it something like `"type": "$generated_phone_number$" or similar. Otherwise it looks like 'phone' is coming from the data object

shakyShane (changes requested):

src/features/broker-protection/actions/fill-form.js
```diff
@@ -18,16 +18,18 @@ export function fillForm (action, userData) {
...
         const inputElem = getElement(form, element.selector)
-        // this works for IDs (i.e. #url wouldb e form.elements['url'])
+        // this works for IDs (i.e. #url would be form.elements['url'])
         // let inputElem = form.elements[element.selector]
         // find the correct userData to put in the form
         if (inputElem) {
             if (element.type === '$file_id$') {
                 results.push(setImageUpload(inputElem))
             } else {
+                const value = element.type === "phone" ? generatePhoneNumber() : userData[element.type]
```
shakyShane, line 28: let's add another separate condition here - `if (element.type === '$generated_phone_number$')` or similar.

integration-test/test-pages/broker-protection/actions/fill-form.json
brianhall: Ah, I hadn't originally understood why you went with that naming convention, now it makes total sense. Updated.

brianhall:
Addressing the linting issues now... 😄

shakyShane (approved):

PR #931, opened 2024-03-04
PIR: prevent invalid value for age

shakyShane:
See https://app.asana.com/0/0/1206745203094838/f

Building upon #930 - this is an actual fix that is reproducible in the tests now.

PR #932, opened 2024-03-06
PIR: don't exclude addressFull

shakyShane:
https://app.asana.com/0/0/1206775911419667/f

In the previous change #931 - we're now respecting when `extract` returns null (as a way of validating data)

In doing that, it accidentally excludes valid values for `addressFull` since that was the only field that does not incur processing.

**NOTE** I have verified no other fields are affected by checking our JSON files.

### DIFF
--- a/src/features/broker-protection/actions/extract.js
+++ b/src/features/broker-protection/actions/extract.js
@@ -152,13 +152,7 @@ function findFromElements (elements, key, extractField) {
     const elementValues = []
     if (elements) {
         for (const element of elements) {
-            let elementValue = rules[key]?.(element) ?? element?.innerText ?? null
-            if (extractField?.afterText) {
-                elementValue = elementValue?.split(extractField.afterText)[1]?.trim() || elementValue
-            }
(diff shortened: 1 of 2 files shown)


shakyShane:
* **#932** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/932?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/features/broker-protection/actions/extract.js
```diff
@@ -152,13 +152,7 @@ function findFromElements (elements, key, extractField) {
     const elementValues = []
     if (elements) {
         for (const element of elements) {
-            let elementValue = rules[key]?.(element) ?? element?.innerText ?? null
```
brianhall, line 155: Weird, I hadn't even noticed that we were duplicating this logic in `findFromElement`. 🤦

brianhall (approved):

brianhall:
This LGTM

PR #933, opened 2024-03-06
pir: parsing accuracy

shakyShane:
https://app.asana.com/0/0/1206887336167127/f

## Changes
- This change allows each field to be processed by separate `extractors`. 
- I've reduced the amount of places where types were sometimes strings, and sometimes arrays of strings
- Addresses are now represented everywhere as just city/state pairs 
  - We also exclude addresses where the state is not in a valid format
  - This makes matching simpler and removes most of the code related to addresses
- We now do a better job of handling duplicates

## New
- We now remove common suffixes/prefixes from strings before passing to extractors
  - This makes our declarative format less flaky and the list is small/focussed

## Extractors

Using Typescript to describe the new extractors:

```ts
interface ExtractorParams {
  afterText?: string
  beforeText?: string
}

interface Extractor<JsonValue> {
  extract(strings: string[], params: ExtractorParams): JsonValue
}

/**
 * Example extractor - note how the type param `JsonValue` is filled
 * in by the implementor.
 */
const myNewExtractor: Extractor<string[]> = {
  extract: function (strings: string[], params: ExtractorParams): string[] {
    throw new Error("Function not implemented.")
  }
}

```


shakyShane:
* **#933** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/933?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/features/broker-protection/actions/extract.js
```diff
@@ -128,12 +134,15 @@ export function createProfile (elementFactory, extractData) {
...
-            // Note: This can return a string, string[], or null
-            const extractedValue = extractValue(key, value, evaluatedValue)
+            // extract all strings first
+            const evaluatedValues = stringValuesFromElements(elements, key, value)
+
+            // clean them up - trimming, removing empties
+            const noneEmptyArray = cleanArray(evaluatedValues)
+
+            // Note: This can return any valid JSON valid, it depends on the extractor used.
+            const extractedValue = extractValue(key, value, noneEmptyArray)
```
shakyShane, line 145: This is a bigger change than it looks. Previously the extract step would sometimes get an array, and sometimes a string. This made it difficult to reason about. Now all extractors receive a `string[]` for consistency

src/features/broker-protection/actions/extract.js
```diff
@@ -205,33 +202,19 @@ function scrapedDataMatchesUserData (userData, scrapedData) {
...
+    ]
+
+    for (const addressField of addressFields) {
+        if (addressField in scrapedData) {
+            if (addressMatch(userData.addresses, scrapedData[addressField])) {
+                matchedFields.push(addressField)
+                return { matchedFields, score: matchedFields.length, result: true }
+            }
         }
     }
```
shakyShane, line 219: because addresses are standardised everywhere now, via the address extractors, it means we can check them all for matches in the exact same way :)

src/features/broker-protection/actions/extract.js
```diff
@@ -250,21 +233,34 @@ function scrapedDataMatchesUserData (userData, scrapedData) {
...
-
+    // addresses
+    const combinedAddresses = [
+        ...profile.addressCityState || [],
+        ...profile.addressCityStateList || [],
+        ...profile.addressFullList || [],
+        ...profile.addressFull || []
+    ]
+    const addressMap = new Map(combinedAddresses.map(addr => [`${addr.city},${addr.state}`, addr]))
+    const addresses = [...addressMap.values()]
```
shakyShane, line 244: we can now report all found addresses, since we know they'll be valid if we got this far. The Map helps to de-dupe. I avoided JSON.stringify since the ordering of the keys could affect it.

src/features/broker-protection/actions/extract.js
```diff
@@ -278,114 +274,46 @@ export function aggregateFields (profile) {
...
-
-        const state = words.pop()
-        const city = words.join(' ')
-
-        output.push({ city, state: state || null })
-    }
-    return output
+    const defaultSeparator = /[|\n•·]/
+    return cleanArray(inputList.split(separator || defaultSeparator))
 }
```
shakyShane, line 317: in the end, this function was just 'split with a separator or default separator'

src/features/broker-protection/comparisons/address.js
```diff
@@ -1,74 +1,17 @@
...
-    if ((city === comparisonCity) && (state === comparisonState)) { return true }
-
-    return false
+export function addressMatch (userAddresses, foundAddresses) {
+    return userAddresses.some((user) => {
+        return foundAddresses.some(found => {
+            return matchingPair(user.city, user.city) && matchingPair(user.state, found.state)
+        })
+    })
 }
```
shakyShane, line 15: Now that addresses are the same everywhere, matching is much easier

src/features/broker-protection/extractors/address.js
```diff
@@ -0,0 +1,73 @@
...
+    extract (strs, extractorParams) {
+        const cityStateList = strs.map(str => stringToList(str, extractorParams.separator)).flat()
+        return getCityStateCombos(cityStateList)
+    }
+}
+
+/**
+ * @implements {Extractor<{city:string; state: string|null}[]>}
+ */
+export class AddressFullExtractor {
```
shakyShane, line 24: This was just moved from the original place so that it can live alongside the extractor

src/features/broker-protection/extractors/address.js
```diff
@@ -0,0 +1,73 @@
...
+                return { city: addr.city, state: addr.state || null }
+            })
+    }
+}
+
+/**
+ * @param {string[]} inputList
+ * @return {{ city: string, state: string|null }[] }
+ */
+function getCityStateCombos (inputList) {
```
shakyShane, line 47: This was just moved from the original place so that it can live alongside the extractor

unit-test/broker-protection-extractors.js
```diff
@@ -0,0 +1,16 @@
...
+describe('individual extractors', () => {
+    describe('PhoneExtractor', () => {
+        it('should extract digits only', () => {
+            fc.assert(fc.property(fc.array(fc.string()), (s) => {
+                const cleanInput = cleanArray(s)
+                const numbers = new PhoneExtractor().extract(cleanInput, {})
+                const cleanOutput = cleanArray(numbers)
+                return cleanOutput.every(num => num.match(/^\d+$/))
+            }))
+        })
```
shakyShane, line 14: just an example of how we can start to stress-test the extractors as needed

shakyShane:
@brianhall please take a look at these changes - in particular to understand the changes I recommend looking at the new tests and the ones that were altered. Every change made here is represented in those tests one way or another

src/features/broker-protection/actions/extract.js
```diff
@@ -278,114 +274,46 @@ export function aggregateFields (profile) {
...
+    case 'addressCityState':
+    case 'addressCityStateList':
+        return new CityStateExtractor().extract(elementValues, extractorParams)
+
+    case 'alternativeNamesList': return new AlternativeNamesExtractor().extract(elementValues, extractorParams)
+    case 'relativesList': return new RelativesExtractor().extract(elementValues, extractorParams)
+    case 'phone':
+    case 'phoneList':
+        return new PhoneExtractor().extract(elementValues, extractorParams)
+    case 'profileUrl': return new ProfileUrlExtractor().extract(elementValues, extractorParams)
```
brianhall, line 304: Should we have a default case that throws a warning? We can leave that for another time if you'd like since we're not doing anything with warnings yet.

src/features/broker-protection/extractors/profile-url.js
```diff
@@ -0,0 +1,28 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { Extractor } from '../types.js'
+import { getIdFromProfileUrl } from '../actions/extract.js'
```
brianhall, line 3: Should we just move `getIdFromProfileUrl` into this file? I don't think it's called from anywhere else.

src/features/broker-protection/comparisons/address.js
```diff
@@ -1,74 +1,17 @@
...
-    state = state.toLowerCase()?.trim()
-    comparisonState = comparisonState.toLowerCase()?.trim()
-
-    if ((city === comparisonCity) && (state === comparisonState)) { return true }
-
-    return false
+export function addressMatch (userAddresses, foundAddresses) {
+    return userAddresses.some((user) => {
+        return foundAddresses.some(found => {
+            return matchingPair(user.city, user.city) && matchingPair(user.state, found.state)
```
brianhall, line 12: Ah, I think this should be:

```suggestion
            return matchingPair(user.city, found.city) && matchingPair(user.state, found.state)
```

brianhall (changes requested):
Mostly just a bunch of smaller comments and found one little 🐛.

src/features/broker-protection/comparisons/address.js
shakyShane: good catch - I need to add a new test case to cover this

jotaemepereira (approved):
I’ve tested these changes in the macOS browser, and it works as expected 👍🏼 .

src/features/broker-protection/extractors/profile-url.js
shakyShane: yep - will follow up :)

src/features/broker-protection/actions/extract.js
```diff
@@ -414,3 +342,56 @@ export function getIdFromProfileUrl (profileUrl, identifierType, identifier) {
...
+ * @return {string}
+ */
+function removeCommonSuffixesAndPrefixes (elementValue) {
+    const regexes = [
+        // match text such as +3 more when it appears at the end of a string
+        /\+\s*\d+.*$/
+    ]
+    // strings that are always safe to remove from the start
+    const startsWith = [
+        'Associated persons:',
```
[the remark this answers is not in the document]
shakyShane: let's follow up with more targeted fixes like this.

src/features/broker-protection/actions/extract.js
shakyShane: yep, let's follow up

src/features/broker-protection/comparisons/address.js
shakyShane: resolved

PR #934, opened 2024-03-13
Performance metrics feature

SlayterDev:
This PR creates the web-vitals feature which collects the `first-contentful-paint` start time.

https://app.asana.com/0/72649045549333/1206601617546788/f

### DIFF
--- a/src/features/web-vitals.js
+++ b/src/features/web-vitals.js
@@ -3,18 +3,29 @@ import { onLCP, onFID, onCLS } from 'web-vitals';
 
 export default class WebVitals extends ContentFeature {
     init() {
+        console.log('VITALS ALIVE')
         this.messaging.subscribe('getVitals', async (/** @type {any} */params) => {
+            console.log('MESSAGE RECEIVED')
             this.messaging.notify('vitalsResult', { 'vitals': [this.lcp, this.fid, this.cls] })
         })
+    }
 
-        onLCP(metric => {
-            this.lcp = metric.value
-        })
-        onFID(metric => {
-            this.fid = metric.value
-        })
-        onCLS(metric => {
-            this.cls = metric.value
-        })
+    load() {
+        // onLCP((metric) => {
+        //     console.log(metric)
+        //     this.lcp = metric.value
+        // }, {reportAllChanges: true})
+        // onFID((metric) => {
+        //     console.log(metric)
+        //     this.fid = metric.value
+        // }, {reportAllChanges: true})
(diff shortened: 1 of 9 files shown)


shakyShane:
@SlayterDev a quick `npm run lint-fix` should fix up the failing build and make the review easier - also it looks to be still in progress? if so you can mark it as a draft 👌🏻

SlayterDev:
@shakyShane its ready for review.

src/features/web-vitals.js
```diff
@@ -0,0 +1,11 @@
+import ContentFeature from '../content-feature'
+
+export default class WebVitals extends ContentFeature {
+    init () {
+        this.messaging.subscribe('getVitals', () => {
+            const paintResources = performance.getEntriesByType('paint')
+            const firstPaint = paintResources.find((entry) => entry.name === 'first-contentful-paint')
+            this.messaging.notify('vitalsResult', { vitals: [firstPaint?.startTime] })
```
shakyShane, line 8: if the entry is absent then this will send `vitals: [null]` (array with a single `null` item in it) as the message params - not sure if that's how you intended to handle this situation on the native side?

if not, you could just wrap in an `if` condition and only send if `firstPaint` is found

shakyShane (changes requested):

src/features/web-vitals.js
SlayterDev: Its not a problem on the native side. It just parses to `nil` eventually. But I can change it if it feels wrong on the JS side.

shakyShane: yeah, I know this is super picky - but very soon we're landing something that enforces all boundary messages like this to be defined in a schema, so we wouldn't want to enable what is effectively `(string|null)[]`.

anyway, if you just do this instead it'll send an empty array 👌🏻 

```js
const paintResources = performance.getEntriesByType('paint')
const vitals = paintResources.filter((entry) => entry.name === 'first-contentful-paint')
this.messaging.notify('vitalsResult', { vitals })
```

shakyShane: actually scratch that

or, to prevent needing to pluck out 'startTime' in that example, this instead?

```js
const paintResources = performance.getEntriesByType('paint')
const firstPaint = paintResources.filter((entry) => entry.name === 'first-contentful-paint')
const vitals = firstPaint ? [firstPaint.startTime] : []
this.messaging.notify('vitalsResult', { vitals })
```

jonathanKingston (approved):

PR #935, opened 2024-03-14
fix(pir): support filling select fields

shakyShane:
https://app.asana.com/0/0/1206840388477213/f

adds explicit support for select fields with the same event emulation we do with inputs.

### DIFF
--- a/integration-test/playwright/page-objects/broker-protection.js
+++ b/integration-test/playwright/page-objects/broker-protection.js
@@ -43,9 +43,10 @@ export class BrokerProtectionPage {
      * @return {Promise<void>}
      */
     async isFormFilled () {
-        expect(await this.page.getByLabel('First Name:').inputValue()).toBe('John')
-        expect(await this.page.getByLabel('Last Name:').inputValue()).toBe('Smith')
-        expect(await this.page.getByLabel('Phone Number:').inputValue()).toMatch(/^\d{10}$/)
+        await expect(this.page.getByLabel('First Name:')).toHaveValue('John')
+        await expect(this.page.getByLabel('Last Name:')).toHaveValue('Smith')
+        await expect(this.page.getByLabel('Phone Number:')).toHaveValue(/^\d{10}$/)
+        await expect(this.page.getByLabel('State:')).toHaveValue('IL')
     }
 
     /**
--- a/integration-test/test-pages/broker-protection/actions/fill-form.json
+++ b/integration-test/test-pages/broker-protection/actions/fill-form.json
@@ -16,6 +16,10 @@
         {
           "type": "$generated_phone_number$",
           "selector": "#user_phone"
+        },
+        {
+          "type": "state",
+          "selector": "#state"
         }
       ]
     },
--- a/src/features/broker-protection/actions/fill-form.js
+++ b/src/features/broker-protection/actions/fill-form.js
@@ -4,11 +4,10 @@ import { ErrorResponse, SuccessResponse } from '../types.js'
 /**
  * @param {Record<string, any>} action
  * @param {Record<string, any>} userData
- * @param {Document | HTMLElement} [root]
  * @return {import('../types.js').ActionResponse}
  */
-export function fillForm (action, userData, root = document) {
-    const form = getElement(root, action.selector)
+export function fillForm (action, userData) {
+    const form = getElement(document, action.selector)
     if (!form) return new ErrorResponse({ actionID: action.id, message: 'missing form' })
 
     // todo: also utilize 'warnings' here later.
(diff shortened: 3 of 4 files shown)


shakyShane:
* **#935** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/935?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/features/broker-protection/actions/fill-form.js
```diff
@@ -4,36 +4,15 @@ import { ErrorResponse, SuccessResponse } from '../types.js'
 /**
  * @param {Record<string, any>} action
  * @param {Record<string, any>} userData
+ * @param {Document | HTMLElement} [root]
  * @return {import('../types.js').ActionResponse}
  */
-export function fillForm (action, userData) {
-    const form = getElement(document, action.selector)
+export function fillForm (action, userData, root = document) {
```
brianhall, line 10: Is the idea behind adding the root parameter that we could potentially do something similar to what you've done with extract (where you can pass a parent element)?

integration-test/playwright/page-objects/broker-protection.js
```diff
@@ -43,9 +43,10 @@ export class BrokerProtectionPage {
      * @return {Promise<void>}
      */
     async isFormFilled () {
-        expect(await this.page.getByLabel('First Name:').inputValue()).toBe('John')
-        expect(await this.page.getByLabel('Last Name:').inputValue()).toBe('Smith')
-        expect(await this.page.getByLabel('Phone Number:').inputValue()).toMatch(/^\d{10}$/)
+        await expect(this.page.getByLabel('First Name:')).toHaveValue('John')
+        await expect(this.page.getByLabel('Last Name:')).toHaveValue('Smith')
+        await expect(this.page.getByLabel('Phone Number:')).toHaveValue(/^\d{10}$/)
+        await expect(this.page.getByLabel('State:')).toHaveValue('IL')
```
shakyShane, line 49: this test would have failed before this change 👌🏻

integration-test/test-pages/broker-protection/actions/fill-form.json
```diff
@@ -16,6 +16,10 @@
         {
           "type": "$generated_phone_number$",
           "selector": "#user_phone"
+        },
+        {
+          "type": "state",
```
brianhall, line 21: Not needed for this PR, but it occurs to me that we might need some additional flexibility in the future, like `stateFull` instead of `state` for the value, depending on how the broker has their `select` defined. Just making a mental note more than anything else. 😄

src/features/broker-protection/actions/fill-form.js
```diff
@@ -47,35 +26,80 @@ export function fillForm (action, userData) {
...
+        return { result: false, error: `input type was not supported: ${el.tagName}` }
+    }
+
+    try {
+        el.dispatchEvent(new Event('keydown', { bubbles: true }))
+
+        const originalSet = Object.getOwnPropertyDescriptor(target.prototype, 'value')?.set
+        originalSet?.call(el, val)
+
+        const events = [
```
brianhall, line 88: I noticed that the now the events we fire here for `select` are different than the [events fired for `select` in autofill](https://github.com/duckduckgo/duckduckgo-autofill/blob/6493e296934bf09277c03df45f11f4619711cb24/src/autofill-utils.js#L123-L136). Should we consider matching the implementation here?

src/features/broker-protection/actions/fill-form.js
```diff
@@ -47,35 +26,80 @@ export function fillForm (action, userData) {
...
+ * }}
+ */
+export function fillMany (root, elements, data) {
+    const results = []
+    const warnings = []
+
+    for (const element of elements) {
+        const inputElem = getElement(root, element.selector)
+        if (!inputElem) {
+            // todo: report this? it can occur when a selector is incorrect, but should we fail the entire action?
```
brianhall, line 46: I'm also wondering if we should fail here. In most cases, at least in the opt out form, we're only adding the required fields to the json, so if any of them aren't filled out then the form won't submit (and I think we'd want to know).

brianhall (commented):
This is looking nice! Added a few comments to get your thoughts.

src/features/broker-protection/actions/fill-form.js
shakyShane: yes we should, good catch

src/features/broker-protection/actions/fill-form.js
shakyShane: @brianhall ok, if you agree then we can simply push an error here instead

brianhall: Yeah, let's do it. Thank you!

shakyShane:
@brianhall I've addressed the 3 feedback points :)

brianhall (approved):
This LGTM

PR #938, opened 2024-03-15
pir: allow the root document to be specified in all actions

shakyShane:
https://app.asana.com/0/0/1206854307964888/f

This change ensures we can execute all actions in the context of our choosing, for debugging and testing.

### DIFF
--- a/src/features/broker-protection/actions/captcha.js
+++ b/src/features/broker-protection/actions/captcha.js
@@ -6,11 +6,12 @@ import { ErrorResponse, SuccessResponse } from '../types.js'
  * Gets the captcha information to send to the backend
  *
  * @param action
+ * @param {Document | HTMLElement} root
  * @return {import('../types.js').ActionResponse}
  */
-export function getCaptchaInfo (action) {
+export function getCaptchaInfo (action, root = document) {
     const pageUrl = window.location.href
-    const captchaDiv = getElement(document, action.selector)
+    const captchaDiv = getElement(root, action.selector)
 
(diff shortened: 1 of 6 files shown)


shakyShane:
* **#938** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/938?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

src/features/broker-protection/actions/captcha.js
```diff
@@ -80,14 +81,16 @@ export function getCaptchaInfo (action) {
 * Takes the solved captcha token and injects it into the page to solve the captcha
 *
 * @param action
+* @param {string} token
+* @param {Document} root
```
brianhall, line 85: Does this also need to be `{Document|HTMLElement}`?

brianhall (approved):

src/features/broker-protection/actions/captcha.js
shakyShane: Actually `HTMLElement` in this 1 particular case doesn't work - because `getElementsByName` is only on Document.

We should change the body of this function later to improve this, but for now I don't want to alter anything other than the params, so we can follow up :)

brianhall: Sounds perfect, thanks for the explanation.

shakyShane:
@brianhall thanks - I'll merge this now but won't cut a release until my fixes tomorrow :)

PR #944, opened 2024-03-26
pir: scroll element into view first

shakyShane:
https://app.asana.com/0/0/1206930392257127/f

The integration test is the best explanation here. 

Before the expectation action only supported `url` and `text`. This PR adds support for a third kind `element`

When the expectation has `type: "element"`, we'll first check if it needs a parent to be scrolled into view, and then we'll verify if the element exists.

NOTE: This change is safe because we're not currently using any other `types` (only `url` and `text` so far).

### DIFF
--- a/integration-test/test-pages/broker-protection/pages/form.html
+++ b/integration-test/test-pages/broker-protection/pages/form.html
@@ -45,6 +45,12 @@
     document.forms.testform.onsubmit = (e) => {
       e.preventDefault();
     }
+    setTimeout(() => {
+        const next = document.createElement('div')
+        next.classList.add('slow-element');
+        next.textContent = 'Slow loading element'
+        document.forms.testform.appendChild(next)
+    }, 1000)
 </script>
 </body>
 </html>
--- a/integration-test/test-pages/broker-protection/pages/form.html
+++ b/integration-test/test-pages/broker-protection/pages/form.html
@@ -46,6 +46,7 @@
       e.preventDefault();
     }
     setTimeout(() => {
+        document.body.classList.add('delay-complete')
         const next = document.createElement('div')
         next.classList.add('slow-element');
         next.textContent = 'Slow loading element'
--- a/src/features/broker-protection.js
+++ b/src/features/broker-protection.js
@@ -25,7 +25,10 @@ export default class BrokerProtection extends ContentFeature {
                     ? action.retry
                     : undefined
 
-                if (action.actionType === 'extract') {
+                /**
+                 * Special case for the exact action
+                 */
+                if (!retryConfig && action.actionType === 'extract') {
                     retryConfig = {
                         interval: { ms: 1000 },
                         maxAttempts: 30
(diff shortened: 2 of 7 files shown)


shakyShane:
* **#944** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/944?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

integration-test/test-pages/broker-protection/pages/form.html
```diff
@@ -45,6 +45,13 @@
     document.forms.testform.onsubmit = (e) => {
       e.preventDefault();
     }
+    setTimeout(() => {
+        document.body.classList.add('delay-complete')
+        const next = document.createElement('div')
+        next.classList.add('slow-element');
+        next.textContent = 'Slow loading element'
+        document.forms.testform.appendChild(next)
+    }, 1000)
```
shakyShane, line 54: just making changes after 1second, so that the integration test is as real as possible

src/features/broker-protection.js
```diff
@@ -25,13 +25,28 @@ export default class BrokerProtection extends ContentFeature {
...
+                 */
+                if (!retryConfig && action.actionType === 'expectation') {
+                    if (action.expectations.some(x => x.type === 'element')) {
+                        retryConfig = {
+                            interval: { ms: 1000 },
+                            maxAttempts: 30
+                        }
+                    }
+                }
+
```
shakyShane, line 49: This is the second such 'special case' for the retry config. Soon (if we add a third) we should consider where this _should_ live

src/features/broker-protection/actions/expectation.js
```diff
@@ -7,24 +7,135 @@ import { ErrorResponse, SuccessResponse } from '../types.js'
...
+/**
+ * Return a true/false result for every expectation
+ *
+ * @param {import("../types").Expectation[]} expectations
+ * @param {Document | HTMLElement} root
+ * @return {import("../types").BooleanResult[]}
+ */
+export function expectMany (expectations, root) {
+    return expectations.map(expectation => {
+        switch (expectation.type) {
```
brianhall, line 33: Worth a test to validate non-happy path cases here (invalid or missing type) to ensure we're getting back what we expect?

brianhall (approved):
This looks good. Neither of my comments are blocking.

src/features/broker-protection/actions/fill-form.js
```diff
@@ -12,6 +12,9 @@ export function fillForm (action, userData, root = document) {
     if (!form) return new ErrorResponse({ actionID: action.id, message: 'missing form' })
     if (!userData) return new ErrorResponse({ actionID: action.id, message: 'user data was absent' })
 
+    // ensure the element is in the current viewport
+    form.scrollIntoView?.()
```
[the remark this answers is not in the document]
shakyShane: This is not *technically* needed any more for the particular problem we were facing. But, it's a good thing to have in place anyway so I think we should just keep it in.

PR #951, opened 2024-04-10
only accept first 11 chars of video id

shakyShane:
https://app.asana.com/0/0/1206957893781164/f

### TL;DR

Modified validation and sanitization of video IDs in the duck-player module.

### What changed?

1. The `validateVideoId` function now checks first if the input is a string. If not, it returns `null`. If the input is a string, the function extracts the first 11 characters, validates them against a regex pattern and returns these characters if they pass the validation.
2. The function `getValidVideoId` now potentially returns `null` instead of `false`.

### How to test?

The testing responsibilities have been accounted for in the `duckplayer.spec.js` file. A test has been added that checks whether the ID validation correctly accepts only the first 11 characters of a video ID.

### Why make this change?

This change was required to improve the validation of video IDs, considering only the necessary characters and properly handling different data types. It also paves the way for more robust error handling by returning `null` instead of `false` in certain functions.

---

### DIFF
--- a/packages/special-pages/pages/duckplayer/src/js/index.js
+++ b/packages/special-pages/pages/duckplayer/src/js/index.js
@@ -278,6 +278,7 @@ const Comms = {
     messaging: undefined,
     /**
      * NATIVE NOTE: Gets the video id from the location object, works for MacOS < > 12
+     * @return {string}
      */
     getVideoIdFromLocation: () => {
         /**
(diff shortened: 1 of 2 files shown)


shakyShane:
* **#951** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/951?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/special-pages/pages/duckplayer/src/js/index.js
```diff
@@ -301,18 +302,20 @@ const Comms = {
...
-     * @returns {(string|false)}
+     * @returns {(string|null)}
      */
     validateVideoId: (input) => {
-        if (/^[a-zA-Z0-9-_]+$/g.test(input)) {
-            return input
+        if (typeof input !== 'string') return null
+        const subject = input.slice(0, 11)
+        if (/^[a-zA-Z0-9-_]+$/g.test(subject)) {
+            return subject
```
jonathanKingston, line 311: Should we return the 11nth onwards chars if they're valid? Do we know for certain that 11 is enough always etc?

packages/special-pages/pages/duckplayer/src/js/index.js
```diff
@@ -301,18 +302,20 @@ const Comms = {
      * Validates that the input string is a valid video id.
      * If so, returns the video id otherwise returns false.
      * @param {string} input
-     * @returns {(string|false)}
+     * @returns {(string|null)}
      */
     validateVideoId: (input) => {
```
jonathanKingston, line 307: This should probably be renamed to sanitise rather than validate if we're intending to manipulate.

packages/special-pages/pages/duckplayer/src/js/index.js
shakyShane: In this instance, we never want anything after the first 11 (other params are handled elsewhere) and yeah that number seems stable.

shakyShane:
@jonathanKingston I renamed that method.

jonathanKingston (approved):

PR #952, opened 2024-04-10
don't default to addtional properties in schemas

shakyShane:
https://app.asana.com/0/1201614831475344/1207053312757015/f

### TL;DR

This PR restricts additional properties in the build-types script.

### What changed?

The `createTypesForSchemaFile` and `createTypesForSchemaMessages` functions in `build-types.mjs` now include the `additionalProperties: false` option.

### Why make this change?

The library that generates the types makes open-ended objects by default by adding the index signature.

without this change, it's really tedious to create schemas since every top-level and nested object requires you to put `"additonalProperties": false`

---


shakyShane:
@jonathanKingston just a QOL thing, I noticed whilst trying to add types for PIR that this was a pain.

jonathanKingston (approved):

PR #956, opened 2024-04-24
APIs for shimming standard APIs

muodov:
Project: https://app.asana.com/0/72649045549333/1207011810532429/f

This PR adds ContentFeature APIs to shim standard interfaces classes and global instances. These APIs are designed to be used in webcompat shims, when we write polyfills for APIs that are missing on specific platforms.

---
  - https://app.asana.com/0/0/1207057867725390


src/features/web-compat.js
```diff
@@ -440,54 +440,30 @@ export class WebCompat extends ContentFeature {
...
+                /** @type {MediaSession['playbackState']} */
+                playbackState = 'none'
+
+                setActionHandler () {}
+                setCameraActive () {}
+                setMicrophoneActive () {}
+                setPositionState () {}
+            }
+
+            this.shimInterface('MediaSession', MyMediaSession, {
```
jonathanKingston, line 454: Could this be just:
```
this.shimInterface(MediaSession, {...})
```
And internally it use `arg[0].name` for the typing?

Would this omit the need to then mock `.name`?

muodov: That would require the class to always have the correct name, which we can't verify inside the method.
This file is not ready btw, the correct usages are in the test file.

src/content-feature.js
```diff
@@ -23,6 +23,8 @@ import { extensionConstructMessagingConfig } from './sendmessage-transport.js'
  */
 
 const globalObj = typeof window === 'undefined' ? globalThis : window
+// special proeprty that is set on classes used to shim standard interfaces
```
jonathanKingston, line 26:
```suggestion
// special property that is set on classes used to shim standard interfaces
```

src/content-feature.js
```diff
@@ -424,15 +425,211 @@ export default class ContentFeature {
...
+            /** @type {StrictDataDescriptor} */
+            // @ts-expect-error - As long as ImplClass is a normal class, it should have the prototype property
+            const descriptor = getOwnPropertyDescriptor(ImplClass.prototype, 'constructor')
+            if (descriptor.writable) {
+                ImplClass.prototype.constructor = Interface
+            }
+        }
+
+        // mark the class as a shimmed class
+        objectDefineProperty(ImplClass, ddgShimMark, {
```
jonathanKingston, line 566: I believe this is still visible in `getOwnPropertySymbols`, please validate :)

src/content-feature.js
```diff
@@ -424,15 +425,211 @@ export default class ContentFeature {
...
+        if (fullOptions.wrapToString) {
+            // mask toString() on class methods. `ImplClass.prototype` is non-configurable: we can't override or proxy it, so we have to wrap each method individually
+            for (const [prop, descriptor] of objectEntries(getOwnPropertyDescriptors(ImplClass.prototype))) {
+                if (prop !== 'constructor' && descriptor.writable && typeof descriptor.value === 'function') {
+                    ImplClass.prototype[prop] = new Proxy(descriptor.value, toStringProxyMixin(descriptor.value, `function ${prop}() { [native code] }`))
+                }
+            }
+
+            // wrap toString on the constructor function itself
+            Object.assign(proxyHandler, toStringProxyMixin(ImplClass, `function ${interfaceName}() { [native code] }`))
```
jonathanKingston, line 547: Here we're assign a `get` trap, I think make that clearer in the comment. (I can't think of a cleaner way to do this myself but the traps might collide is my concern).

src/content-feature.js
muodov: Yeah, for sure. I thought it was an acceptable compromise so we can split `shimProperty` from `shimInterface` while still verifying the correct call order. Calling `shimProperty` before `shimInterface` would cause very tricky bugs.

We can also merge these two methods in one, what do you think?

src/content-feature.js
muodov: good point. I'll see what I can do

package.json
```diff
@@ -54,22 +54,23 @@
   ],
   "dependencies": {
     "immutable-json-patch": "^5.1.3",
+    "parse-address": "^1.1.2",
     "seedrandom": "^3.0.5",
-    "sjcl": "^1.0.8",
-    "parse-address": "^1.1.2"
```
jonathanKingston, line 59: I think we can revert this and package-lock.json as there should be no changes now.

src/features/web-compat.js
```diff
@@ -360,16 +358,17 @@ export class WebCompat extends ContentFeature {
             if ('credentials' in navigator && 'get' in navigator.credentials) {
                 return
             }
-            // TODO: change the property descriptor shape to match the original
             const value = {
                 get () {
                     return Promise.reject(new Error())
                 }
             }
+            // TODO: original property is an accessor descriptor
```
jonathanKingston, line 366: What's the expected change here? Can you make it etc? Or file a follow up?

src/features/web-compat.js
jonathanKingston: > which we can't verify inside the method.

You can with TS no?

src/features/web-compat.js
```diff
@@ -498,29 +477,55 @@ export class WebCompat extends ContentFeature {
...
+            this.shimInterface('PresentationRequest', class {
+                // class definition is empty because there's no way to get an instance of it anyways
+            }, {
+                disallowConstructor: true,
+                allowConstructorCall: false,
+                wrapToString: true
             })
+
+            /** TODO: add shims for other classes in the Presentation API:
+             * PresentationConnection,
```
jonathanKingston, line 520: Can you raise a task to track this?

src/wrapper-utils.js
```diff
@@ -47,3 +131,300 @@ export function wrapFunction (functionValue, realTarget) {
...
+        /** @type {StrictDataDescriptor} */
+        // @ts-expect-error - As long as ImplClass is a normal class, it should have the prototype property
+        const descriptor = getOwnPropertyDescriptor(ImplClass.prototype, 'constructor')
+        if (descriptor.writable) {
+            ImplClass.prototype.constructor = Interface
+        }
+    }
+
+    // mark the class as a shimmed class
+    definePropertyFn(ImplClass, ddgShimMark, {
```
jonathanKingston, line 305: Can/should this be wrapped with an integration check?

src/features/web-compat.js
muodov: Original descriptor is an accessor with a getter and no setter. If we go all the way, we should also shim the `CredentialsContainer` class, but I'm not sure it's worth it, and I didn't want to bloat the scope here.
I will make a task.

src/content-feature.js
muodov: Yes, but it doesn't necessarily mean breakage. I do like your suggestion below to do it in test mode only though, so I'll do that.

src/wrapper-utils.js
muodov: neat idea!

package.json
muodov: Current changes only update some dev dependencies and sort entries in the package.json. This seems harmless, let's keep it?

src/features/web-compat.js
muodov: https://app.asana.com/0/1204556816597738/1207323488899175/f

jonathanKingston (approved):
👍🏻 nice one thanks for all your work here.

src/features/web-compat.js
muodov: I think this can be followed-up with the same task as below: https://app.asana.com/0/1204556816597738/1207323488899175/f

src/features/web-compat.js
muodov: Just to close the loop, I am still reluctant to do this change because I fear it could be confusing. Having to name the implementation class exactly the same as the native class is not something I'd expect. This is reversible, and we can change it later.

PR #966, opened 2024-05-15
Generate a profile id in absence of a unique profile URL

brianhall:
Asana: https://app.asana.com/0/0/1207328440043371/f
Tech Design: https://app.asana.com/0/608920331025329/1207315133262511/f

The Native Apps use an identifier (typically a profile URL) to track if a record has been removed. In some cases, records don’t have unique profile URLs, so this change generates an id based on the serialized profile JSON.

Note: this unfortunately breaks the broker json playground! 😭 Will fix that next.


src/features/broker-protection/actions/extract.js
```diff
@@ -323,26 +332,6 @@ const rules = {
...
 }
 
-/**
- * Parse a profile id from a profile URL
- * @param {string} profileUrl
- * @param {IdentifierType} identifierType
- * @param {string} identifier
- * @return {string}
- */
-export function getIdFromProfileUrl (profileUrl, identifierType, identifier) {
```
brianhall, line 333: Finally got a chance to move this to the extractor as it was bugging me here. 😄

src/features/broker-protection/actions/extract.js
```diff
@@ -395,3 +380,22 @@ function removeCommonSuffixesAndPrefixes (elementValue) {
...
+
+/**
+ * @typedef {Pick<ExtractorParams, 'identifierType'>} ProfileUrlParams
+ *
+ * @param {Record<string, any>} originalProfile
+ * @param {ProfileUrlParams=} profileUrl
+ * @param {Crypto=} cryptoLibrary
+ * @return {Promise<Record<string, any>>}
+ */
+export async function generateProfileId (originalProfile, profileUrl, cryptoLibrary) {
```
brianhall, line 392: I don’t love this. Very open to other ideas. 😄

src/features/broker-protection/actions/extract.js
```diff
@@ -41,18 +43,21 @@ import { ProfileUrlExtractor } from '../extractors/profile-url.js'
...
 
     if ('error' in extractResult) {
         return new ErrorResponse({ actionID: action.id, message: extractResult.error })
     }
 
-    const filtered = extractResult.results
+    const filteredPromises = extractResult.results
         .filter(x => x.result === true)
         .map(x => aggregateFields(x.scrapedData))
+        .map(profile => generateProfileId(profile, action?.profile?.profileUrl))
```
shakyShane, line 58: I think we can make this pattern more general - I can see the use-case, it's all about a transform that happens, but it can only be computed *after* you've aggregated all the fields etc. So, I think we can have a list of  `AsyncProfileTransform`s to apply in a loop. that way, we can co-locate more profile url/hash stuff. 

See: https://github.com/duckduckgo/content-scope-scripts/commit/0c9bdfcd0c7b48915a1fc864c21e88555bdedf9d

shakyShane (approved):

PR #967, opened 2024-05-23
add noResultsSelector to extract action

brianhall:
Asana: https://app.asana.com/0/608920331025329/1207396293353341/f

Adds support for a 'noResultsSelector' in the extract action which is checked as a fallback when the profile selector fails.


integration-test/test-pages/broker-protection/actions/results-not-found-invalid.json
```diff
@@ -0,0 +1,52 @@
...
+    "state": {
+      "action": {
+        "actionType": "extract",
+        "selector": ".search-item-invalid",
+        "noResultsSelector": ".none",
+        "retry": {
+          "environment": "web",
+          "interval": { "ms": 1000 },
+          "maxAttempts": 1
+        },
```
shakyShane, line 11: @brianhall I added this as a way to opt-out of the `extract` actions hard-coded retry logic. we should address this later if it becomes more annoying.

brianhall: Ah, didn’t occur to me to set the retry lower than the playwright timeout threshold. That’s a good workaround!

shakyShane (approved):

PR #970, opened 2024-05-31
Support clicking multiple items

brianhall:
Asana: https://app.asana.com/0/1206873150423133/1206998874293254

Occasionally we need to be able to click multiple items with a single "click" action (example below), so adding support for a "multiple" flag which loop over all the items that match a selector and attempt to execute a click.

Screenshot 2024-05-30 at 1 45 18 PM


shakyShane (approved):

integration-test/playwright/broker-protection.spec.js
```diff
@@ -381,6 +381,17 @@ test.describe('Broker Protection communications', () => {
...
+            const dbp = BrokerProtectionPage.create(page, workerInfo)
+            await dbp.enabled()
+            await dbp.navigatesTo('click-multiple.html')
+            await dbp.receivesAction('click-multiple.json')
+            const response = await dbp.waitForMessage('actionCompleted')
+
+            dbp.isSuccessMessage(response)
+            await page.waitForURL(url => url.hash === '#1-2', { timeout: 2000 })
+        })
+
```
shakyShane, line 394: @brianhall - thanks! this is a really good test as it's executing every single branch of affected code 👍🏻

src/features/broker-protection/actions/click.js
```diff
@@ -12,18 +12,25 @@ export function click (action, userData, root = document) {
...
 
-        if (!elem) {
+        if (!elements?.length) {
             return new ErrorResponse({ actionID: action.id, message: `could not find element to click with selector '${element.selector}'!` })
         }
-        if ('disabled' in elem) {
-            if (elem.disabled) {
-                return new ErrorResponse({ actionID: action.id, message: `could not click disabled element ${element.selector}'!` })
+
+        const loopLength = element.multiple && element.multiple === true ? elements.length : 1
```
shakyShane, line 21: not worth changing right now, but the double check here is redundant :)
```suggestion
        const loopLength = element.multiple === true ? elements.length : 1
```

src/features/broker-protection/actions/click.js
```diff
@@ -12,18 +12,25 @@ export function click (action, userData, root = document) {
...
+        for (let i = 0; i < loopLength; i++) {
+            const elem = elements[i]
+
+            if ('disabled' in elem) {
+                if (elem.disabled) {
+                    return new ErrorResponse({ actionID: action.id, message: `could not click disabled element ${element.selector}'!` })
+                }
+            }
+            if ('click' in elem && typeof elem.click === 'function') {
+                elem.click()
```
shakyShane, line 32: no need to change now, but if you wanted to avoid the `for` loop here you could do something like this:

```suggestion

        const subset = element.multiple === true ? elements : elements.slice(0, 1)

        for (let elem of subset) {
            if ('disabled' in elem) {
                if (elem.disabled) {
                    return new ErrorResponse({ actionID: action.id, message: `could not click disabled element ${element.selector}'!` })
                }
            }
            if ('click' in elem && typeof elem.click === 'function') {
                elem.click()
```

shakyShane (approved):

shakyShane (approved):

PR #974, opened 2024-06-10
Fix site copy/cut/paste/text selection restrictions for password/input fields

mallexxx:
Task URL: https://app.asana.com/0/42792087274227/1201373420681703/f

Steps to validate:
- Enable `copyPasteMenuSelectionOverride` in the config
- Validate paste restrictions disabled on: 
 - https://danielabar.github.io/messing_with_paste/
 - https://www.iobnet.co.in/ibanking/login.do
- Validate no alert is displayed on right click on: https://www.iobnet.co.in/ibanking/login.do
- Validate context menu is shown on the 2nd right click on:
 - https://www.iobnet.co.in/ibanking/login.do
 - https://akkords.pro/songs/agata-kristi/kak-na-vojne/
 - https://shuttermagik.com/portfolio/#
- Validate text can be selected, copied, no copyright message added:
 - https://easyinfoblog.com/technology/rip-amazon-prime-video/#6_Ways_to_rip_Amazon_Prime_Video_2020
 - https://akkords.pro/songs/agata-kristi/kak-na-vojne/
 - https://www.livelib.ru/series/1588570-vestalinneya
- Validate copy/paste of stylized text/multiple cells works on:
 - https://sheets.google.com
 - https://docs.google.com

### DIFF
--- a/src/features/copy-paste-menu-selection-override.js
+++ b/src/features/copy-paste-menu-selection-override.js
@@ -1,0 +1,215 @@
+import ContentFeature from "../content-feature.js";
+
+export default class CopyPasteMenuSelectionOverride extends ContentFeature {
+    init () {
+        try {
+            (() => {
+                var lastContextMenuEventX = -1;
+                var lastContextMenuEventY = -1;
+                var lastContextMenuEventWasPrevented = false;
+                var isSelectionColorOverrideStyleAdded = false;
+
+                // - Suppress alerts displayed on right click
+                document.addEventListener('DOMContentLoaded', function () {
+                    // add alert suppression script to page content world.
+                    // use invisible `_ddg-suppress-alert-flag` element to pass `shouldSuppressAlert` value
+                    // from client content world script handling `contextmenu` events.
+                    var alertSuppressionScript = document.createElement('script');
+                    alertSuppressionScript.textContent = `
+                    (() => {
+                        const flagElement = document.createElement('_ddg-suppress-alert-flag');
+                        flagElement.style.display = 'none';
+                        flagElement.setAttribute('value', 'false');
+                        document.head.appendChild(flagElement);
+
+                        function shouldSuppressAlert() {
+                            return flagElement.getAttribute('value') === 'true';
+                        }
+
+                        const originalWindowAlert = window.alert;
+                        // prevent websites displaying alerts on right click
+                        window.alert = function(msg) {
+                            if (shouldSuppressAlert()) {
+                                console.log("suppressed alert on contextmenu: " + msg);
+                            } else {
+                                originalWindowAlert(msg);
+                            }
+                        };
+                    })();
+                    `;
+                    document.head.appendChild(alertSuppressionScript);
+                });
+
+                // - Disable context menu prevention if user right-clicks at the same point for a second time
+                document.addEventListener("contextmenu", function(e) {
+                    if (lastContextMenuEventWasPrevented && e.clientX == lastContextMenuEventX && e.clientY == lastContextMenuEventY) {
+                        // second same point click: disable next `contextmenu` event handlers
+                        e.stopImmediatePropagation();
+                        lastContextMenuEventWasPrevented = false;
+                        return true;
+                    }
+                    // remember click position
+                    lastContextMenuEventX = e.clientX;
+                    lastContextMenuEventY = e.clientY;
+                    // prevent websites displaying alerts on right click
+                    const suppressAlertFlagElement = document.getElementsByTagName("_ddg-suppress-alert-flag")[0];
+                    suppressAlertFlagElement.setAttribute('value', 'true');
+                    // check if the context menu event handling was prevented
+                    setTimeout(function() {
+                        lastContextMenuEventWasPrevented = e.defaultPrevented;
+                        // stop alerts suppression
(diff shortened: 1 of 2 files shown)


src/features/copy-paste-menu-selection-override.js
```diff
@@ -0,0 +1,215 @@
...
+                var isSelectionColorOverrideStyleAdded = false;
+
+                // - Suppress alerts displayed on right click
+                document.addEventListener('DOMContentLoaded', function () {
+                    // add alert suppression script to page content world.
+                    // use invisible `_ddg-suppress-alert-flag` element to pass `shouldSuppressAlert` value
+                    // from client content world script handling `contextmenu` events.
+                    var alertSuppressionScript = document.createElement('script');
+                    alertSuppressionScript.textContent = `
+                    (() => {
```
jonathanKingston, line 19: I don't think this needs to be an injected script and instead just a method.

src/features/copy-paste-menu-selection-override.js
```diff
@@ -0,0 +1,215 @@
...
+                    // enumerate all text nodes in selection
+                    forEachTextNodeInRange(range, ((text) => {
+                        const node = text.parentNode;
+                        if (node.classList.contains('__ddg-override-selection-background-color')) {
+                            return true; // selection color is already overriden – proceed next
+                        }
+
+                        const selectionColor = window.getComputedStyle(node, '::selection').backgroundColor;
+                        // if text node selection color is set to `transparent` (0,0,0,0) or is not set explicitly - set it to default (`highlight`)
+                        if (selectionColor === 'rgba(0, 0, 0, 0)' || selectionColor === 'transparent') {
```
jonathanKingston, line 199: There's a bunch of other values here that are could cause it not to show. Like visibility:hidden, also does #00000000 appear in computed styles or is it translated to rgba?

src/features/copy-paste-menu-selection-override.js
```diff
@@ -0,0 +1,215 @@
...
+
+                    // use default (highlight) selection color
+                    const cssRules = `
+                        *.__ddg-override-selection-background-color:not(input):not(textarea)::selection {
+                            background-color: highlight !important;
+                        }
+                    `;
+                    styleElement.textContent = cssRules;
+
+                    document.head.appendChild(styleElement);
```
jonathanKingston, line 144: It would be better to add these styles inline rather than injecting DOM nodes.

src/features/copy-paste-menu-selection-override.js
```diff
@@ -0,0 +1,215 @@
...
+
+                        function shouldSuppressAlert() {
+                            return flagElement.getAttribute('value') === 'true';
+                        }
+
+                        const originalWindowAlert = window.alert;
+                        // prevent websites displaying alerts on right click
+                        window.alert = function(msg) {
+                            if (shouldSuppressAlert()) {
+                                console.log("suppressed alert on contextmenu: " + msg);
```
jonathanKingston, line 33: Remove the console log.

src/features/copy-paste-menu-selection-override.js
jonathanKingston: It seems like the default in Chrome is `rgba(0, 0, 0, 0)`, I'm not sure how the style applies but we should probably not apply in default case.

src/features/copy-paste-menu-selection-override.js
mallexxx: the reason behind this is the `ContentScopeScript` being injected as a `defaultClient` scope script and the `alert` method is called by the document from the `page` scope, so they don‘t intersect, this hack is here to pass `shouldSuppressAlert` flag between the `page` and `defaultClient` scope and handle the `alert` call in the `page` scope. I‘m open for suggestions on how to handle this in "the right way".

src/features/copy-paste-menu-selection-override.js
mallexxx: yes, the default is detected `rgba(0, 0, 0, 0)` (i.e. not explicitly defined), but may also be used from the client side to hide selection.

>  but we should probably not apply in default case

I agree, but there seems no way to distinguish the "not defined" case and an explicit `rgba(0, 0, 0, 0)`. The overridden style sets selection background color to `highlight` which is a default browser selection color, so in case it‘s not explicitly defined – the selection colors remains the same.

src/features/copy-paste-menu-selection-override.js
```diff
@@ -0,0 +1,215 @@
...
+                        // second same point click: disable next `contextmenu` event handlers
+                        e.stopImmediatePropagation();
+                        lastContextMenuEventWasPrevented = false;
+                        return true;
+                    }
+                    // remember click position
+                    lastContextMenuEventX = e.clientX;
+                    lastContextMenuEventY = e.clientY;
+                    // prevent websites displaying alerts on right click
+                    const suppressAlertFlagElement = document.getElementsByTagName("_ddg-suppress-alert-flag")[0];
```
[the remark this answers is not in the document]
mallexxx: to pass value between `page` and `defaultClient` scopes, see the comment above.

shakyShane:
Closing as it's been open a year - if there's any desire please re-open as needed :)

PR #981, opened 2024-06-25
fix _Package.swift -> Package.swift

shakyShane:
Renamed by mistake - fixing quickly now and will follow with a test to prevent this.


brindy (approved):
LGTM - thanks

PR #995, opened 2024-07-09
Duck Player overlay copy experiment

mgurgel:
Asana: https://app.asana.com/0/0/1207588185313774/f
CC: @shakyShane


integration-test/playwright/page-objects/duckplayer-overlays.js
```diff
@@ -33,6 +33,22 @@ const userValues = {
...
+    /** @type {import("../../../src/features/duck-player.js").UISettings} */
+    'overlay copy a1': {
+        overlayCopy: 'a1'
+    },
+    /** @type {import("../../../src/features/duck-player.js").UISettings} */
+    'overlay copy b1': {
+        overlayCopy: 'b1'
+    }
+}
+
```
shakyShane, line 51: You can annotate the object to be a `Record<string, import("../../../src/features/duck-player.js").UISettings>` instead to make this simpler

src/features/duckplayer/text.js
```diff
@@ -8,15 +10,30 @@ const text = {
...
+    },
+    videoOverlayTitle_b1: {
+        title: 'Drowning in ads in YouTube? {newline} Turn on Duck Player.'
+    },
     videoOverlaySubtitle: {
         title: 'provides a clean viewing experience without personalized ads and prevents viewing activity from influencing your YouTube recommendations.'
     },
+    videoOverlaySubtitle_a1_b1: {
+        title: 'What you watch in DuckDuckGo won’t influence your recommendations on YouTube.'
+    },
```
shakyShane, line 24: is this key `videoOverlaySubtitle_a1_b1` created to indicate that this string is used in 2 places in the app? if so I don't think we should allow experiment details leak into this file. This file is just a bunch of strings that all need unique identifiers - any mappings can be done in your code.

src/features/duckplayer/overlay-messages.js
```diff
@@ -12,19 +12,34 @@ import * as constants from './constants.js'
...
+            return Promise.resolve({
+                userValues: {
+                    overlayInteracted: false,
+                    privatePlayerMode: { alwaysAsk: {} }
+                },
+                ui: {
+                    overlayCopy
+                }
+            })
+        }
```
shakyShane, line 42: We need to avoid accessing the window url like this here. There's an example over here in the Environment class https://github.com/duckduckgo/content-scope-scripts/blob/main/src/features/duckplayer/overlays.js#L140-L160

the idea is that we use the `if (this.debug)` flag to guard against anything running in production. Also we keep all access to things like URLs/search params here too.

I'd prefer if the check here was purely in the Environment class, and it can be a combination of checks like injectName, debug etc.

Remember this code is injected into web pages, so we have different constraints/preferences here.

shakyShane (changes requested):
@mgurgel - looking good! 

There are a couple of comments to look through, let me know if you have any questions :)

integration-test/playwright/page-objects/duckplayer-overlays.js
mgurgel: Yeah I tried it at first but then autocomplete didn’t work as well for some reason. Will try it again and see.

mgurgel: Maybe I’m doing something wrong, but when I use `Record<string…` then set `@param {keyof uiSettings} uiSetting` in a function further down, it doesn’t pick up on the keys and defers to the type string in the Record instead. Screenshots below

Using individual type declarations 
Screenshot 2024-07-10 at 15 09 00

Using the top level Record<> declaration
Screenshot 2024-07-10 at 15 10 13

The main reason I went with individual type declarations is that I noticed you used the same pattern just a few lines above, in `userValues` and I figured it was for the same reason (to get around this keyof limitation)

mgurgel: The following screenshots summarise my problem a bit better

image

image

src/features/duckplayer/text.js
mgurgel: Makes sense. I’ve updated the translation string ids

src/features/duckplayer/overlay-messages.js
mgurgel: I’ve moved that check into the Environment class and changed DuckPlayerOverlayMessages to accept an instance of Environment in the constructor

shakyShane (approved):

PR #999, opened 2024-07-24
retry initial message in release notes (webkit)

shakyShane:
https://app.asana.com/0/0/1207888506630401/f

This PR just makes sure that the initialSetup message in release-notes has a retry mechanism to allow for any delay in webkit handlers becoming available.


mgurgel (approved):
👍

PR #1006, opened 2024-08-01
Breakage reporting feature

nshuba:
See [tech design](https://app.asana.com/0/72649045549333/1207814627649677/f)

For testing, see https://github.com/duckduckgo/windows-browser/pull/2067

### DIFF
--- a/src/features/breakage-reporting.js
+++ b/src/features/breakage-reporting.js
@@ -1,15 +1,14 @@
 import ContentFeature from '../content-feature'
+import { getJsPerformanceMetrics } from './breakage-reporting/utils.js'
 
 export default class BreakageReporting extends ContentFeature {
     init () {
         this.messaging.subscribe('getBreakageReportValues', () => {
-            const paintResources = performance.getEntriesByType('paint')
-            const firstPaint = paintResources.find((entry) => entry.name === 'first-contentful-paint')
-            const firstPaintStart = firstPaint ? firstPaint.startTime : -1.0
-
+            const jsPerformance = getJsPerformanceMetrics()
             const referrer = document.referrer
+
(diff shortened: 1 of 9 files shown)


src/features/breakage-reporting.js
```diff
@@ -0,0 +1,17 @@
+import ContentFeature from '../content-feature'
+
+export default class BreakageReporting extends ContentFeature {
+    init () {
+        this.messaging.subscribe('getBreakageReportValues', () => {
+            const paintResources = performance.getEntriesByType('paint')
+            const firstPaint = paintResources.find((entry) => entry.name === 'first-contentful-paint')
+            const firstPaintStart = firstPaint ? firstPaint.startTime : -1.0
```
jonathanKingston, line 8: 1. Can we move this code into a shared function? (probably into `features/performance/vitals.js` or similar subdir)
2. Why -1? The other just reports a blank array, I'm not fussy just interested if the difference is needed etc.

nshuba: Thanks for taking an early look!
1. I had it that way to start, but then realized we may not want to use what we have in the `PerformanceMetrics` feature as-is. Mostly because `PerformanceMetrics` returns an array, whereas here I'd like to keep optionality in how we will package the performance results. I think it will be cleaner to return these values separately in an object as opposed to packaging multiple values into an array. See also [this discussion](https://app.asana.com/0/0/1207542264574558/1207954703409947/f).
2. Same as point (1) above

nshuba: Ignore the above, [the conversion](https://app.asana.com/0/0/1207542264574558/f) has evolved 😅 Addressed both points and ready for review!

jonathanKingston (approved):

PR #1009, opened 2024-08-05
Special Error Page on Preact

mgurgel:
https://app.asana.com/0/1206594217596623/1207950300631193/f

- Migrates Special Error page to Preact
- Implements iOS SSL error page

See [PR task](https://app.asana.com/0/1206594217596623/1208035400782761/f) for notes


packages/special-pages/pages/special-error/app/components/AdvancedInfo.jsx
```diff
@@ -53,11 +52,8 @@ export function AdvancedInfoContent() {
 }
 
 export function AdvancedInfo() {
-    /** @type {import("preact/hooks").MutableRef<HTMLSpanElement|null>} */
-    const visitSiteRef = useRef(null)
-
     const animationDidEnd = () => {
-        visitSiteRef.current?.scroll()
+        window.dispatchEvent(new CustomEvent('advanced-info-animation-end'))
```
shakyShane, line 59: if you don't send any values, this can be just `new Event`

packages/special-pages/shared/components/Text/Text.js
```diff
@@ -7,11 +7,22 @@ import styles from './Text.module.css'
...
+ * @param {boolean} [props.strictSpacing] - Apply Design System letter spacing. Default: true
  * @param {import("preact").ComponentChild} [props.children]
+ * @param {import("preact").JSX.MouseEventHandler<EventTarget>} [props.onClick]
  */
-export function Text ({ as: Comp = 'p', variant, className, children }) {
+export function Text ({ as: Comp = 'p', variant, strictSpacing = true, className, children, onClick }) {
     return (
-        <Comp className={classNames({ [styles[`${variant}`]]: variant }, className)}>
+        <Comp className={classNames({ [styles[`${variant}`]]: variant, [styles.strictSpacing]: strictSpacing }, className)}
+            onClick={
```
shakyShane, line 17: This is adding an event handler for every use of the Text component. since onClick will only be given to <a> elements, I wouldn't use a generic <Text> for <a> elements. The dance you're having to do with the event handler indicates that this is trying to do too much. 

Why not just have a <Link> element in this same file, and that can re-use styles, but be explicit about handlers.

shakyShane (changes requested):
Thanks this is looking really good. I have requested a small change to the <Text> component, I think we can/should avoid going too generic yet.

PR #1010, opened 2024-08-05
reload instead of re-throwing

shakyShane:
https://app.asana.com/0/0/1207978041075737/f

We're seeing some further delays where where the webkit handlers take some tie to appear. Instead of increasing the retries further (it's already a 10x retry at 300ms intervals), we should just try reloading the page instead.

This has a better chance of resolving the issue - multiple internal testers have reported that reloading fixed the issue. 

We should try it first with onboarding - and if it's stable + fixes bugs, we should use this everywhere.


shakyShane:
* **#1010** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/1010?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

PR #1012, opened 2024-08-09
naming: use injectName vs platform in the correct places

shakyShane:
We were incorrectly using 'platform' in places where we meant the 'injectName'

- `platform.name` throughout all of CSS is something that can be set at runtime, to values like `ios`, `macos`, `android` etc.
- `import.meta.injectName` is the build target and has higher-level values like `windows`, `apple` etc. 

This PR is just being more precise with naming to avoid confusion.

### DIFF
--- a/packages/special-pages/pages/onboarding/app/components/Switch.js
+++ b/packages/special-pages/pages/onboarding/app/components/Switch.js
@@ -16,7 +16,7 @@ import { useEnv } from '../../../../shared/components/EnvironmentProvider'
 export function Switch ({ checked = false, variant, ...props }) {
     const { onChecked, onUnchecked, ariaLabel, pending } = props
     const env = useEnv()
-    const platform = variant || env.platform
+    const platform = variant || env.injectName
     function change (e) {
         if (e.target.checked === true) {
             onChecked()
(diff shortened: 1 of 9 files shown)


shakyShane:
* **#1012** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/1012?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/special-pages/pages/onboarding/app/components/Switch.js
```diff
@@ -16,7 +16,7 @@ import { useEnv } from '../../../../shared/components/EnvironmentProvider'
 export function Switch ({ checked = false, variant, ...props }) {
     const { onChecked, onUnchecked, ariaLabel, pending } = props
     const env = useEnv()
-    const platform = variant || env.platform
+    const platform = variant || env.injectName
```
mgurgel, line 19: I know this is just a local var, but is it worth changing it to injectName too to further avoid confusion?

mgurgel:
Are we keeping `document.documentElement.dataset.platform = import.meta.injectName` on inline.js as is?

mgurgel (approved):
Just two non-blocking questions above. Looks good otherwise 👍

PR #1021, opened 2024-08-30
Generate a random street address for data broker opt out forms

brianhall:
Asana: https://app.asana.com/0/0/1208157142708080/f

### DIFF
--- a/src/features/broker-protection/actions/fill-form.js
+++ b/src/features/broker-protection/actions/fill-form.js
@@ -1,5 +1,6 @@
 import { getElement, generateRandomInt } from '../utils.js'
 import { ErrorResponse, SuccessResponse } from '../types.js'
+import { generatePhoneNumber, generateZipCode, generateStreetAddress } from './generators.js'
 
 /**
  * @param {Record<string, any>} action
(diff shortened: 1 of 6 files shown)


src/features/broker-protection/actions/fill-form.js
```diff
@@ -177,22 +180,3 @@ function setImageUpload (element) {
         return { result: false, error: e.toString() }
     }
 }
-
-export function generatePhoneNumber () {
```
brianhall, line 181: Moved to `./generators.js`

unit-test/broker-protection.js
```diff
@@ -556,25 +556,32 @@ describe('Actions', () => {
...
+            expect(typeof zipCode).toEqual('string')
+            expect(zipCode.length).toBe(5)
+            expect(zipCode).toMatch(/^\d{5}$/)
+        })
+    })
+    describe('generateStreetAddress', () => {
+        it('generates a string of integers of an appropriate size', () => {
+            const streetAddress = generateStreetAddress()
+            expect(typeof streetAddress).toEqual('string')
+            expect(streetAddress).toMatch(/^\d+ [A-Za-z]+(?: [A-Za-z]+)?$/)
```
shakyShane, line 584: Can we either call this many times in a loop, or put it through fast-check? the fast-check test should look almost identical since there are no inputs - but it would be good to see these running more than once in tests.

shakyShane (changes requested):

unit-test/broker-protection.js
brianhall: Yep, fair point. Will do.

shakyShane: Actually, in this instance, for 'generateStreetAddress' fast-check won't be any better than a loop

brianhall: Added in [https://github.com/duckduckgo/content-scope-scripts/pull/1021/commits/c2a9e1de9f8fbf42d4c8dfa7436e3b9d4e6515dc](c2a9e1de9f8fbf42d4c8dfa7436e3b9d4e6515dc)

shakyShane (approved):

PR #1025, opened 2024-09-05
Adopt Stylelint in CI

mgurgel:
- Adds Stylelint scripts to package.json
- Adds stylelint step to test workflow
- Initial stylelint config with multiple disabled rules. I suggest we gradually adopt these rules by removing each one from the config, running `npm run stylelint-fix` and reviewing/committing the resulting changes.


jonathanKingston (approved):

shakyShane:
Just rebasing this and then we can merge

PR #1060, opened 2024-09-16
Make variant a1 the new default for duck player desktop overlay

mgurgel:
https://app.asana.com/0/1207252092703676/1207352313241844/f

Adopts variant a1 from the copy experiment as the default for Duck Player overlays

### DIFF
--- a/src/features/duckplayer/text.js
+++ b/src/features/duckplayer/text.js
@@ -80,25 +80,11 @@ export function nl2br (text) {
  */
 
 /**
- *  @type {Record<import('../duck-player').UISettings['overlayCopy'], OverlayCopyTranslation>}
+ *  @type {Record<'default', OverlayCopyTranslation>}
  */
 export const overlayCopyVariants = {
     default: {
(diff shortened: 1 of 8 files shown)


src/features/duckplayer/text.js
```diff
@@ -80,29 +80,15 @@ export function nl2br (text) {
  */
 
 /**
- *  @type {Record<import('../duck-player').UISettings['overlayCopy'], OverlayCopyTranslation>}
+ *  @type {Record<'default', OverlayCopyTranslation>}
```
mgurgel, line 83: I didn’t want to roll back the experiment logic completely, as to leave the door open for future experiments.

mgurgel: @shakyShane Plus, string ids such as `videoOverlayTitle2` are already being used on the mobile overlay, so I figured it best not to touch that part of the code just yet. But happy to refactor those names if you think it’s the right time to do that.

shakyShane (approved):

PR #1066, opened 2024-09-19
Remove runtime-checks

jonathanKingston:
Task: https://app.asana.com/0/1201614831475344/1208341763117531/f

Removes the runtime-checks feature and all it's related features.

PR #1089, opened 2024-10-03
move remaining tests to playwright

shakyShane:
https://app.asana.com/0/1201614831475344/1206978084282840/f

I've done this on top of removing runtimeChecks https://github.com/duckduckgo/content-scope-scripts/pull/1066


shakyShane:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/1089?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1089** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/1089?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#1066** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/1066?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

integration-test/helpers/harness.js
```diff
@@ -1,150 +1,133 @@
...
+ * arguments to content-scope's init methods + waits for a known
+ * indicators to avoid race conditions
+ *
+ * @param {import("@playwright/test").Page} page
+ * @param {string} urlString
+ * @param {Record<string, any>} [args]
+ * @param {string|null} [evalBeforeInit]
+ * @returns {Promise<void>}
+ */
+export async function gotoAndWait (page, urlString, args = {}, evalBeforeInit = null) {
```
shakyShane, line 97: this hasn't changed, just moved out of the closure it was in before

integration-test/test-fingerprint.js
```diff
@@ -17,170 +19,118 @@ const expectedFingerprintValues = {
...
+        })
     })
 
-    tests.forEach(test => {
-        it('Fingerprints should not match across first parties', async () => {
-            const page = await browser.newPage(pagePath)
+    test.describe.serial('Verify injected script is not visible to the page', () => {
+        tests.forEach(testCase => {
+            test(`Fingerprints should not match across first parties ${testCase.url}`, async ({ page, altServerPort }) => {
+                console.log('running:', altServerPort)
```
shakyShane, line 118: these comments are to keep typescript/eslint happy

scripts/server.mjs
```diff
@@ -0,0 +1,10 @@
+import { createServer } from "http-server";
+
+const server = createServer({ root: process.env.SERVER_DIR });
+server.listen(process.env.SERVER_PORT);
+server.server.on('listening', () => {
+    process.send?.({port: server.server.address().port});
+})
+server.server.on('error', () => {
+    process.exit(1)
+})
```
shakyShane, line 10: this script just allows us to programmatically create a server on the fly (using the same library we already use) - this is used in the fingerprinting tests to start the second server

integration-test/helpers/harness.js
```diff
@@ -1,150 +1,133 @@
...
+            const launchOptions = {
+                devtools: true,
+                headless: false,
+                viewport: {
+                    width: 1920,
+                    height: 1080
+                },
+                args: [
+                    '--disable-extensions-except=integration-test/extension',
+                    '--load-extension=integration-test/extension'
```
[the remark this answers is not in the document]
shakyShane: Only tests that require the extension will use this helper `testContext` - I think I should rename it and that might avoid the confusion :)

integration-test/test-pages.js
```diff
@@ -66,20 +53,22 @@ describe('Test integration pages', () => {
         )
         for (const key in pageResults) {
             for (const result of pageResults[key]) {
-                expect(result.result).withContext(key + ':\n ' + result.name).toEqual(result.expected)
+                expect(result.result).toEqual(result.expected)
```
jonathanKingston, line 56: Can we not keep this context? It was useful for seeing the individual test failures.

integration-test/test-fingerprint.js
jonathanKingston: We could just drop the port usage in the definition instead?

integration-test/test-cookie.js
```diff
@@ -53,11 +41,11 @@ describe('Cookie protection tests', () => {
...
     })
 
-    it('Erroneous values do not throw', async () => {
-        const page = await browser.newPage()
-        await page.goto('http://localhost:8080/index.html')
-
+    test('Erroneous values do not throw', async ({ page }) => {
+        await gotoAndWait(page, '/index.html')
         const result = await page.evaluate(async () => {
+            document.cookie = 'a=b; expires=Wed, 21 Aug 2040 20:00:00 UTC;'
```
jonathanKingston, line 47: Is this needed here? There's a test above for it

integration-test/test-pages.js
jonathanKingston: Nice:

```
await test.step(`${key}:\n ${result.name}`, async () => {
    expect(result.result).toEqual(result.expected);
});
```

integration-test/test-fingerprint.js
```diff
@@ -17,170 +19,118 @@ const expectedFingerprintValues = {
...
 const tests = [
-    { url: `localhost:8080${pagePath}` },
-    { url: `127.0.0.1:8383${pagePath}` }
+    { url: `http://localhost:3220${pagePath}` },
+    { url: `http://127.0.0.1:8383${pagePath}` }
 ]
 
 function testFPValues (values) {
     for (const [name, prop] of Object.entries(values)) {
-        expect(prop).withContext(`${name}`).toEqual(expectedFingerprintValues[name])
```
jonathanKingston, line 26: Add `test.step` here also :)

integration-test/test-cookie.js
shakyShane: yeah it's needed because this test looks for 'a'. it worked before because the harness setup was leaking state across tests.

integration-test/test-fingerprint.js
shakyShane: no because the `altServerPort` variable is only available inside the test. Could just eslint-disable this warning if we prefer.

jonathanKingston: Maybe I'm misunderstanding...

But we're destructuring a property? Are you saying without this we'd not trigger the altServerPort extension?
I'll not block on this if that's the case but it's a weird side effect, can it not be a method that's triggered before the FP tests?

shakyShane: I'm trying to follow docs/best-practices so that we can point people to official sources where it makes sense. In that context, I used https://playwright.dev/docs/test-fixtures to create & tear down the server on demand, per test. The docs there give the [reasons for it](https://playwright.dev/docs/test-fixtures#with-fixtures).

Actually this is exactly how `page` works too, so technically whilst it's a little odd for us (coming from puppeteer) it's the 'blessed' way I believe.

Having said that, I also understand the weirdness. In this case we just need the argument to be there for the side effects to run and be cleaned up, but because the test URL is hard-coded above in this file, we don't actually end up using it 😂 

This was all in the idea of not changing too much, maybe we should?

jonathanKingston: Makes sense, it's a bit weird but I can live with it.

Maybe make a task to discuss?

- as server.mjs isn't reused elsewhere so we have a bunch of code remapping code into altServerPort methods and could be unified.
- In that work we could refactor to a method that accepts a port/filepath as you say.

scripts/inject.js
```diff
@@ -42,7 +42,6 @@ const builds = {
         output: [
             'build/integration/contentScope.js',
             'integration-test/extension/contentScope.js',
-            'integration-test/pages/build/contentScope.js',
```
jonathanKingston, line 45: Ah did we do this process twice? The rationale for the separation for pages/ and test-pages/ was so that test-pages explicitly are shared into privacy-test pages where as the former are little helpers kinda for these legacy tests. (where possible the test-pages pattern should be used in my mind).

I don't mind with blending the two directories, so long as you're aware this directory is also going to be served as a web page: https://privacy-test-pages.site/content-scope-scripts/ (we could make the index.html a directory file for example and it be helpful anyway)

jonathanKingston (approved):

jonathanKingston:
Cheers for doing this, remember to raise the few tasks we discussed:
- moving pupeteer dir
- consider simplifying altServerPort + server.mjs / make function task.

PR #1106, opened 2024-10-14
[WIP] W3 validator for CSS

mgurgel:
Validates CSS via the W3C/NU validation service at build time


jonathanKingston:
I think we agreed not to do this direction, reopen if that's not the case.

PR #1144, opened 2024-10-19
Support conditional actions with expectations

brianhall:
**Asana Task/Github Issue:** https://app.asana.com/0/481882893211075/1208563922804591/f 

## Testing Steps

-  Run test suite and drag/drop into XCode

## Checklist

*Please tick all that apply:*


injected/src/features/broker-protection/actions/expectation.js
```diff
@@ -1,23 +1,43 @@
...
+            runActions = false
+
+            if (!x.failSilently) {
+                errors.push('error' in x ? x.error : 'unknown error')
+            }
 
-    const errors = results.filter(x => x.result === false).map(x => {
-        if ('error' in x) return x.error
-        return 'unknown error'
+            delete x.failSilently
```
shakyShane, line 27: we can avoid passing this around, see https://github.com/duckduckgo/content-scope-scripts/commit/3676158835a8325108db37e81933d0fff462b1fa for a possible approach - the idea is just to reach back to the original array using the index. It will prevent needing to change anything else in this file

injected/src/features/broker-protection/actions/expectation.js
```diff
@@ -1,23 +1,43 @@
...
+    /**
+     * @type {Array<import('../types.js').BooleanResult & { failSilently?: boolean }>}
+     */
     const results = expectMany(action.expectations, root)
+    const errors = []
+    let runActions = true
+
+    results.forEach((x) => {
+        if (x.result === false) {
+            runActions = false
```
shakyShane, line 21: this ends up being 'if any result is false, dont run' - that can be represented as this instead - to prevent needing these inner/outer values.

```js
const runActions = results.every(x => x.result === true);
```

injected/src/features/broker-protection/actions/expectation.js
```diff
@@ -1,23 +1,43 @@
...
+        }
     })
 
     if (errors.length > 0) {
         return new ErrorResponse({ actionID: action.id, message: errors.join(', ') })
     }
 
+    if (action.actions?.length && runActions) {
+        action.actions.forEach((expectationAction) => {
+            execute(expectationAction, userData, root)
```
shakyShane, line 37: execute is async, and it can fail.

I believe your intention here is just allow the more simple actions, like 'clicks' to run afterwards. Still we need to call these sequentially. you might also want to catch errors for observibility? see my commit https://github.com/duckduckgo/content-scope-scripts/commit/3676158835a8325108db37e81933d0fff462b1fa#diff-588b8450ad6aff7276f65312571cd134fa5afb0cd4fc4066d106bd74a47891f7R33-R41 here - I left a TODO to indicate where you might need to think about the errors?

injected/src/features/broker-protection/actions/expectation.js
```diff
@@ -1,23 +1,43 @@
...
 
 /**
  * @param {Record<string, any>} action
- * @param {Document | HTMLElement} root
+ * @param {Record<string, any>} userData
+ * @param {Document} root
  * @return {import('../types.js').ActionResponse}
  */
-export function expectation (action, root = document) {
+export function expectation (action, userData, root = document) {
```
shakyShane, line 11: because `execute` is async, we'll need this to be async now - an easy fix

shakyShane (changes requested):
2 miain things I noticed:

1 - we can prevent changing the signatures of the exiting code (see the linked commit)
2 - if you want to execute actions from within another - you need to consider that it's an async operation and capture/communicate those errors for observability. If you already validated that errors are propagated correctly, at the very least we should await each execution to prevent dangling promises. 

this commit has a couple of ideas, happy to talk if you need anything https://github.com/duckduckgo/content-scope-scripts/commit/3676158835a8325108db37e81933d0fff462b1fa#diff-588b8450ad6aff7276f65312571cd134fa5afb0cd4fc4066d106bd74a47891f7R33-R41

injected/src/features/broker-protection/actions/expectation.js
```diff
@@ -1,21 +1,46 @@
...
+            if (x.result === true) return false
+            if (action.expectations[index].failSilently) return false
+            return true
+        }).map((x) => {
+            return 'error' in x ? x.error : 'unknown error'
+        })
 
     if (errors.length > 0) {
-        return new ErrorResponse({ actionID: action.id, message: errors.join(', ') })
+        return new ErrorResponse({ actionID: action.id, message: 'bees ' + errors.join(', ') + ' error length ' + errors.length })
```
shakyShane, line 25: is the text here a mistake?

shakyShane (changes requested):
Can you check the error text :)

injected/src/features/broker-protection/actions/expectation.js
brianhall: Ugh, thanks. 🤦 Good catch. Fixed.

shakyShane (approved):

PR #1157, opened 2024-10-22
[Google password manager] Android isolated module for password import 

dbajpeyi:
**Asana Task/Github Issue:** 
Tech Design (isolated module for passwordimport in android): https://app.asana.com/0/1208273769335188/1208330161079810/f

JS tasks: https://app.asana.com/0/1208273769335188/1208330161079810/f
Main task: https://app.asana.com/0/1208273769335188/1207413838874779/f

## Description
We are introducing a new google password import flow in Android. For that, the android platform loads this isolated module, and injects it using `addDocumentStartJavascript` in a special tab (so this JS is only injected in that isolated tab).

The logic looks for some elements with exponential backoff and animates them.

## Testing Steps
1. On Android, checkout `feature/craig/autofill_gpm_ui_to_launch_from_password_manager`
2. Make sure webview version is updated to latest,
3. Install C-S-S with `npm i github:duckduckgo/content-scope-scripts#2510d3b88fe0620833ebf7e53da85351078c5527`
4. Build DDG using `./gradlew clean && installID`
5. Go to settings -> autofill dev settings -> launch google password manager
6. Go through the flow, you should see the "Sign in" button, the settings cogwheel, and the final password export button animated with blue background.

## Checklist

*Please tick all that apply:*


dbajpeyi:
PS: It's not clear to me why is this windows test breaking 🤔

injected/src/features/password-import.js
```diff
@@ -0,0 +1,233 @@
...
+                        scale: 1.01,
+                        backgroundColor: 'rgba(0, 39, 142, 0.5)'
+                    },
+                    element,
+                    shouldTap: this.#exportButtonSettings.shouldAutotap ?? false
+                }
+                : null
+        } else if (path === '/intro') {
+            const element = await this.findSignInButton()
+            return element != null
```
jonathanKingston, line 51: Do you mind if we use if statements instead of multi-line ternaries with complex objects.

injected/src/features/password-import.js
```diff
@@ -0,0 +1,233 @@
...
+     * @param {any} settings
+     */
+    setButtonSettings (settings) {
+        this.#exportButtonSettings = settings?.exportButton
+        this.#settingsButtonSettings = settings?.settingsButton
+        this.#signInButtonSettings = settings?.signInButton
+    }
+
+    init (args) {
+        this.setButtonSettings(args?.featureSettings?.passwordImport || {})
```
jonathanKingston, line 213: There should be a getter for settings to use instead of this argument. We're trying to deprecate this args usuage.

injected/src/features/password-import.js
```diff
@@ -0,0 +1,233 @@
...
+        return this.#settingsButtonSettings?.labelTexts
+            .map(text => `a[aria-label="${text}"]`)
+            .join(',')
+    }
+
+    /**
+     * @returns {string}
+     */
+    get settingsButtonSelector () {
+        return `${this.#settingsButtonSettings?.selectors?.join(',')}, ${this.settingsLabelTextSelector}`
```
jonathanKingston, line 200: Should we have defaults in case they're not specified in the config?

injected/src/utils.js
```diff
@@ -744,3 +744,35 @@ export function legacySendMessage (messageType, options) {
     return originalWindowDispatchEvent && originalWindowDispatchEvent(createCustomEvent('sendMessageProxy' + messageSecret, { detail: { messageType, options } }))
     // TBD other platforms
 }
+
+/**
+ * Takes a function that returns an element and tries to find it with exponential backoff.
+ * @param {number} delay
```
jonathanKingston, line 750: Can you specify the other params here?

injected/src/features/password-import.js
dbajpeyi: Wondering if it'd be not too bad if the object was simplified. Trying to avoid too much nesting here.

injected/src/features/password-import.js
dbajpeyi: We were just talking about this to craig that the android app could have a default object, for some reason if remote config is not available. 

On the JS side we could also fallback to something, not sure if it's necessary 🤔

injected/src/features/password-import.js
dbajpeyi: Done!

injected/src/utils.js
dbajpeyi: Done!

injected/integration-test/autofill-password-import.spec.js
```diff
@@ -0,0 +1,116 @@
...
+    await passwordImportFeature.enabled()
+    await passwordImportFeature.navigate()
+    await passwordImportFeature.clickOnElement('Home page')
+    await passwordImportFeature.assertElementIsAnimating('a[aria-label="Password options"')
+
+    await passwordImportFeature.clickOnElement('Signin page')
+    await passwordImportFeature.assertElementIsAnimating('a[aria-label="Sign in"')
+
+    await passwordImportFeature.clickOnElement('Export page')
+    await passwordImportFeature.assertElementIsAnimating('button[aria-label="Export"')
```
jonathanKingston, line 20: Will this ordering not be flaky?
Would it be more reliable to have a `waitForAnimate` style function, something like:
```JavaScript
const didAnimateExport = passwordImportFeature.waitForAnimation('button[aria-label="Export"', timeout)
await passwordImportFeature.clickOnElement('Export page')
await didAnimateExport
```

Where `waitForAnimation` could wait for an `animationstart` event.

injected/src/features/autofill-password-import.js
```diff
@@ -0,0 +1,251 @@
...
+        this.#exportButtonSettings = this.getFeatureSetting('exportButton')
+        this.#signInButtonSettings = this.getFeatureSetting('signInButton')
+        this.#settingsButtonSettings = this.getFeatureSetting('settingsButton')
+    }
+
+    init () {
+        this.setButtonSettings()
+
+        const handleElementForPath = this.handleElementForPath.bind(this)
+        const historyMethodProxy = new DDGProxy(this, History.prototype, 'pushState', {
```
jonathanKingston, line 234: Did we do this already for element hiding? We should probably share that same code eventually. Can you make it a follow up (I'm aware they currently won't run at the same time)

injected/integration-test/autofill-password-import.spec.js
dbajpeyi: Changed to await after click event: https://github.com/duckduckgo/content-scope-scripts/pull/1157/commits/e18ebbc3e7797139e02c8fc07706e986966dc098

injected/src/features/autofill-password-import.js
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * Checks if the path is supported and animates/taps the element if it is.
+     * @param {string} path
+     */
+    async handleElementForPath (path) {
+        // FIXME: we need to check if the path is supported, otherwise
+        // for some reason google doesn't wait to proceed with the signin step.
+        // Not too sure why this is happening, there are no errors on the console.
+
+        if ([this.#exportButtonSettings?.path, this.#settingsButtonSettings?.path, this.#signInButtonSettings?.path].indexOf(path) !== -1) {
```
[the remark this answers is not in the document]
dbajpeyi: The fixme is an older comment, removed now https://github.com/duckduckgo/content-scope-scripts/pull/1157/commits/e18ebbc3e7797139e02c8fc07706e986966dc098. Now the whole block is within a `try`-`catch` block, so were' safe anyway.

We should probably anyway avoid running unnecessary code, so this is a nice guard anyway.

injected/src/features/password-import.js
jonathanKingston: Probably fine given we're only injecting this in one place. We don't really have much confidence of data passed in though so one for the future; happy to leave as is :)

jonathanKingston:
I think [injected/integration-test/extension/contentScope.js](https://github.com/duckduckgo/content-scope-scripts/pull/1157/files#diff-52d8aac4403de219c5fb72889e1594d72589725f3bf8c08fa0364a8ab2435b7e) shouldn't be checked in, I suspect that's what the error is about.

jonathanKingston (approved):
LGTM! Thanks for getting this in!

PR #1165, opened 2024-10-25
Add getInterestGroupAdAuctionData removal

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/0/0/1208626149160771/f

## Description

Conditionally removes the `getInterestGroupAdAuctionData` API as specified here: https://github.com/WICG/privacy-preserving-ads/blob/main/API%20Details.md

## Testing Steps

- 

## Checklist

*Please tick all that apply:*

### DIFF
--- a/injected/src/features/google-rejected.js
+++ b/injected/src/features/google-rejected.js
@@ -21,6 +21,12 @@ export default class GoogleRejected extends ContentFeature {
             if ('adAuctionComponents' in Navigator.prototype) {
                 delete Navigator.prototype.adAuctionComponents
             }
+            // https://github.com/WICG/privacy-preserving-ads/blob/main/API%20Details.md
+            if (this.getFeatureSetting('getInterestGroupAdAuctionData')) {
+                if ('getInterestGroupAdAuctionData' in Navigator.prototype) {
+                    delete Navigator.prototype.getInterestGroupAdAuctionData
+                
+            }
         } catch {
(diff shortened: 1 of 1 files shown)


injected/src/features/google-rejected.js
```diff
@@ -21,6 +21,12 @@ export default class GoogleRejected extends ContentFeature {
             if ('adAuctionComponents' in Navigator.prototype) {
                 delete Navigator.prototype.adAuctionComponents
             }
+            // https://github.com/WICG/privacy-preserving-ads/blob/main/API%20Details.md
+            if (this.getFeatureSetting('getInterestGroupAdAuctionData')) {
```
kdzwinel, line 25: Why put it behind additional flag?

kdzwinel (commented):
This API has not really shipped yet (AFIAK - tested in 132.0.6797.0 ), so we are adding it early, right? I can see it in chromium's source code, so I guess it's coming.

injected/src/features/google-rejected.js
jonathanKingston: Good practice really. In case we want to enable the API on a particular website, additionally there's higher risk as we've not seen it in the wild yet.

kdzwinel (approved):
LGTM (that said, I wasn't able to fully test it because I'm not sure how to enable this API even via chrome://flags).

PR #1174, opened 2024-10-29
extract mocks into mock-transport

shakyShane:
**Asana Task/Github Issue:** n/a

## Description

This change ensures the mock transport code is stripped from the production build. It's also a more powerful pattern going forward since it's mocking exactly the boundary and nothing else.

## Testing Steps

- 

## Checklist

*Please tick all that apply:*


special-pages/pages/release-notes/src/js/mock-transport.js
shakyShane, line 1: @mgurgel - now you can import anything you want in this file, safe in the knowledge it's stripped from the production build. Next up: need to do the same with the components file, but that can follow once I've figured it out.

mgurgel (approved):
👍

PR #1184, opened 2024-10-31
[Autofill password import] Misc fixes

dbajpeyi:
**Asana Task/Github Issue:** https://app.asana.com/0/1205996472158114/1208671198302487/f

## Description
1. **Bugfix** It seems like I missed out on replacing `indexOf` by `includes`, however I replaced the functionality 😄 
Original PR: https://github.com/duckduckgo/content-scope-scripts/pull/1157. This was causing login to break, and also causing other animations to break.
2. **Feature**  Modifying the animation approach by adding an overlay element, so that the main element doesn't accidentally scale and transformation is only added to the overlay
3. **Bugfix**  Adding a timeout before animation, as sometimes `scrollIntoView` fails if the layout is not fully rendered (edge case).
4. **Feature** Some fixes in styles as suggested in the ship review.
5. **Feature** Update overlay position on scroll
6. **Test** Update test to check for overlay

## Testing Steps

- Build android (on the branch `feature/craig/autofill_gpm_ui_to_launch_from_password_manager`) with the config `https://jsonblob.com/api/1301563859532636160`
- Hit the fire button, go to "passwords" menu
- Start importing password using the new "Import passwords from google" button
- You should be able to go through the whole flow (including adding new account on google.com)

## Checklist

*Please tick all that apply:*

- [x] This change was covered by a ship review


special-pages/pages/new-tab/app/update-notification/update-notification.service.js
```diff
@@ -41,7 +41,7 @@ export class UpdateNotificationService {
      */
     dismiss () {
         this.ntp.messaging.notify('updateNotification_dismiss')
-        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
dbajpeyi, line 44: Some new eslint changes seem to have forced this when I did a lint-fix

injected/src/features/autofill-password-import.js
```diff
@@ -76,39 +99,116 @@ export default class AutofillPasswordImport extends ContentFeature {
...
+     * @returns {HTMLElement|Element|null}
+     */
+    insertOverlayElement (mainElement, style) {
+        this.removeOverlayIfNeeded()
+
+        const overlay = document.createElement('div')
+        overlay.setAttribute('id', OVERLAY_ID)
+        const svgElement = mainElement.parentNode?.querySelector('svg') ?? mainElement.querySelector('svg')
+
+        const isRound = style.borderRadius === '100%'
```
jonathanKingston, line 136: Should we be using `getComputedStyle` for this?

injected/src/features/autofill-password-import.js
```diff
@@ -157,17 +257,27 @@ export default class AutofillPasswordImport extends ContentFeature {
...
             try {
-                const { element, style, shouldTap } = await this.getElementAndStyleFromPath(path) ?? {}
+                const { element, style, shouldTap, shouldWatchForRemoval } = await this.getElementAndStyleFromPath(path) ?? {}
                 if (element != null) {
-                    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-                    shouldTap ? this.autotapElement(element) : this.animateElement(element, style)
+                    if (shouldTap) {
+                        this.autotapElement(element)
+                    } else {
+                        setTimeout(() => this.animateElement(element, style), 300)
```
jonathanKingston, line 273: Can you make this `300` a constant, mainly to document what the wait is for a little bit (comment is probably unnecessary with a well named var).

injected/src/features/autofill-password-import.js
dbajpeyi: @jonathanKingston this is just the "preset style" that _we're_ setting here, if you're talking about the line mentioned here. That `style` parameter gets passed on to this function.

injected/src/features/autofill-password-import.js
```diff
@@ -76,39 +99,116 @@ export default class AutofillPasswordImport extends ContentFeature {
...
+        overlay.setAttribute('id', OVERLAY_ID)
+        const svgElement = mainElement.parentNode?.querySelector('svg') ?? mainElement.querySelector('svg')
+
+        const isRound = style.borderRadius === '100%'
+        const elementToCenterOn = isRound ? svgElement : mainElement
+        if (elementToCenterOn) {
+            const { top, left, width, height } = elementToCenterOn.getBoundingClientRect()
+            overlay.style.position = 'absolute'
+    
+            overlay.style.top = `calc(${top}px + ${window.scrollY}px - ${isRound ? height / 2 : 0}px - 1px - ${style.offsetTopEm}em)`
```
jonathanKingston, line 143: I'd prefer these ternaries as variables outside of the template string to simplify the expression a little. Feel free to skip if you care about keeping.

injected/src/features/autofill-password-import.js
jonathanKingston: Ah sorry misread thanks for the clarification. :)

injected/src/features/autofill-password-import.js
```diff
@@ -157,17 +258,27 @@ export default class AutofillPasswordImport extends ContentFeature {
...
             try {
-                const { element, style, shouldTap } = await this.getElementAndStyleFromPath(path) ?? {}
+                const { element, style, shouldTap, shouldWatchForRemoval } = await this.getElementAndStyleFromPath(path) ?? {}
                 if (element != null) {
-                    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-                    shouldTap ? this.autotapElement(element) : this.animateElement(element, style)
+                    if (shouldTap) {
+                        this.autotapElement(element)
+                    } else {
+                        setTimeout(() => this.animateElement(element, style), ANIMATION_TIMEOUT)
```
jonathanKingston, line 274: Maybe this is safer:

```js
const domLoaded = new Promise((resolve) => {
  if (document.readyState === "loading") {
    document.addEventListener("DOMContentLoaded", resolve);
  } else {
    resolve();
  }
})
```

```suggestion
                        await domLoaded
                        this.animateElement(element, style)
```

injected/src/features/autofill-password-import.js
```diff
@@ -77,38 +102,164 @@ export default class AutofillPasswordImport extends ContentFeature {
...
 
+    /**
+     * Removes the overlay if it exists.
+     */
+    removeOverlayIfNeeded() {
+        const existingOverlay = document.getElementById(OVERLAY_ID);
+        if (existingOverlay != null) {
+            existingOverlay.style.display = 'none';
+            existingOverlay.remove();
+            document.removeEventListener('scroll', this.updateOverlayPosition.bind(this));
```
jonathanKingston, line 121: This won't remove the other listener as there is a closure involved (but it wouldn't be equivelant even if you added it here and also the bind produces a completely new function instance too).

So there's wither two ways:

1. Store the scroll listener in a global:

```js
this.scrollListener = () => {...}
document.addEventListener('scroll', this.scrollListener);
....
document.removeEventListener('scroll', this.scrollListener);
```

2. Class based event handler:
```js
export default class AutofillPasswordImport extends ContentFeature {
    insertOverlay() {
        document.addEventListener('scroll', this);
    }
    removeOverlayIfNeeded() {
        document.removeEventListener('scroll', this);
    }
    handleEvent(e) {
        if (e.type === 'scroll') {
            requestAnimationFrame(() => this.updateOverlayPosition(overlay, style));
        }
    }
}
```

My preferred is 2. purely as it doesn't require another global and also centralises event handling.

injected/src/features/autofill-password-import.js
```diff
@@ -77,38 +102,164 @@ export default class AutofillPasswordImport extends ContentFeature {
...
+
+            const mainElementRect = mainElement.getBoundingClientRect();
+            overlay.style.width = `${mainElementRect.width}px`;
+            overlay.style.height = `${mainElementRect.height}px`;
+            overlay.style.zIndex = style.zIndex;
+
+            // Ensure overlay is non-interactive
+            overlay.style.pointerEvents = 'none';
+
+            document.addEventListener('scroll', () => {
```
jonathanKingston, line 173: Should we use the `passive` argument here?

injected/src/features/autofill-password-import.js
```diff
@@ -157,13 +308,33 @@ export default class AutofillPasswordImport extends ContentFeature {
...
+                this.setElementToCenterOn(element, style);
                 if (element != null) {
-                    // eslint-disable-next-line @typescript-eslint/no-unused-expressions
-                    shouldTap ? this.autotapElement(element) : this.animateElement(element, style);
+                    if (shouldTap) {
+                        this.autotapElement(element);
+                    } else {
+                        const domLoaded = new Promise((resolve) => {
+                            if (document.readyState === 'loading') {
+                                document.addEventListener('DOMContentLoaded', resolve);
```
jonathanKingston, line 323: Also we could pass `once` here. Not really a problem just might promote it's use for when it's more needed.

dbajpeyi: Sorry not sure I fully understood but I suppose https://github.com/duckduckgo/content-scope-scripts/pull/1184/commits/52362406bd13197f731571649ade03c91f79b5fa

injected/src/features/autofill-password-import.js
dbajpeyi: Yep makes sense here!

injected/src/features/autofill-password-import.js
```diff
@@ -222,22 +478,28 @@ export default class AutofillPasswordImport extends ContentFeature {
...
 
-        document.addEventListener('DOMContentLoaded', async () => {
-            await handleElementForPath(window.location.pathname);
+        this.#domLoaded = new Promise((resolve) => {
+            document.addEventListener('DOMContentLoaded', async () => {
+                // @ts-expect-error - caller doesn't expect a value here
+                resolve();
+                const path = window.location.pathname;
+                await handlePath(path);
+            });
```
jonathanKingston, line 502:
```suggestion
        this.#domLoaded = new Promise((resolve) => {
            if (document.readyState !== 'loading') {
                resolve();
                return;
            }
            document.addEventListener('DOMContentLoaded', async () => {
                // @ts-expect-error - caller doesn't expect a value here
                resolve();
                const path = window.location.pathname;
                await handlePath(path);
            }, {once: true});
```

jonathanKingston (approved):

PR #1185, opened 2024-10-31
Shared eslint config

muodov:
https://app.asana.com/0/1201807839809179/1208561434802767/f

- Upgrade to eslint 9
- use shared DDG config
- fix some lint errors


injected/integration-test/fingerprint.spec.js
```diff
@@ -5,6 +5,7 @@ import { test as base, expect } from '@playwright/test'
 import { testContextForExtension } from './helpers/harness.js'
 import { createRequire } from 'node:module'
 
+// eslint-disable-next-line no-redeclare
```
jonathanKingston, line 8: Are we importing 'require' as a global, we shouldn't

injected/src/content-feature.js
```diff
@@ -214,7 +217,7 @@ export default class ContentFeature {
         })
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-empty-function
```
jonathanKingston, line 217: Did we drop these checks?

injected/src/features/broker-protection/types.js
```diff
@@ -94,7 +94,7 @@ export class AsyncProfileTransform {
      * @param {Record<string, any>} profileParams - the original action params from `action.profile`
      * @return {Promise<Record<string, any>>}
      */
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+
```
jonathanKingston, line 97: Can we delete these newlines? I think we should be checking for unused method though.

injected/src/features/click-to-load.js
```diff
@@ -1887,7 +1887,7 @@ export default class ClickToLoad extends ContentFeature {
         })
         // Listen to message from Platform letting CTL know that we're ready to
         // replace elements in the page
-        // eslint-disable-next-line promise/prefer-await-to-then
```
jonathanKingston, line 1890: Can we make sure we use `--report-unused-disable-directives`? Looks like we nolonger needed this.

injected/src/features/broker-protection/types.js
jonathanKingston: Ah are we not intending to adopt prettier in the same PR? I think we should tbh.

I'm hoping prettier drops unnecessary whitespace at the start of lines like this.

eslint.config.mjs
```diff
@@ -0,0 +1,104 @@
+import tseslint from 'typescript-eslint';
+import ddgConfig from "@duckduckgo/eslint-config";
+import globals from "globals";
+
+// @ts-check
+export default tseslint.config(
+    ...ddgConfig,
```
jonathanKingston, line 7: Is this spreading not possible in an eslintrc? (that seems the only reason to change file format)

injected/integration-test/duckplayer.spec.js
```diff
@@ -39,7 +39,7 @@ test.describe('Thumbnail Overlays', () => {
         const navigation = overlays.requestWillFail()
         await overlays.clicksFirstShortsThumbnail()
         const url = await navigation
-        await overlays.opensShort(url)
+        overlays.opensShort(url)
```
jonathanKingston, line 42: This changes the meaning of the test. I believe we're intending to validate if this method throws which won't happen without an `await`.

jonathanKingston: Ignore, checked the fn signature

eslint.config.mjs
muodov: Since we're updating to eslint 9, I wanted to use the flat config to avoid possible incompatibilities

jonathanKingston: Should use `eslint.config.js` then?

I've got zero context on the new format just I'm really not a fan of config files written in dynamic languages.

injected/src/features/click-to-load.js
muodov: Most of the changes in this PR were a result of `eslint . --fix`. This and other `eslint-disable` lines were removed because they were no longer disabling any errors. I reviewed all occurrences, and they were all correct.

For example, a bunch of "unused variables" are not flagged anymore because we [allow unused function arguments](https://github.com/duckduckgo/eslint-config/blob/main/index.js#L148). That rule came [directly from Standard](https://github.com/standard/eslint-config-standard/blob/v17.1.0/.eslintrc.json#L176), so I suspect we were not applying those rules correctly before.

eslint.config.mjs
muodov: It's their default config format now, for better or for worse. It is dynamic, but the actual behaviour is much more simple now.

injected/src/content-feature.js
muodov: `@typescript-eslint/no-unused-vars` is now [allowed for function arguments](https://github.com/duckduckgo/eslint-config/blob/main/index.js#L148). I'm not sure why it was flagged before because [it's in Standard](https://github.com/standard/eslint-config-standard/blob/v17.1.0/.eslintrc.json#L176).

`@typescript-eslint/no-empty-function` is not flagged on this line even in the main branch.

build-output.eslint.config.js
```diff
@@ -0,0 +1,14 @@
...
+
+export default [
+    {
+        languageOptions: {
+            ecmaVersion: "latest",
+            sourceType: "script",
+        },
+        rules: {
+            "no-implicit-globals": "error",
+        }
```
muodov, line 12: If I understood correctly, we only care about this rule when checking the bundle, so I didn't enable any plugins or recommended rules.

injected/src/features/broker-protection/types.js
muodov: I'm hoping to follow up with the Prettier PR (it will need changes in the config repo too). So I'll leave it up to you whether to merge this now or wait until the second PR.

injected/integration-test/fingerprint.spec.js
muodov: The new config [assumes NodeJS globals in test files](https://github.com/duckduckgo/content-scope-scripts/pull/1185/files#diff-a32a0887ed9d1d707bbb3b845b7df7fd40e673c47e7b60a3ebd896b68d3b8839R69-R76)

muodov:
Ok this is ready for review now. I recommend looking commit-by-commit for better context

jonathanKingston (approved):

jonathanKingston (approved):

PR #1207, opened 2024-11-06
Add API manipulation feature

jonathanKingston:
**Asana Task/Github Issue:**  https://app.asana.com/0/481882893211075/1208706976317187/f

## Description

Adding the ability to modify DOM apis remotely through the config.

## Testing Steps

This is draft currently however the manual steps I've been taking are:
- Building the extension with `npm build`
- Then loading the extension with:
    - `web-ext run -s integration-test/extension/ -t chromium`
- Opening example.com and validating the APIs have been modified as expected based on: `injected/entry-points/integration.js`

**Note:** My expectation is to move all these modifications into a test and out of the sample extension so keep a track of that sample config changing for the test.

## Checklist

*Please tick all that apply:*

### DIFF
--- a/injected/entry-points/integration.js
+++ b/injected/entry-points/integration.js
@@ -28,10 +28,70 @@ function generateConfig() {
             domain: topLevelUrl.hostname,
             isBroken: false,
             allowlisted: false,
-            enabledFeatures: ['fingerprintingCanvas', 'fingerprintingScreenSize', 'navigatorInterface', 'cookie'],
+            enabledFeatures: [
+                'apiManipulation',
+                'fingerprintingCanvas',
+                'fingerprintingScreenSize',
+                'navigatorInterface',
+                'cookie'
+            ],
         },
-        trackerLookup,
-    };
+        featureSettings: {
+            apiManipulation: {
+                "apiChanges": {
+                    "Navigator.prototype.joinAdInterestGroup": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.leaveAdInterestGroup": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.clearOriginJoinedAdInterestGroups": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.updateAdInterestGroups": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.createAuctionNonce": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.runAdAuction": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.adAuctionComponents": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.deprecatedURNToURL": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.deprecatedReplaceInURN": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.getInterestGroupAdAuctionData": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.createAdRequest": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.finalizeAd": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.canLoadAdAuctionFencedFrame": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.deprecatedRunAdAuctionEnforcesKAnonymity": {
+                        "type": "remove"
+                    },
+                    "Navigator.prototype.protectedAudience": {
+                        "type": "wrapPropertyValue",
+                        "value": {
+                            "type": "undefined"
+                        }
+                    },
+                },
+            },
+        },
+        trackerLookup
+    }
 }
 
 /**
--- a/injected/entry-points/integration.js
+++ b/injected/entry-points/integration.js
@@ -28,61 +28,7 @@ function generateConfig() {
             domain: topLevelUrl.hostname,
             isBroken: false,
             allowlisted: false,
-            enabledFeatures: ['apiManipulation', 'fingerprintingCanvas', 'fingerprintingScreenSize', 'navigatorInterface', 'cookie'],
-        },
-        featureSettings: {
-            apiManipulation: {
-                apiChanges: {
-                    'Navigator.prototype.joinAdInterestGroup': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.leaveAdInterestGroup': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.clearOriginJoinedAdInterestGroups': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.updateAdInterestGroups': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.createAuctionNonce': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.runAdAuction': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.adAuctionComponents': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.deprecatedURNToURL': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.deprecatedReplaceInURN': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.getInterestGroupAdAuctionData': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.createAdRequest': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.finalizeAd': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.canLoadAdAuctionFencedFrame': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.deprecatedRunAdAuctionEnforcesKAnonymity': {
-                        type: 'remove',
-                    },
-                    'Navigator.prototype.protectedAudience': {
-                        type: 'descriptor',
-                        getterValue: {
-                            type: 'undefined',
-                        },
-                    },
-                },
-            },
+            enabledFeatures: ['fingerprintingCanvas', 'fingerprintingScreenSize', 'navigatorInterface', 'cookie'],
         },
         trackerLookup,
     };
--- a/injected/entry-points/integration.js
+++ b/injected/entry-points/integration.js
@@ -28,70 +28,64 @@ function generateConfig() {
             domain: topLevelUrl.hostname,
             isBroken: false,
             allowlisted: false,
-            enabledFeatures: [
-                'apiManipulation',
-                'fingerprintingCanvas',
-                'fingerprintingScreenSize',
-                'navigatorInterface',
-                'cookie'
-            ],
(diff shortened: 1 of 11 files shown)


injected/src/features/api-manipulation.js
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Wraps a property with descriptor.
+     * @param {object} api
+     * @param {string} key
+     * @param {APIChange} change
+     */
+    wrapApiDescriptor(api, key, change) {
+        this.wrapProperty(api, key, {
+            get: () => processAttr(change.getterValue, undefined),
+            enumerable: change.enumerable || false,
```
jonathanKingston, line 92: `true` might be a better default here for both of these.

injected/src/features/api-manipulation.js
```diff
@@ -0,0 +1,118 @@
...
+            }
+        }
+    }
+
+    /**
+     * Checks if the config API change is valid.
+     * @param {any} change
+     * @returns {change is APIChange}
+     */
+    checkIsValidAPIChange(change) {
```
jonathanKingston, line 23: This basic validation is to prevent a broken config. I'm intending to get this generated from TS used in the config repo when we make that change. This will prevent errors in the config ever happening and also ensure no typing issues on our code side.

jonathanKingston: IIRC there's similar checking that fails safe on the `processAttr` method but I'll check.

injected/src/features/api-manipulation.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * Removes a method from an API.
+     * @param {object} api
+     * @param {string} key
+     */
+    removeApiMethod(api, key) {
+        try {
+            if (key in api) {
```
jonathanKingston, line 77: I think this is the safest way to check for the existence of the `key`, but happy to be wrong. Ideally we don't want to call the getter if it exists. For example `if (api[key]) {` threw.

jonathanKingston:
@muodov this still needs tests, just a heads up if you had feedback. I'm sure we can add a bunch of resilience later. I'm mostly just aiming the current use-case and we can expand further later.

injected/src/features/api-manipulation.js
jonathanKingston: I switched to `true` as it's much more common.

injected/src/features/api-manipulation.js
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Wraps a property with descriptor.
+     * @param {object} api
+     * @param {string} key
+     * @param {APIChange} change
+     */
+    wrapApiDescriptor(api, key, change) {
+        const getterValue = change.getterValue;
+        if (getterValue) {
+            this.wrapProperty(api, key, {
```
jonathanKingston, line 101: @muodov I think `wrapProperty` should be the default here; happy for you to say otherwise.

I had this POC patch that added the following config support:
```js
                    'Navigator.prototype.bloop': {
                        type: 'descriptor',
                        value: {
                            type: 'function',
                            functionName: 'debug'
                        }
                    },
                    'Navigator.prototype.bloop2': {
                        type: 'descriptor',
                        define: true,
                        value: {
                            type: 'function',
                            functionName: 'debug'
                        }
                    },
                    'Navigator.prototype.bloop3': {
                        type: 'descriptor',
                        define: true,
                        getterValue: {
                            type: 'number',
                            value: 1222
                        }
                    },
```


This changed this code to:

```js
        const getterValue = change.getterValue;
        const defineMethod = change.define ? this.defineProperty.bind(this) : this.wrapProperty.bind(this);
        if (getterValue) {
            defineMethod(api, key, {
                get: () => processAttr(getterValue, undefined),
                enumerable: change.enumerable ?? true,
                configurable: change.configurable ?? true,
            });
        } else if (change.value) {
            defineMethod(api, key, {
                value: processAttr(change.value, undefined),
                writable: change.writable ?? true,
                enumerable: change.enumerable ?? true,
                configurable: change.configurable ?? true,
            });
        }
```

The current default not to define seems what we want and adding support for this `define` and `value` later will be simple.

injected/entry-points/integration.js
```diff
@@ -28,7 +28,61 @@ function generateConfig() {
             domain: topLevelUrl.hostname,
             isBroken: false,
             allowlisted: false,
-            enabledFeatures: ['fingerprintingCanvas', 'fingerprintingScreenSize', 'navigatorInterface', 'cookie'],
+            enabledFeatures: ['apiManipulation', 'fingerprintingCanvas', 'fingerprintingScreenSize', 'navigatorInterface', 'cookie'],
+        },
+        featureSettings: {
```
muodov, line 33: This is only for tests, right? Is it necessary to keep the default settings in this file?

injected/src/features/api-manipulation.js
muodov: Do we really need runtime validation for config? It feels like we could just get away with static checks on the config side and avoid the runtime tax?

injected/src/features/api-manipulation.js
```diff
@@ -0,0 +1,130 @@
...
+            });
+        }
+    }
+
+    /**
+     * Looks up a global object from a scope, e.g. 'Navigator.prototype'.
+     * @param {string} scope the scope of the object to get to.
+     * @returns {[object, string]|null} the object at the scope.
+     */
+    getGlobalObject(scope) {
```
muodov, line 114: I could swear I wrote this function before, but maybe it was in previous life.
Maybe move this to wrapper-utils.js?

injected/src/features/api-manipulation.js
muodov: Agreed. I tried a few different API signatures when iterating on wrapper functions, but it gets messy. I think making it explicit in config is the cleanest way.

injected/src/features/api-manipulation.js
```diff
@@ -0,0 +1,130 @@
...
+     * @param {string} key
+     * @param {APIChange} change
+     */
+    wrapApiDescriptor(api, key, change) {
+        const getterValue = change.getterValue;
+        if (getterValue) {
+            this.wrapProperty(api, key, {
+                get: () => processAttr(getterValue, undefined),
+                enumerable: change.enumerable ?? true,
+                configurable: change.configurable ?? true,
```
muodov, line 104: `wrapProperty()` works only on existing properties, so it always defaults to the original descriptor parameters.
I suggest not passing `enumerable` and `configurable` unless we really want them to be different.

injected/entry-points/integration.js
jonathanKingston: No, I was going to remove after you'd reviewed :). This is just for the ability to manually run it easily (I'm going to suggest we add this as a feature in future for development that we can take a known config and run with it across page navs).

injected/src/features/api-manipulation.js
jonathanKingston: Yeah we had one in runtime checks also. I did try looking if we had another.

injected/src/features/api-manipulation.js
jonathanKingston: I've removed passing the defaults, so if undefined it should work as you mentioned :)

injected/src/features/api-manipulation.js
jonathanKingston: I'm normally of the same opinion too. For this I wanted to block on us adding TS to the config repo before we harden this up further rather than JSON schema it and then rewrite.

Given the risk of misconfig I think the cost here isn't significant and worthwhile to keep at least until we have that hardening.

injected/src/features/api-manipulation.js
jonathanKingston: Branch: https://github.com/duckduckgo/content-scope-scripts/compare/jkt/apiManipulation...jkt/apiModificationUnitTestGetGlobalObject?expand=1

injected/src/features/api-manipulation.js
```diff
@@ -100,8 +100,9 @@ export default class ApiManipulation extends ContentFeature {
         if (getterValue) {
             this.wrapProperty(api, key, {
                 get: () => processAttr(getterValue, undefined),
-                enumerable: change.enumerable ?? true,
-                configurable: change.configurable ?? true,
+                // wrapProperty takes care of setting these values to the api[key] default values.
+                enumerable: change.enumerable,
+                configurable: change.configurable,
```
muodov, line 105: this would pass undefined values, but the keys would still be there. I think you need to avoid adding the key altogether.

jonathanKingston: Seems like an oversight on the API side tbh but thanks for the heads up :)

injected/src/features/api-manipulation.js
muodov: Supposedly we'll only want to manipulate own properties, so how about using `Object.hasOwnProperty()`?

injected/src/features/api-manipulation.js
muodov: Yeah maybe. I don't remember if `Object.defineProperty()` throws when it gets `enumerable: undefined`, but if it does, we should do this check inside the wrapper.

injected/src/features/api-manipulation.js
jonathanKingston: We agreed to add a removal notice task with a timeline in Asana.

injected/src/features/api-manipulation.js
```diff
@@ -0,0 +1,137 @@
...
+     * @param {string} key
+     * @param {APIChange} change
+     */
+    wrapApiDescriptor(api, key, change) {
+        const getterValue = change.getterValue;
+        if (getterValue) {
+            const descriptor = {
+                get: () => processAttr(getterValue, undefined),
+            };
+            if (change.enumerable) {
```
muodov, line 106:
```suggestion
            if (typeof change.enumerable === 'boolean') {
```

injected/src/features/api-manipulation.js
```diff
@@ -0,0 +1,137 @@
...
+                descriptor.enumerable = change.enumerable;
+            }
+            if (change.configurable) {
```
muodov, line 109: if (typeof change.configurable === 'boolean') {

injected/src/features/api-manipulation.js
jonathanKingston: Isn't there a risk of that being hooked on the regular? Happy to change though.

injected/src/features/api-manipulation.js
jonathanKingston: We already check this above?

injected/src/features/api-manipulation.js
jonathanKingston: We already check this above?

muodov (approved):
lgtm

injected/src/features/api-manipulation.js
muodov: I meant that if it's `false` this wouldn't work

jonathanKingston: ah gah, probably better to just check `in` then.

injected/src/features/api-manipulation.js
muodov: I mean, we are relying on captured-globals all over the place now, so this should work too.
Practically it's not a big deal because hopefully we'll never do this. And if we do, it wouldn't work anyways because the wrapper function below works only on own properties.

muodov (approved):

PR #1213, opened 2024-11-11
messaging: page-world bridge

shakyShane:
**Asana Task/Github Issue:** https://app.asana.com/0/1201141132935289/1208740665800207/f

## Description

- [x] added a new feature called `messageBridge` (apple-isolated only for now)
- [x] added `createMessageBridge` on `navigatorInterface` - to return the MessagingInterface
   -  this is always present, but will throw an exception if the bridge is not allowed, installed or enabled
   - exception is just a generic message, it does not reveal a reason

## Testing Steps

- See the macOS PR for where I implemented this E2E https://github.com/duckduckgo/macos-browser/pull/3558

## Checklist

*Please tick all that apply:*


shakyShane:
* **#1213** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/1213?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

injected/integration-test/test-pages/bridge/index.html
```diff
@@ -0,0 +1,58 @@
...
+            return Reflect.apply(target, thisArg, argArray)
+        }
+    })
+
+    globalThis.CustomEvent = new Proxy(globalThis.CustomEvent, {
+        construct(target, argArray, newTarget) {
+            console.log('💣, CustomEvent constructor');
+            return Reflect.construct(target, argArray, newTarget);
+        }
+    })
```
shakyShane, line 33: just proving that the usual suspects can't see the messages

injected/src/features/message-bridge/create-page-world-bridge.js
```diff
@@ -0,0 +1,134 @@
...
+    console.log('will createPageWorldBridge for', featureName, 'with token', token);
+
+    /**
+     * TODO: what is the value of this token? - since 'createPageWorldBridge' is available to all page scripts, probably none
+     * @param {string} eventName
+     * @return {`${string}-${string}`}
+     */
+    function appendToken(eventName) {
+        return `${eventName}-${token}`;
+    }
```
shakyShane, line 40: i'm not tool convinced there's any value in using a token across both ends of the proxy - even one that's safe. Because in the end we're still exposing the API to the page world - meaning scripts call `navigator.duckduckgo.createMessagingBridge()`. The main difference this PR introduces, is that it allows messaging without **directly** exposing the webkit handlers to the page - it's not here to be 100% secure, unfortunately

injected/entry-points/apple.js
```diff
@@ -8,7 +8,34 @@ import { WebkitMessagingConfig, TestTransportConfig } from '../../messaging/inde
...
+                        {
+                            op: 'replace',
+                            path: '/allowed',
+                            value: ['aiChat', 'duckPlayer'],
+                        },
+                    ],
+                },
+            ],
+        },
+    };
```
shakyShane, line 31: An example of how the config might look for this feature. Open to ideas, but mainly I want to stick within the feature set already supported by CSS (the example here is already supported for domain restrictions)

injected/src/features/message-bridge.js
```diff
@@ -0,0 +1,168 @@
...
+    ProxyResponse,
+    SubscriptionRequest,
+    SubscriptionResponse,
+    SubscriptionUnsubscribe,
+} from './message-bridge/schema.js';
+
+/**
+ * This part has access to messaging handlers
+ */
+export class MessageBridge extends ContentFeature {
```
shakyShane, line 16: Everything in this class is running within the isolated world on macOS. If we know that windows/android will not, then we'll need to do some globals-capturing here too.

injected/src/features/message-bridge.js
```diff
@@ -0,0 +1,168 @@
...
+
+    init(args) {
+        console.log(`[isolated] will install bridge with ${args.sessionKey}`);
+        /**
+         * TODO: what is the value of this token? - since 'createPageWorldBridge' is available to all page scripts, probably none
+         * @param {string} eventName
+         * @return {`${string}-${string}`}
+         */
+        function appendToken(eventName) {
+            return `${eventName}-${args.sessionKey}`;
```
shakyShane, line 28: again I'm not too convinced this token/key has any value.

injected/src/features/message-bridge.js
```diff
@@ -0,0 +1,168 @@
...
+     */
+    installProxyFor(featureName, config) {
+        if (this.proxies.has(featureName)) return;
+        const allowed = this.getFeatureSetting('allowed');
+        if (!allowed.includes(featureName)) return;
+
+        const ctx = { ...this.messaging.messagingContext, featureName };
+        const messaging = new Messaging(ctx, config);
+        this.proxies.set(featureName, messaging);
+        console.log('[isolated] did install proxy for ', featureName);
```
shakyShane, line 74: This is the magic - it created a new `Messaging` instance in the fly, but crucially it re-uses the per-platform config. This is very important since each platform may have varying levels of global side-effects.

injected/src/features/message-bridge/schema.js
shakyShane, line 1: This file is a bit noisy, but since we lack any run-time validation stuff, I wanted to be extra careful when receiving 
 and sending messages

injected/src/features/navigator-interface.js
```diff
@@ -27,6 +29,13 @@ export default class NavigatorInterface extends ContentFeature {
                     isDuckDuckGo() {
                         return DDGPromise.resolve(true);
                     },
+                    /**
+                     * @param {string} featureName
+                     * @return {Pick<import("@duckduckgo/messaging").Messaging, "notify"|"subscribe"|"request">}
+                     */
+                    createMessageBridge(featureName) {
+                        return createPageWorldBridge(featureName, args.sessionKey);
+                    },
```
shakyShane, line 38: finally, this is the point where we expose to the page.

injected/src/features/navigator-interface.js
```diff
@@ -13,6 +14,7 @@ export default class NavigatorInterface extends ContentFeature {
     }
 
     injectNavigatorInterface(args) {
+        console.log('page-world args', args);
```
jonathanKingston, line 17:
```suggestion
```

injected/src/features/broker-protection/utils.js
```diff
@@ -240,3 +240,19 @@ export async function hashObject(profile) {
 
     return hashHex;
 }
+
+/**
+ * @param {unknown} input
+ * @return {input is Object}
+ */
+export function isObject(input) {
```
jonathanKingston, line 248: This feels like it should be in the general utils possibly? Kinda shocked we don't have this already.

injected/src/features/message-bridge.js
```diff
@@ -0,0 +1,168 @@
...
+            this.proxyNotification(notification);
+        });
+        window.addEventListener(appendToken(ProxyRequest.NAME), (/** @type {CustomEvent<unknown>} */ e) => {
+            console.log(`[isolated:incoming] ${ProxyRequest.NAME}`, e.detail);
+            const request = ProxyRequest.create(e.detail);
+            if (!request) return;
+            this.proxyRequest(request, appendToken);
+        });
+        window.addEventListener(appendToken(SubscriptionRequest.NAME), (/** @type {CustomEvent<unknown>} */ e) => {
+            console.log(`[isolated:incoming] ${SubscriptionRequest.NAME}`, e.detail);
```
jonathanKingston, line 49: This debugging feels like it might be useful, can you make a file specific debug method that is disabled by default?

injected/src/features/message-bridge/create-page-world-bridge.js
```diff
@@ -0,0 +1,134 @@
...
+    ProxyNotification,
+    ProxyRequest,
+    ProxyResponse,
+    SubscriptionRequest,
+    SubscriptionResponse,
+    SubscriptionUnsubscribe,
+} from './schema.js';
+
+/**
+ * todo: use the already-captured ones
```
jonathanKingston, line 12: Yeah import all of these

injected/src/features/message-bridge/create-page-world-bridge.js
jonathanKingston: I don't see much harm in keeping the method to wrap it with a string that reduces the chance of colliding with pages.

I'd rather we not use the session token as it might leak.

We could use a derived key from it but as you mention the security doesn't matter that much (we need to gate the client side)

jonathanKingston: aka a prefix of `ddg-something-message-something-` is probably more useful than a unique identifier.

The unique token would reduce the pages ability to snoop on the message contents as well.

injected/src/features/message-bridge.js
```diff
@@ -0,0 +1,168 @@
...
+
+        const proxy = this.proxies.get(featureName);
+        if (!proxy) return console.warn('[isolated] proxy was not installed');
+
+        const responseName = appendToken(ProxyResponse.NAME + '-' + id);
+        console.log('[isolated] will proxy', request);
+
+        proxy
+            ?.request(method, params)
+            // eslint-disable-next-line promise/prefer-await-to-then
```
jonathanKingston, line 92: Can you make the method async?

injected/src/features/message-bridge.js
```diff
@@ -0,0 +1,168 @@
...
+            if (!request) return;
+            this.proxyRequest(request, appendToken);
+        });
+        window.addEventListener(appendToken(SubscriptionRequest.NAME), (/** @type {CustomEvent<unknown>} */ e) => {
+            console.log(`[isolated:incoming] ${SubscriptionRequest.NAME}`, e.detail);
+            const subscription = SubscriptionRequest.create(e.detail);
+            if (!subscription) return;
+            this.proxySubscription(subscription, appendToken);
+        });
+        window.addEventListener(appendToken(SubscriptionUnsubscribe.NAME), (/** @type {CustomEvent<unknown>} */ e) => {
```
jonathanKingston, line 54: These listeners all look the same, can you make a method to dedupe?

injected/src/features/message-bridge/create-page-world-bridge.js
```diff
@@ -0,0 +1,134 @@
...
+};
+
+/**
+ * This part can be called from any script.
+ * @param {string} featureName
+ * @param {string} token
+ * @return {Pick<import("@duckduckgo/messaging").Messaging, 'notify' | 'subscribe' | 'request'>}
+ */
+export function createPageWorldBridge(featureName, token) {
+    console.log('will createPageWorldBridge for', featureName, 'with token', token);
```
jonathanKingston, line 31: Again add this to a debug method, this method could be hooked as it is and steal the token. I do think there's value in obfuscating the token a little and obvious ways to capture it should be avoided.

injected/src/features/message-bridge/create-page-world-bridge.js
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    // first, install the proxy
+    send(new InstallProxy({ featureName }));
+
+    /**
+     * return the `Messaging` interface for consumers
+     * Note:
+     */
+    return {
```
jonathanKingston, line 61: Can this be a class definition to make the code more left bound?

injected/src/features/message-bridge.js
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    /**
+     * @param {SubscriptionUnsubscribe} unsubscribe
+     */
+    removeSubscription(unsubscribe) {
+        const { id } = unsubscribe;
+        const fn = this.subscriptions.get(id);
+        console.log(`[isolated] will remove subscription`, id);
+        fn?.();
```
jonathanKingston, line 152: Is this indended?

injected/src/features/message-bridge.js
```diff
@@ -0,0 +1,168 @@
...
+        this.subscriptions.set(
+            id,
+            proxy.subscribe(subscriptionName, (data) => {
+                const responseEvent = new SubscriptionResponse({
+                    subscriptionName,
+                    featureName,
+                    params: data,
+                    id,
+                });
+                const evt = new CustomEvent(responseName, { detail: responseEvent });
```
jonathanKingston, line 138: Does this not need the wrapped name also?

jonathanKingston: Maybe make a central dispatch method rather than calling createEvent and dispatch in many places?

injected/src/features/broker-protection/utils.js
shakyShane: yep will move them both

injected/src/features/message-bridge.js
muodov: I kinda like that it's not async. It's conceptually "pass on and return immediately", and we're not using the result anyways. If we need it in the future, we can easily add a result.

injected/src/features/message-bridge.js
muodov: agreed that a central dispatch method sounds nice.

injected/src/features/message-bridge.js
muodov: I think we need this to actually unsubscribe in the messaging code, don't we?

injected/src/features/message-bridge/create-page-world-bridge.js
muodov: I don't care about style here, but having an explicit type for this (instead of `Pick<...>`) would be neat.

injected/entry-points/apple.js
muodov: I'm not a fan of patches in the base config.. Have you considered mapping feature -> domains instead of domain -> features? I assume that the number of features will always be limited, but the number of domains may grow, so it could be a more compact structure:

```javascript
"messageBridge": {
  "exceptions": [],
  "state": "enabled",
  "settings": {
    "features": {
      // we could type keys here to be a subset of the top-level "features"
      "aiChat": {
        "allowedOn": ["ddgshane.ngrok.io", "duckduckgo.com"]
      },
      "duckPlayer": {
        "allowedOn": ["ddgshane.ngrok.io", "duck.ai"]
      }
    }
  }
}
```

muodov (commented):
The code looks good! I just want to make sure that we give the config structure enough time.

injected/entry-points/apple.js
shakyShane: @muodov - I chose that format because it's already fully supports domain-restrictions, meaning my code is just this: `this.getFeatureSettings('allowed')`. Wouldn't this format mean I need to do the checks again?

@jonathanKingston  is there a way to have best of both worlds? FWIW I honestly have zero preference on this, and I very rarely work with config stuff - so if either of you have strong preferences then I'm happy to see them in action.

injected/src/features/message-bridge.js
shakyShane: yep exactly, I renamed it to make it clearer.

injected/src/features/message-bridge/create-page-world-bridge.js
```diff
@@ -0,0 +1,134 @@
...
+            send(
+                new ProxyNotification({
+                    method,
+                    params,
+                    featureName,
+                }),
+            );
+        },
+        request(method, params) {
+            // todo: use crypto + captured, this is just a demo
```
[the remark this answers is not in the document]
shakyShane: Not sure should we even both with crypto? (genuine question)

injected/src/features/message-bridge/create-page-world-bridge.js
```diff
@@ -0,0 +1,134 @@
...
+                    if (response && response.id === id) {
+                        if ('error' in response && response.error) {
+                            reject(new Error(response.error.message));
+                        } else if ('result' in response) {
+                            resolve(response.result);
+                        }
+                        captured.removeEventListener(responseName, handler);
+                    }
+                };
+                captured.addEventListener(responseName, handler, { once: true });
```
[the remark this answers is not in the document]
shakyShane: Actually that was a bug too, thanks.

injected/src/features/message-bridge/create-page-world-bridge.js
```diff
@@ -9,126 +10,146 @@ import {
...
+    /**
+     * @param {object} params
+     * @param {string} params.featureName
+     * @param {(evt: {name: string} & Record<string, any>) => void} params.send
+     * @param {(s: string) => string} params.appendToken
+     */
+    constructor({ featureName, send, appendToken }) {
+        this.featureName = featureName;
+        this.send = send;
+        this.appendToken = appendToken;
```
muodov, line 64: hm, should we not expose this to the consumers?

injected/src/features/message-bridge.js
```diff
@@ -10,56 +11,83 @@ import {
...
+        };
+
+        /**
+         * These are all the messages we accept from the page-world.
+         */
+        accept(InstallProxy, (install) => this.installProxyFor(install.featureName, args.messagingConfig));
+        accept(ProxyNotification, (notification) => this.proxyNotification(notification));
+        accept(ProxyRequest, (request) => this.proxyRequest(request, send));
+        accept(SubscriptionRequest, (subscription) => this.proxySubscription(subscription, send));
+        accept(SubscriptionUnsubscribe, (unsubscribe) => this.removeSubscription(unsubscribe.id));
```
muodov, line 84: As far as I remember, `init()` is called asynchronously (not immediately on document start). I see that navigator interface is injected early on "privileged domains", which may create a race condition (page tries to use messaging before the proxy is ready).

Should we move this initialization to `load()`?

muodov: or, maybe we should defer injecting the navigator interface until the proxy is ready?

injected/entry-points/apple.js
jonathanKingston: > is there a way to have best of both worlds?

I think we did support a domains method once, can't remember.
Why invent a new pattern though?

injected/integration-test/test-pages/bridge/index.html
```diff
@@ -0,0 +1,58 @@
...
+    window.dispatchEvent = new Proxy(window.dispatchEvent, {
+        apply(target, thisArg, argArray) {
+            console.log('💣, dispatchEvent');
+            return Reflect.apply(target, thisArg, argArray)
+        }
+    })
+
+    globalThis.CustomEvent = new Proxy(globalThis.CustomEvent, {
+        construct(target, argArray, newTarget) {
+            console.log('💣, CustomEvent constructor');
```
jonathanKingston, line 30: Does this get tested at all?

injected/entry-points/apple.js
```diff
@@ -8,7 +8,41 @@ import { WebkitMessagingConfig, TestTransportConfig } from '../../messaging/inde
...
+    // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
+    const userPreferences = $USER_PREFERENCES$;
+
+    const processedConfig = processConfig(config, userUnprotectedDomains, userPreferences);
+
+    /**
+     * This is what we'll add on each native side. The name was chosen
+     * to match what Android already use
+     */
+    processedConfig.messageSecret = 'ABC-DEF';
```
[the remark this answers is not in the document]
shakyShane: yeah

injected/integration-test/test-pages/bridge/index.html
shakyShane: not yet, thinking today about how to test this fully - might remove these.

injected/src/features/message-bridge.js
shakyShane: Yeah I'll verify all these things today

injected/entry-points/apple.js
shakyShane: will move to an asana discussion

injected/src/features/message-bridge/create-page-world-bridge.js
shakyShane: Yeah good point actually - I'll move to function + closure to hide stuff

injected/src/features/message-bridge.js
muodov: Is it also true for the extension?
IIRC full config is not available at `load` stage, so feature might not be disable-able, so if you move, pls double check that

injected/entry-points/apple.js
jonathanKingston: `matchDomainFeatureSetting` is the alternate tbh.

injected/src/features/message-bridge/create-page-world-bridge.js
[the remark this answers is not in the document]
jonathanKingston: I think we shouldn't support insecure contexts.

We have a UUID method anyway right?

injected/entry-points/apple.js
shakyShane: yeah this works 

```
    config.features.messageBridge = {
        exceptions: [],
        state: 'enabled',
        settings: {
            allowed: [
                {
                    domain: 'ddgshane.ngrok.io',
                    features: ['aiChat', 'duckPlayer'],
                },
            ],
        },
    };
```

injected/src/features/message-bridge.js
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * Installing a feature proxy is the act of creating a fresh instance of 'Messaging', but
+     * using the same underlying transport
+     *
+     * @param {string} featureName
+     * @param {import('@duckduckgo/messaging').MessagingConfig} config
+     */
+    installProxyFor(featureName, config) {
+        if (this.proxies.has(featureName)) return;
+        const allowed = this.getFeatureSetting('allowed');
```
jonathanKingston, line 98: Is this changing shape?

injected/integration-test/page-objects/results-collector.js
shakyShane, line 1: We can move all other test setup logic to use this too - but I didn't want to blow up the PR, so for now it's just the messageBridge feature that uses it. The main purpose is to give easy-access to the in-page tests along with built artifacts.

injected/integration-test/test-pages/shared/utils.js
```diff
@@ -81,7 +81,7 @@ async function renderResults() {
     }
     // @ts-expect-error - ongoingTests is not defined in the type definition
     for (const test of window.ongoingTests) {
-        const result = await test.test();
+        const result = await test.test().catch((e) => console.error(`${test.name} threw`, e));
```
shakyShane, line 84: exceptions from the in-page test runner were being swallowed, this just exposes them until there's a more advances runner.

injected/src/features/navigator-interface.js
```diff
@@ -28,6 +29,15 @@ export default class NavigatorInterface extends ContentFeature {
...
                         return DDGPromise.resolve(true);
                     },
+                    /**
+                     * @import { MessagingInterface } from "./message-bridge/schema.js"
+                     * @param {string} featureName
+                     * @return {MessagingInterface}
+                     * @throws {Error}
+                     */
+                    createMessageBridge(featureName) {
+                        return createPageWorldBridge(featureName, args.messageSecret);
```
shakyShane, line 39: this `messageSecret` matches what android call theres - in the macOS PR I've added it (with the lifespan of a tab)

injected/entry-points/apple.js
```diff
@@ -2,13 +2,20 @@
...
     // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
-    const processedConfig = processConfig($CONTENT_SCOPE$, $USER_UNPROTECTED_DOMAINS$, $USER_PREFERENCES$);
+    const config = $CONTENT_SCOPE$;
+    // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
+    const userUnprotectedDomains = $USER_UNPROTECTED_DOMAINS$;
+    // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
+    const userPreferences = $USER_PREFERENCES$;
+
+    const processedConfig = processConfig(config, userUnprotectedDomains, userPreferences, alwaysEnabledFeatures);
+
```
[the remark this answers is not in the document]
jonathanKingston: Appreciate that!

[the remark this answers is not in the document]
jonathanKingston: (I think all the other files do the same btw)

injected/entry-points/apple.js
```diff
@@ -2,13 +2,20 @@
...
 function initCode() {
     // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
-    const processedConfig = processConfig($CONTENT_SCOPE$, $USER_UNPROTECTED_DOMAINS$, $USER_PREFERENCES$);
+    const config = $CONTENT_SCOPE$;
+    // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
+    const userUnprotectedDomains = $USER_UNPROTECTED_DOMAINS$;
+    // @ts-expect-error https://app.asana.com/0/1201614831475344/1203979574128023/f
+    const userPreferences = $USER_PREFERENCES$;
+
+    const processedConfig = processConfig(config, userUnprotectedDomains, userPreferences, alwaysEnabledFeatures);
```
jonathanKingston, line 17: On the `processConfig` call site it's called `platformSpecificFeatures`, I don't really know why we pass it in anyway. I think we could just use it directly within utils.js to simplify the signature. (happy to be a follow up but that argument is named wrongly now).

injected/src/features/message-bridge/create-page-world-bridge.js
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    return createMessagingInterface(featureName, send, appendToken);
+}
+
+/**
+ * We are executing exclusively in secure contexts, so this should never fail
+ */
+function random() {
+    return captured.randomUUID?.() || 'n/a';
```
jonathanKingston, line 100: Can we just make 'n/a' a never type or something. It seems weird to pacify the type system like this tbh. Optionally we could just throw as it should never happen as I understand it.

shakyShane: yep will fix

injected/entry-points/apple.js
shakyShane: hmm, actually `computeEnabledFeatures` is called from a few other places - I think we should actually do a rename for now to reduce the functional changes in the PR

injected/src/features/message-bridge/create-page-world-bridge.js
shakyShane: Fixed

injected/entry-points/apple.js
shakyShane: I renamed it to `platformSpecificFeatures` for now.

injected/integration-test/message-bridge-apple.spec.js
```diff
@@ -0,0 +1,107 @@
...
+
+    // add user preferences to both
+    isolated.withUserPreferences({ messageSecret: 'ABC' });
+    pageWorld.withUserPreferences({ messageSecret: 'ABC' });
+
+    return { pageWorld, isolated };
+}
+
+test('message bridge when enabled (apple)', async ({ page }, testInfo) => {
+    // page.on('console', (msg) => console.log(msg.text()));
```
jonathanKingston, line 32:
```suggestion
```

injected/integration-test/message-bridge-apple.spec.js
```diff
@@ -0,0 +1,107 @@
...
+    });
+
+    if (!('id' in request)) throw new Error('unreachable');
+
+    expect(typeof request.id).toBe('string');
+    expect(request.id.length).toBeGreaterThan(10);
+});
+
+test('message bridge when disabled (apple)', async ({ page }, testInfo) => {
+    // page.on('console', (msg) => console.log(msg.text()));
```
jonathanKingston, line 89:
```suggestion
```

injected/integration-test/page-objects/results-collector.js
```diff
@@ -0,0 +1,200 @@
...
+    mockWebkitMessaging,
+    mockWindowsMessaging,
+    simulateSubscriptionMessage,
+    wrapWebkitScripts,
+    wrapWindowsScripts,
+} from '@duckduckgo/messaging/lib/test-utils.mjs';
+import { perPlatform } from '../type-helpers.mjs';
+
+/**
+ * This is designed to allow you to execute Playwright tests using the various
```
jonathanKingston, line 13: Why do we keep copying this boilerplate over and over? I get that there are subtle differences but overall this is a lot of code for how small the actual test is.

Can we file an issue to clean this up at the minimum?

shakyShane:
I commented on it, maybe it got lost when I pushed b more code?
Basically this the version that will allow all the others to be cleaned
up, so it’s a first step in that direction, I’ll fast follow with
removing all the rest. 

On November 15, 2024, GitHub notifications ***@***.***>
wrote:
> @jonathanKingston commented on this pull request.
>

> In injected/integration-test/page-objects/results-collector.js
> <https://github.com/duckduckgo/content-scope-
> scripts/pull/1213#discussion_r1844482796>:
>
> > @@ -0,0 +1,200 @@ +import { readFileSync } from 'fs'; +import { +
> mockAndroidMessaging, + mockWebkitMessaging, + mockWindowsMessaging, +
> simulateSubscriptionMessage, + wrapWebkitScripts, +
> wrapWindowsScripts, +} from ***@***.***/messaging/lib/test-
> utils.mjs'; +import { perPlatform } from '../type-helpers.mjs'; + +/**
> + * This is designed to allow you to execute Playwright tests using
> the various
> Why do we keep copying this boilerplate over and over? I get that
> there are subtle differences but overall this is a lot of code for how
> small the actual test is.
>
> Can we file an issue to clean this up at the minimum?
>

—
Reply to this email directly, view it on GitHub
<https://github.com/duckduckgo/content-scope-
scripts/pull/1213#pullrequestreview-2439646350>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAMRIAR2FQBYQJ3ICEVFQ6T2AZOE3AVCNFSM6AAAAABRSIOBKKVHI2DSMVQWIX3LMV43YUDVNRWFEZLROVSXG5CSMV3GSZLXHMZDIMZZGY2DMMZVGA>.
You are receiving this because you were mentioned.Message ID:
<duckduckgo/content-scope-
***@***.***>

injected/integration-test/page-objects/results-collector.js
jonathanKingston: Like the original point was pages.spec.js should handle all the tests and instead we've copied that into this file, web-compat and DBP etc.

It's 200 loc boilerplate we've copied per ~100 loc test code.

jonathanKingston:
> I’ll fast follow with removing all the rest. 

Cool yeah do we need to rethink them all being mapped into that pages test? Happy to cover in a follow up anyway :)

shakyShane:
> > I’ll fast follow with removing all the rest. 
> 
> 
> 
> Cool yeah do we need to rethink them all being mapped into that pages test? Happy to cover in a follow up anyway :)

Yeah, we should just have 2 boiler plates. Extension setup + all the rest. I actually have a branch with this all sorted so I will bring it back to life now that this is getting worse.

shakyShane:
I'll merge this, and follow immediately with boilerplate removal where possible.

PR #1261, opened 2024-11-22
Allow for optional types

brianhall:
**Asana Task/Github Issue:** https://app.asana.com/0/608920331025329/1208592440863093/f

## Description

Allows element types (specifically, middle name) to be marked as optional so that we don’t throw an error when trying to fill a form using `dataSource: userProfile`, but where the user hasn’t set a middle name.

Screenshot 2024-11-28 at 17 58 09

## Testing Steps

- Integration tests, check 

## Checklist

*Please tick all that apply:*


injected/src/features/broker-protection/actions/fill-form.js
```diff
@@ -110,6 +113,20 @@ export function fillMany(root, elements, data) {
...
+ * @returns Boolean
+ */
+function isElementTypeOptional(type) {
+    if (type === 'middleName') {
+        return true;
+    }
+
+    return false;
+}
+
```
shakyShane, line 129: Did you consider having this optionality encoded in the json? Like I see things like 'min' 'max'? could it also supported 'required' or 'optional'?

Or, was there a specific reason to not alter the JSON? (just curious)

brianhall:
Fair question! No particular reason not to alter the json. My feeling is that this should be a system-wide (vs. per-broker) setting since middle name is always optional in the user profile form.

We haven’t had any brokers to date that support middle name in a form field (vs. url), hence why this hadn’t come up.

Definitely happy to consider other suggestions if you think of an alternative.

injected/src/features/broker-protection/actions/fill-form.js
brianhall: Fair question! No particular reason not to alter the json. My feeling is that this should be a system-wide (vs. per-broker) setting since middle name is always optional in the user profile form.

We haven’t had any brokers to date that support middle name in a form field (vs. url), hence why this hadn’t come up.

Definitely happy to consider other suggestions if you think of an alternative.

shakyShane: how about in the script that generates the JSON for `fillForm`, we can add `optional: ['middleName']`? a bit like the opposite to how JSON Schema has `required: string[]`. 

Then in that 'else' block you can just check if the key that failed is present in the 'optional' list? somethink like that will keep knowledge out of the JS and be easier to add to later?

brianhall: We've actually deprecated the generator in favor of just editing the production JSON files directly, so there’s no place to generate a field. I could manually add it to each JSON file, but it still feels weird because the middleName being optional isn’t a per-broker configuration.

I think the ideal solution would be to have a list of optional/required fields sent at the layer that enforces this on profile entry (in this case, the native layer) alongside the steps, but that grows the scope here by quite a bit.

Propose I start that conversation with the native folks in Asana, and we merge this change in the interim. What do you think?

shakyShane (approved):

PR #1269, opened 2024-11-25
ntp: virtual list for favorites

shakyShane:
**Asana Task/Github Issue:** https://app.asana.com/0/1201141132935289/1208792034336660/f

## Description

- [x] only render the rows of favorites that are visible
  - this addresses a major pain-point from our native desktop browsers
- [x] when expanded, defer rendering children until the main thread is free
  - this means users on the very slowest devices still get an instant feedback on click 

## Example
Part of this PR is optimizing the expand/collapse action of the favorites list. On slow devices this can cause over a second of input delay since the act of rendering a screen full of favorites can be an expensive main-thread-blocking operation.

To combat this problem, I separated the expand/collapse action from the rendering of the actual tiles, which had the following impact: 💪🏻 

**BEFORE** (20x CPU slowdown)
Screenshot 2024-11-26 at 10 13 02 AM

**AFTER** (20x CPU slowdown)

Screenshot 2024-11-26 at 10 22 32 AM

## TODO:
- [x] add some perf metrics to explain the change

## Testing Steps

- https://deploy-preview-1269--content-scope-scripts.netlify.app/build/pages/new-tab/?favorites=100
- Verify that the page works as expected with showing/hiding along with drag+drop

## Checklist

*Please tick all that apply:*

### DIFF
--- a/special-pages/pages/new-tab/app/favorites/components/Favorites.js
+++ b/special-pages/pages/new-tab/app/favorites/components/Favorites.js
@@ -170,15 +170,17 @@ function VirtualizedGridRows({ WIDGET_ID, rowHeight, favorites, expansion, openF
     const subsetOfRowsToRender = expansion === 'collapsed'
         // if it's collapsed, just 1 row to show (the first one)
         ? [rows[0]]
-        // otherwise select the window between start/end
+        // otherwise, select the window between start/end
         : rows.slice(start, end);
 
     //
-    const inline_height = expansion === 'collapsed' ? rowHeight : rows.length * rowHeight;
+    const container_height = expansion === 'collapsed' ? rowHeight : rows.length * rowHeight;
+    const dropped = document.documentElement.dataset.dropped;
+
     return (
         <div
             className={styles.grid}
-            style={{ height: inline_height + 'px' }}
+            style={{ height: container_height + 'px' }}
             id={WIDGET_ID}
             ref={safeAreaRef}
             onContextMenu={getContextMenuHandler(openContextMenu)}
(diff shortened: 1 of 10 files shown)


special-pages/pages/new-tab/app/favorites/components/Favorites.js
```diff
@@ -30,65 +36,224 @@ export const FavoritesMemo = memo(Favorites);
...
+                add={add}
+                expansion={expansion}
+                openFavorite={openFavorite}
+                openContextMenu={openContextMenu}
+            />
+            {canToggleExpansion && (
+                <div
+                    className={cn({
+                        [styles.showhide]: true,
+                        [styles.showhideVisible]: canToggleExpansion,
```
mgurgel, line 64: Won’t this always be true?

mgurgel (approved):

special-pages/pages/new-tab/app/favorites/components/Favorites.js
shakyShane: no, until you have more than 6 items, it might never be able to show the extra row :)

special-pages/pages/new-tab/app/favorites/components/Favorites.js
```diff
@@ -1,21 +1,28 @@
...
  * @typedef {import('../../../../../types/new-tab.js').Favorite} Favorite
  * @typedef {import('../../../../../types/new-tab.js').FavoritesOpenAction['target']} OpenTarget
  */
 export const FavoritesMemo = memo(Favorites);
+export const ROW_CAPACITY = 6;
+/**
+ * Note: These values MUST match exactly what's defined in the CSS.
+ */
+const ITEM_HEIGHT = 98;
+const ROW_GAP = 8;
```
dbajpeyi, line 25: Any way to tie them together? For me it was not obvious which property in C-S-S? Maybe the comment can mention it.

shakyShane: Yeah we could technically render an item 'offset' so to speak, and measure it. good idea for a follow up for sure.

shakyShane:
I have tested this on Windows ✅

PR #1314, opened 2024-12-11
Remove XRay code and switch to main world content script in Firefox

muodov:
**Asana Task/Github Issue:** 
https://app.asana.com/0/1201614831475344/1207277833550158/f

## Description
Firefox now supports execution worlds for content scripts in their MV2 manifest. This allows us to remove xray code. See also https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/2852

## Testing Steps

- 

## Checklist

*Please tick all that apply:*

### DIFF
--- a/injected/entry-points/mozilla.js
+++ b/injected/entry-points/mozilla.js
@@ -1,96 +1,46 @@
 /**
- * @module Mozilla integration
+ * @module Mozilla MV2 main world integration
  */
 import { load, init, update } from '../src/content-scope-features.js';
 import { isTrackerOrigin } from '../src/trackers';
 import { computeLimitedSiteObject } from '../src/utils.js';
 
-const allowedMessages = [
-    'getClickToLoadState',
-    'getYouTubeVideoDetails',
-    'openShareFeedbackPage',
-    'addDebugFlag',
-    'setYoutubePreviewsEnabled',
-    'unblockClickToLoadContent',
-    'updateYouTubeCTLAddedFlag',
-    'updateFacebookCTLBreakageFlags',
-];
-const messageSecret = randomString();
+const secret = (crypto.getRandomValues(new Uint32Array(1))[0] / 2 ** 32).toString().replace('0.', '');
 
-function randomString() {
-    const num = crypto.getRandomValues(new Uint32Array(1))[0] / 2 ** 32;
-    return num.toString().replace('0.', '');
-}
+const trackerLookup = import.meta.trackerLookup;
 
-function initCode() {
-    const trackerLookup = import.meta.trackerLookup;
-    load({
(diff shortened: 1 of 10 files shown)


injected/src/utils.js
```diff
@@ -418,7 +406,7 @@ export class DDGProxy {
             }
             // The normal return value
             if (isExempt) {
-                return DDGReflect.apply(...args);
+                return DDGReflect.apply(args[0], args[1], args[2]);
```
jonathanKingston, line 409: What's the rationale for this? (if none I wouldn't include it, if there is why not the same for the proxyObject version?)

muodov: There was a weird Typescript lint error here

injected/entry-points/mozilla.js
```diff
@@ -1,96 +1,45 @@
 /**
- * @module Mozilla integration
+ * @module Mozilla MV2 main world integration
```
jonathanKingston, line 2: Looks like the files are now unified, should we just merge the entry points?

jonathanKingston: Follow up :). The less of these files the better imo.

injected/src/utils.js
jonathanKingston: Weird!

injected/entry-points/mozilla.js
muodov: How about a symlink? So that extension still treats them as separate, but there's no duplication in C-S-S

jonathanKingston: I'd rather not introduce symlinks tbh (consumers of the repo have to have symlinks enabled, managing the link paths are a pain, etc).

We still build 'entry-points' so renaming these files should have no impact on the extension consumption.

injected/src/utils.js
```diff
@@ -22,25 +21,14 @@ export function getInjectionElement() {
...
-    if (hasMozProxies) {
-        style = document.createElement('link');
-        style.href = 'data:text/css,' + encodeURIComponent(css);
-        style.setAttribute('rel', 'stylesheet');
-        style.setAttribute('type', 'text/css');
-    } else {
-        style = document.createElement('style');
-        style.innerText = css;
-    }
+    const style = document.createElement('style');
```
jonathanKingston, line 30: Did you check these style elements injected against strict CSP sites? (If it works we should change the function description comment).

I think it'd be probably safer to keep this in, instead checking for the platform name instead.

injected/entry-points/mozilla.js
muodov: I merged the entry points

jonathanKingston (approved):

PR #1315, opened 2024-12-12
NTP: FE - Freemium PIR Banner

vkraucunas:
**Asana Task/Github Issue:** https://app.asana.com/0/72649045549333/1208964501989233/f

## Description
- Adds a Freemium PIR banner, leveraging the RMF component styles as a start

## Testing Steps

- 

## Checklist

*Please tick all that apply:*

### DIFF
--- a/special-pages/pages/new-tab/app/freemium-pir-banner/components/FreemiumPIRBanner.js
+++ b/special-pages/pages/new-tab/app/freemium-pir-banner/components/FreemiumPIRBanner.js
@@ -1,0 +1,37 @@
+import { h } from 'preact';
+import { useContext } from 'preact/hooks';
+import { MessageBar } from '../../components/MessageBar';
+import { FreemiumPIRBannerContext } from '../FreemiumPIRBannerProvider.js';
+
+/**
+ * @typedef { import('../../components/MessageBar').MessageBarMessage} MessageBarMessage
+ * @import { FreemiumPIRBannerMessage } from "../../../types/new-tab"
+ * @param {object} props
+ * @param {FreemiumPIRBannerMessage} props.message
+ * @param {(id: string) => void} props.dismiss
+ * @param {(id: string) => void} [props.action]
+ */
+
+export function FreemiumPIRBanner({ message, action, dismiss }) {
+    /** @type {MessageBarMessage} msgBarMessage */
+    const msgBarMessage = {
+        id: message?.id,
+        messageType: message?.messageType,
+        icon: 'Information-Remover-96',
+        titleText: message?.titleText ? message?.titleText : undefined,
+        descriptionText: message?.descriptionText,
+        primaryActionText: message?.actionText,
+    };
+
+    return <MessageBar message={msgBarMessage} primaryAction={action} dismiss={dismiss} />;
+}
+
+export function FreemiumPIRBannerConsumer() {
+    const { state, action, dismiss } = useContext(FreemiumPIRBannerContext);
+
+    // `state.data.content` can be empty - meaning there's no message to display!
+    if (state.status === 'ready' && state.data?.content) {
+        return <FreemiumPIRBanner message={state.data.content} action={action} dismiss={dismiss} />;
+    }
+    return null;
+}
--- a/special-pages/pages/new-tab/app/freemium-pir-banner/components/FreemiumPIRBanner.js
+++ b/special-pages/pages/new-tab/app/freemium-pir-banner/components/FreemiumPIRBanner.js
@@ -5,7 +5,7 @@ import { FreemiumPIRBannerContext } from '../FreemiumPIRBannerProvider.js';
 
 /**
  * @typedef { import('../../components/MessageBar').MessageBarMessage} MessageBarMessage
- * @import { FreemiumPIRBannerMessage } from "../../../types/new-tab"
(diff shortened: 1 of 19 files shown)


special-pages/pages/new-tab/messages/types/freemiumPIRBanner-message.json
```diff
@@ -0,0 +1,42 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Freemium PIR Banner Data",
+    "description": "The 'content' field is optional. Use that fact to show/hide messages",
+    "type": "object",
+    "properties": {
+        "content": {
```
shakyShane, line 7: can you make `content` required, but then have it’s value be `null` or the object. RMF uses an optional content field but that was just something I missed in the review - next steps and update-notification (and some customizer stuff) use the pattern I’m suggesting. Example here https://github.com/duckduckgo/content-scope-scripts/blob/main/special-pages/pages/new-tab/messages/types/next-steps.json#L9

It affects the generated typescript, and in turn the generated docs. we want the field to be present always, but it’s ok to have it’s value be null.

it’s the difference between 

```ts
content?: SomeType
```

and

```ts
content: SomeType | null
```

shakyShane (changes requested):
Very nice work - just a quick request on one of the schemas :)

special-pages/pages/new-tab/app/freemium-pir-banner/components/FreemiumPIRBanner.js
```diff
@@ -0,0 +1,56 @@
...
+
+export function FreemiumPIRBanner({ message, action, dismiss }) {
+    const convertMarkdownToHTML = (markdown) => {
+        // Use a regular expression to find all the words wrapped in **
+        const regex = /\*\*(.*?)\*\*/g;
+
+        // Replace the matched text with the HTML <strong> tags
+        const result = markdown.replace(regex, '<strong>$1</strong>');
+        return result;
+    };
```
shakyShane, line 25: Can you pull this out into a separate file, and then we can write some unit tests against it. Anything we’re going put into `dangerouslySetInnerHTML` should have plenty of test cases to validate that we don’t break on things like XSS

special-pages/pages/new-tab/app/freemium-pir-banner/components/FreemiumPIRBanner.js
```diff
@@ -0,0 +1,56 @@
...
+
+    return (
+        <div id={message?.id} class={cn(styles.root, styles.icon)}>
```
shakyShane, line 28: is `message` truly optional? If typescript is making you put these `?`s everywhere, it’s probably the case that you need to either alter the schema, or guard against it higher up and not let the optionally creep into this component

shakyShane (changes requested):

special-pages/pages/new-tab/app/freemium-pir-banner/components/FreemiumPIRBanner.js
shakyShane: There are some example tests here you can use as reference https://github.com/duckduckgo/content-scope-scripts/blob/main/special-pages/unit-test/translations.mjs#L1-L57

special-pages/pages/new-tab/messages/types/freemiumPIRBanner-message.json
vkraucunas: gotcha. Should we fix RMF or are API contracts already in place for this?

special-pages/pages/new-tab/app/freemium-pir-banner/components/FreemiumPIRBanner.js
vkraucunas: this is related to the comment above about making content required but nullable, instead of optional. and putting the type title at the correct level. is addressed now.

special-pages/package.json
```diff
@@ -9,7 +9,7 @@
     "prebuild": "node types.mjs && node translations.mjs",
     "build": "node index.mjs",
     "build.dev": "npm run build -- --env development",
-    "test-unit": "node --test unit-test/translations.mjs pages/duckplayer/unit-tests/embed-settings.mjs",
```
vkraucunas, line 12: feels like we’ve hit rule of 3 here.

special-pages/package.json
```diff
@@ -9,7 +9,7 @@
...
+    "test-unit": "node --test unit-test/translations.mjs pages/duckplayer/unit-tests/embed-settings.mjs pages/new-tab/app/freemium-pir-banner/unit-tests/utils.spec.mjs",
```
vkraucunas, line 12: feels like we’ve hit rule of 3 here and should probably do something better.

shakyShane (changes requested):
@vkraucunas This will need a quick rebase now I believe, and can you also add something along the lines of https://github.com/duckduckgo/content-scope-scripts/compare/vkraucunas/freemium-pir-banner...shane/freemium-update?expand=1

^ just an example - I don't expect this is an XSS attack area (it's not user input after all), but since it's an area of the system we don't control I'd like us to be cautious with any HTML

shakyShane (approved):

PR #1384, opened 2025-01-10
Support modifying cookies

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/0/1206777341262243/1209124294701647/f

## Description

Support deletion of cookies for use in breakage situations.

## Testing Steps

- Visit privacy-test-pages.site
- Set `document.cookie="keyToBeDeleted=1"` in the console
- Copy the example modify-cookies.json config test file into the browser (jsonblob or similar).
- Reload the page, check the cookie has gone.

## Checklist

*Please tick all that apply:*


injected/integration-test/test-pages/webcompat/pages/modify-cookies.html
```diff
@@ -0,0 +1,44 @@
...
+        // eslint-disable-next-line no-undef
+        test('Only specified cookies should be removed', async () => {
+            const specifiedKey = hasCookie('keyToBeDeleted')
+            const nonexistentKey = hasCookie('nonexistentKey')
+            const otherKey = hasCookie('otherKey')
+
+            return [
+                { name: 'specified cookie entry deleted', result: specifiedKey, expected: false },
+                { name: 'specified cookie entry that is not present on page load', result: nonexistentKey, expected: false },
+                { name: 'other cookie entry untouched', result: otherKey, expected: true }
```
dharb, line 36: Should we add a test that confirms `path` is working as expected?

tsconfig.json
```diff
@@ -29,7 +29,6 @@
     "playwright.config.js",
     "injected",
     "injected/src/globals.d.ts",
-    "injected/src/features/duckplayer/duckplayer-settings.ts",
```
dharb, line 32: Reminder to resolve this and the removed duckplayer-settings.d.ts before merging

dharb (approved):
Changes look good to me, everything seems to be working as expected. Left a couple comments here and one on the corresponding [config PR](https://github.com/duckduckgo/privacy-configuration/pull/2636/files).

tsconfig.json
jonathanKingston: We don't use it, I can trim it out of the PR and make a separate one if you prefer?

injected/integration-test/test-pages/webcompat/pages/modify-cookies.html
jonathanKingston: Added test to check both states with path. I'd rather not add a second sub path test to validate some of the other cases (I don't think we need to validate the visibility of the cookie for example, because the behavior of cookie is path not set when deleting won't delete).

The same is true for domain but that's a bit more problematic as we intend to run these in a few places.

tsconfig.json
dharb: Ah, I thought maybe this wasn't intentional. But if it's intentional then by all means I'm fine with it here.

dharb (approved):
Thanks for adding the test, LGTM 👍

PR #1400, opened 2025-01-14
[Autofill password import] animate export password only once

dbajpeyi:
**Asana Task/Github Issue:** https://app.asana.com/0/1203822806345703/1209143210450862/f

## Description
Whenever there's path change, there's a call to `animateOrTapElement`. This is not a problem for most transitions, but in case of the export button the excessive tapping causes too aggressive UX:
Screen Recording 2025-01-14 at 10 26 57

This PR ads an internal state called `tapOnce` to the element config, to avoid tapping multiple times for certain elements.

## Testing Steps

- 

## Checklist

*Please tick all that apply:*


injected/src/features/autofill-password-import.js
```diff
@@ -176,6 +183,9 @@ export default class AutofillPasswordImport extends ContentFeature {
             this.currentOverlay.remove();
             this.currentOverlay = null;
             document.removeEventListener('scroll', this);
+            if (this.currentElementConfig?.element) {
+                this.#tappedElements.delete(this.currentElementConfig?.element);
```
jonathanKingston, line 187: Should we just clear out all the elements here? (genuine question not telling you to change it)

injected/src/features/autofill-password-import.js
```diff
@@ -50,6 +51,9 @@ export default class AutofillPasswordImport extends ContentFeature {
 
     #domLoaded;
 
+    /** @type {Set<Element>} */
+    #tappedElements = new Set();
```
jonathanKingston, line 55: As this doesn't get enumerated can we use WeakSet to avoid leaks?

injected/src/features/autofill-password-import.js
dbajpeyi: Probably a good idea, although today I won't be able to test it so I'd leave this as a follow-up!

injected/src/features/autofill-password-import.js
dbajpeyi: Sounds reasonable, we only need the values in the set as long as the elements are there.

jonathanKingston (approved):

PR #1423, opened 2025-01-23
Windows support for Malicious website protection

mgurgel:
**Asana Task/Github Issue:** https://app.asana.com/0/72649045549333/1208745889841029/f

## Description

- Malicious site Special Error page for Windows
- Screenshot tests for Special Error pages

Previews:
- [Phishing](https://deploy-preview-1423--content-scope-scripts.netlify.app/build/pages/special-error/?platform=windows&errorId=phishing)
- [Malware](https://deploy-preview-1423--content-scope-scripts.netlify.app/build/pages/special-error/?platform=windows&errorId=malware)

[Figma](https://www.figma.com/design/uP27mEGEaHCI7ZYXAs1815/Safe-Browsing-Warnings?node-id=1347-44219&m=dev) 

## Testing Steps

- In the web preview, check that all styles match Figma, including when Advanced button is clicked
- In native build, check that all buttons work 

## Checklist

*Please tick all that apply:*


RendijsSmukulis:
Some minor nits:

- The radius around the frame seems off: (left is implementation, right is Figma):

image

- The margins/padding for text under "Advanced" seems off:

Impl:
image

Figma:
image

(The latter seems to be an issue when using ?platform=macos as well: 
image)

RendijsSmukulis (changes requested):
Added a few UI issues in a comment - namely the radius of the corners, and the padding for the "Advanced Info" not looking right

mgurgel:
Hey, @RendijsSmukulis 

- The border radius was a recent decision on the macOS Ship Review that hasn't been reflected on the Figma yet [Asana link](https://app.asana.com/0/0/1208858708058948/1208899308409909/f). I'm checking with Thomas whether the change also affects windows. 

- How were you viewing the preview to get the Advanced margin issue? I can't reproduce it myself. This is what I'm seeing in the preview links using a DDG Win browser:

image

image

I’ve recently rebased to main which may have fixed the preview and could explain what you saw earlier. Let me know if you can still reproduce it.

RendijsSmukulis:
(as discussed in MM) the "Advanced margin issue" was caused by "Animate controls and elements inside windows" setting disabled on my machine, which was picked up by `prefers-reduced-motion`  in css. 

The fixed version works great on my machine.

special-pages/pages/special-error/app/components/AdvancedInfo.module.css
```diff
@@ -1,14 +1,17 @@
+.animationContainer {
```
mgurgel, line 1: Fixes a layout bug with with prefers-reduced-motion

special-pages/shared/components/Button/Button.js
```diff
@@ -3,13 +3,19 @@ import classNames from 'classnames';
...
+ * @property {import("preact").ComponentChild} children
+ * @property {import("preact").JSX.MouseEventHandler<EventTarget>} [onClick]
+ * @property {import('preact').ComponentProps<'button'>} [otherProps]
+ */
+
+/**
+ *
+ * @param {ButtonProps} props
+ * @returns
  */
```
shakyShane, line 19: needs cleaning up?

special-pages/pages/special-error/integration-tests/special-error-screenshots.spec.js
```diff
@@ -0,0 +1,132 @@
+/* global process */
+import { expect, test } from '@playwright/test';
+import { SpecialErrorPage } from './special-error';
+
+const maxDiffPixels = 20;
+
+test.describe('screenshots @screenshots', () => {
```
shakyShane, line 7: Does the tag at this level prevent them all running locally?

shakyShane (changes requested):
Question on the amount of screenshot - since we're not verifying in CI yet, is it worth cutting them down to 1 per layout?

for example, phishing-warning-advanced-reduced-motion-macos-darwin.png and 
ssl-expired-cert-reduced-motion-macos-darwin.png - the only difference is text I think? if so, it might be better suited to test those variants with faster more reliable things, like DOM/aria tests.

That could cut down on the amount of screenshots for now until we formalize how we want to capture this stuff going forward. I'm slightly worried that if we do this level of screenshot testing for all features it's going to get into the thousands very quickly.

I think using screenshots to verify a particular layout is great - but please have a look if any of them in this PR can be de-duplicated?

mgurgel:
@shakyShane I’ve cut down the screenshots to the essentials. Thanks for the tip.

shakyShane (approved):

RendijsSmukulis (approved):

PR #1440, opened 2025-01-29
Support elements that only contain textContent, remove commas from the end of addresses

brianhall:
**Asana Task/Github Issue:** 

## Description
When using XPath’s text() node test (e.g. `//strong[contains(text(), 'Locations:')]/following-sibling::text()`), a node is returned that only contains textContent, not innerText.

This change updates the extractor to check for innerText, and fallback to textContent. Also included, a small change to strip out commas at the end of addresses.

## Testing Steps

- Run integration / unit tests
- Drag this version of C-S-S into XCode and run PIR.

## Checklist

*Please tick all that apply:*

### DIFF
--- a/injected/src/features/broker-protection/actions/extract.js
+++ b/injected/src/features/broker-protection/actions/extract.js
@@ -160,15 +160,20 @@ export function createProfile(elementFactory, extractData) {
 }
 
 /**
- * @param {{innerText: string}[]} elements
+ * @param {(HTMLElement | { innerText: string })[]} elements
  * @param {string} key
  * @param {ExtractProfileProperty} extractField
  * @return {string[]}
  */
 function stringValuesFromElements(elements, key, extractField) {
     return elements.map((element) => {
-        // todo: should we use textContent here?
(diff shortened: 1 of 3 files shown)


injected/src/features/broker-protection/actions/extract.js
```diff
@@ -160,15 +160,26 @@ export function createProfile(elementFactory, extractData) {
...
+
+        if ('innerText' in element) {
+            elementValue = rules[key]?.(element) ?? element?.innerText ?? null;
+
+            // In instances where we use the text() node test, innerText will be undefined, and we fall back to textContent
+        } else if ('textContent' in element) {
+            elementValue = rules[key]?.(element) ?? element?.textContent ?? null;
+        }
+
+        if (!elementValue) {
```
brianhall, line 180: I noticed that in cases where elementValue was null, we were throwing errors as `removeCommonSuffixesAndPrefixes` expects non-null input.

shakyShane (approved):

PR #1444, opened 2025-01-30
Allow click actions to fail silently with an additional parameter

brianhall:
**Asana Task/Github Issue:** https://app.asana.com/0/608920331025329/1209274402556155/f 

## Description

Occasionally we need click actions to fail silently without throwing an error. A real life example: on https://www.privaterecords.net we need to click links on each record to expand the addresses / relatives to make matches most accurate, however sometimes these links won’t exist (for instance, if there are no results).

We don’t want to fail, because we want the extract action (with the noResultsFound directive) to run so we can mark a profile as “not found”.

## Testing Steps

- Integration tests
- Drag/drop C-S-S into XCode

## Checklist

*Please tick all that apply:*

### DIFF
--- a/special-pages/pages/new-tab/app/customizer/strings.json
+++ b/special-pages/pages/new-tab/app/customizer/strings.json
@@ -9,15 +9,15 @@
   },
   "customizer_section_title_background": {
     "title": "Background",
-    "note": ""
+    "note": "Section title displayed in the UI for customizing the background."
   },
   "customizer_section_title_theme": {
     "title": "Browser Theme",
-    "note": ""
+    "note": "Section title for a section where users can customize the browser theme. (for example, 'light' or 'dark')"
   },
   "customizer_section_title_sections": {
     "title": "Sections",
-    "note": ""
+    "note": "Section title for a list of toggles for showing/hiding certain sections of the page."
   },
   "customizer_browser_theme_light": {
     "title": "Light",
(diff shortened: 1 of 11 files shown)


injected/src/features/broker-protection/actions/click.js
```diff
@@ -51,6 +51,10 @@ export function click(action, userData, root = document) {
         const elements = getElements(rootElement, element.selector);
 
         if (!elements?.length) {
+            if (element.failSilently) {
```
shakyShane, line 54: @brianhall - just a quick one on naming - when I saw ‘failSilently’ I was expecting something like ‘do nothing in this case’, like an empty `catch` block or something. 

But in both these cases, would `optional: true` be a better way to express it?

then it reads: `if no elements found, but this was optional anyway, return a success` and `if the element is disabled and it’s not optional, return an error`

brianhall: @shakyShane I like that naming better, however we’ve got failsilently in use [elsewhere](https://github.com/duckduckgo/content-scope-scripts/blob/c6b67bbe92b055ea0a44e8bdeca618d3e6627a19/injected/src/features/broker-protection/actions/expectation.js#L16), so probably best to standardize for now. Will definitely consider it for future refactors.

shakyShane (approved):

PR #1550, opened 2025-03-07
feat(pir): captcha providers

madblex:
**Asana Task/Github Issue:** https://app.asana.com/0/0/1209555302496203

- tech design: https://app.asana.com/0/0/1209498481524623

## Description

This PR:
- refactors the captcha action handlers to enable extensibility (see associated [tech design](https://app.asana.com/0/0/1209498481524623))
- adds support for existing captcha providers: recaptcha, recaptcha enterprise, captcha (partially)
- adds scaffolding for cloudflare turnstile
- adds support for the new `navigate` action, already augmented with the new `injectCaptchaHandler` property (see associated [tech design](https://app.asana.com/0/1206054067224240/1209329834965895))
- adds `resolveActionHandlers` resolver for the action handlers that uses a remote config to determine which version is used 

TODO:
- [x] polish up the code, I am still deciding upon some patterns I used, names, etc
- [x] create remote config and wire up the value in `resolveActionHandlers`
- [x] manual testing
- [x] automated testing
- [ ] reference asana tasks next to TODOs

## Testing Steps

- `cd injected`
- `npm run playwright` should pass

## Checklist

*Please tick all that apply:*


injected/src/features/broker-protection/captcha-services/utils/index.js
```diff
@@ -0,0 +1,3 @@
+export * from './sitekey';
+export * from './token';
+export * from './stringify-function';
```
shakyShane, line 3: As you already saw, these are not easy to use in node, and they can mask the true dependency chain of a single file.

We want to be deliberate with imports/exports - even though it's more verbose, it prevents accidental inclusion of code. Bundlers are not smart enough to completely mitigate this yet :(

injected/src/features/broker-protection/captcha-services/captcha.service.js
```diff
@@ -0,0 +1,87 @@
...
+        const response = {
+            url: removeUrlQueryParams(window.location.href), // query params (which may include PII)
+            siteKey: captchaIdentifier,
+            type: captchaProvider.getType(),
+        };
+        return new SuccessResponse({ actionID, actionType, response });
+    } catch (e) {
+        return new ErrorResponse({ actionID, message: `[getCaptchaInfo] ${e.message}` });
+    }
+};
```
shakyShane, line 56: In line with the other comments, we can avoid using throw/try/catch as control flow. In this example, if you want to avoid the duplication, stick a function inline to make it less verbose?

```js
/**
 * Gets the captcha information to send to the backend
 *
 * @param {import('../types.js').PirAction} action
 * @param {Document | HTMLElement} root
 */
export const getCaptchaInfo = (action, root = document) => {
    const { id: actionID, selector, actionType, captchaType } = action;
    const error = (message) => new ErrorResponse({ actionID, message: `[getCaptchaInfo] ${message}` });

    if (!selector) {
        return error('missing selector');
    }

    const captchaContainer = getElement(root, selector);
    if (!captchaContainer) {
        return error(`could not find captchaContainer with selector ${selector}`);
    }

    const captchaProvider = getCaptchaProvider(captchaContainer, captchaType);
    const captchaIdentifier = captchaProvider.getCaptchaIdentifier(captchaContainer);
    if (!captchaIdentifier) {
        return error(`could not extract captcha identifier from ${captchaType} captcha`);
    }

    const response = {
        url: removeUrlQueryParams(window.location.href), // query params (which may include PII)
        siteKey: captchaIdentifier,
        type: captchaProvider.getType(),
    };

    return new SuccessResponse({ actionID, actionType, response });
};
```

Either way once the `getCaptchaIdentifier` method (and others) are refactored to not throw exceptions, then this pattern above would be fine.

shakyShane (changes requested):
Thanks @madblex for your first contribution! 

I like the pattern you're thinking of here, I just noted some small things related to matching the existing codebase's patterns around type-safety etc. Here's a small summary:

- 1. try to design richer return types to avoid needing to use exceptions as control flow
- 2. favour `implements` over `extends` to avoid class inheritance (this is awkward in JSDoc, but I gave an example in the comments)
- 3. remove 'index' files if they only serve to enable re-exports. In this codebase it's preferred to be explicit about import/exports.
- 4. prefer `function foo() {` over `const foo = () => {` in module-level functions

In terms of number 2, I made this change on a branch just to verify the types would still work correctly, you can take a look here https://github.com/duckduckgo/content-scope-scripts/commit/bf391d26fb8719b8ed751ae161e272a884a502d4

injected/src/features/broker-protection/captcha-services/providers/hcaptcha.js
```diff
@@ -0,0 +1,106 @@
+import { CaptchaProvider } from './provider.interface';
+import { injectTokenIntoElement, stringifyFunction } from '../utils';
+import { getUrlParameter, getUrlHashParameter, getElementByName, getElementWithSrcStart } from '../../utils';
+
+/**
+ * Provider for hCaptcha
+ */
+export class HCaptchaProvider extends CaptchaProvider {
```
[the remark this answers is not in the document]
brianhall: Also, feel free to throw out this hcaptcha provider for now as it's not actually in use.

injected/src/features/broker-protection/actions/click.js
```diff
@@ -1,4 +1,4 @@
-import { getElements } from '../utils.js';
+import { getElements } from '../utils';
```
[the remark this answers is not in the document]
madblex: done, thanks!

injected/src/features/broker-protection/actions/index.js
```diff
@@ -0,0 +1,31 @@
...
+            ? {
+                  navigate,
+                  ...captchaHandlers,
+              }
+            : {
+                  navigate: buildUrl,
+                  ...deprecatedCaptchaHandlers,
+              }),
+    };
+};
```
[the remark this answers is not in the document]
madblex: done, thanks!

injected/src/features/broker-protection/actions/navigate.js
```diff
@@ -0,0 +1,31 @@
...
+        const response = {
+            ...urlResult.success.response,
+            ...(codeToInject && { code: codeToInject }),
+        };
+
+        return new SuccessResponse({ actionID, actionType, response });
+    } catch (e) {
+        return new ErrorResponse({ actionID, message: `[navigate] ${e.message}` });
+    }
+};
```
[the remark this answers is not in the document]
madblex: done, thanks!

injected/src/features/broker-protection/captcha-services/providers/cloudflare-turnstile.js
```diff
@@ -0,0 +1,52 @@
+import { CaptchaProvider } from './provider.interface';
+
+export class CloudFlareTurnstileProvider extends CaptchaProvider {
```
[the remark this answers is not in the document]
madblex: Oh, I had no idea that's how you do it, much better, thanks! Applied

injected/src/features/broker-protection/captcha-services/providers/hcaptcha.js
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * @param {HTMLElement} captchaContainerElement - The element to check
+     * @returns {string|null} - The site key or null if not found
+     * @throws {Error}
+     */
+    getCaptchaIdentifier(captchaContainerElement) {
+        const captchaElement = this._getCaptchaElement(captchaContainerElement);
+        if (!captchaElement) {
+            throw new Error('[getCaptchaIdentifier] could not find captcha');
```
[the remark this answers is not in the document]
madblex: this is not about expressing optionality, it's a valid error

injected/src/features/broker-protection/captcha-services/providers/index.js
```diff
@@ -0,0 +1,14 @@
...
+import { ReCaptchaEnterpriseProvider } from './recaptcha-enterprise';
+import { CloudFlareTurnstileProvider } from './cloudflare-turnstile';
+import { HCaptchaProvider } from './hcaptcha';
+
+const captchaFactory = new CaptchaFactory();
+
+captchaFactory.registerProvider(ReCaptchaV2Provider);
+captchaFactory.registerProvider(ReCaptchaEnterpriseProvider);
+captchaFactory.registerProvider(CloudFlareTurnstileProvider);
+captchaFactory.registerProvider(HCaptchaProvider);
```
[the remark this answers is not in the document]
madblex: Oh that is so much better, applied, thanks!

injected/src/features/broker-protection/captcha-services/utils/index.js
madblex: Yea, I won't get into a debate on this 😅 removed all barrel files

injected/src/features/broker-protection/captcha-services/captcha.service.js
madblex: What exactly is the issue with throw/try/catch? I don't think perf is an issue here.

From a consistency pov, we are not dogmatic about not using it for the rest of the codebases.

Error return objects have some drawbacks, notably:
- verbose error checking, especially with JSdoc that makes condition types more awkward
- error propagation becomes more complex
- it's riskier as we may forget to handle cases, especially in JS where types are almost optional (we do have a catch all however)

I will branch off to outline the first 2 points after my mid-day break

@brianhall @shakyShane

injected/src/features/broker-protection/captcha-services/providers/hcaptcha.js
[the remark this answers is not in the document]
brianhall: I think you still want to return an error type so that it’s picked up on the native side like all other errors

injected/src/features/broker-protection/captcha-services/captcha.service.js
```diff
@@ -0,0 +1,107 @@
...
+ * @param {import('../types.js').PirAction} action
+ * @param {Document | HTMLElement} root
+ * @return {import('../types.js').ActionResponse}
+ */
+export function getCaptchaInfo(action, root = document) {
+    const { id: actionID, selector, actionType, captchaType } = action;
+    try {
+        if (!captchaType) {
+            // ensures backward compatibility with old actions
+            return getCaptchaInfoDeprecated(action, root);
```
madblex, line 44: @brianhall @shakyShane Here, in `getSupportingCodeToInject` above `solveCaptcha` below I ensure the new code is backwards-compatible for safety.

I am still not sure this is needed since C-S-S and JSON are **both** remotely delivered and we control releases. There is no harm in having it, but it will make testing more complex (more flows to test) and it may affect how quickly we can deprecate the code.

LMK what you think; we can discuss.

injected/src/features/broker-protection/captcha-services/providers/hcaptcha.js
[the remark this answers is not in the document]
madblex: that is done by the consumer, see the try/catch in `getCaptchaInfo`

injected/src/features/broker-protection/captcha-services/captcha.service.js
jonathanKingston: @shakyShane wasn't the try/catch flow your proposal.

Largely we don't throw in C-S-S inject but for pir it's somewhat irrelevant what we do. What's the rationale for the change in direction?

shakyShane: no, my proposal was to avoid exceptions as control flow (where possible) since it's impossible to represent in Typescript (this matches the rest of PIR)

injected/src/features/broker-protection/captcha-services/captcha.service.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Takes the solved captcha token and injects it into the page to solve the captcha
+ *
+ * @param {import('../types.js').PirAction} action
+ * @param {string} token
+ * @param {Document} root
+ * @return {import('../types.js').ActionResponse}
+ */
+export function solveCaptcha(action, token, root = document) {
+    const { id: actionID, actionType, captchaType } = action;
```
brianhall, line 82: Just noticed something as I'm create the CloudFlare provider - we pass `selector` in the json here as well. We should consider doing the same as [we're doing here](https://github.com/duckduckgo/content-scope-scripts/pull/1550/files#diff-f8d65f97dd5fc02206720d3d9da5275ffd7edb3b3625e60cb06bba113be89d09R49-R52) to validate the container exists.

Curious if we need access to `root` here for canSolve / injectToken, or if we could get away with passing in `captchaContainer` to the supporting functions like `canSolve` and `injectToken`.

My main use-case is that in `canSolve` I'm trying to validate that attributes exist on the container, and I'd like to avoid doing any DOM lookups and instead just receive the captchaContainer directly. Believe this would work for `injectToken` as well.

madblex: @brianhall happy you brought this up. While porting the implementation I had the same questions.

Before I make the code change + retest, do you see or @shakyShane:
1. See any use for the `root` other than getting the `captchaContainer` using the `selector`?
2. Remember why the selector is not used in the [original implementation](https://github.com/duckduckgo/content-scope-scripts/blob/main/injected/src/features/broker-protection/actions/captcha.js#L91-L120)?

brianhall: 1. I can't come up with a valid use-case for it. Even if the solution were in a different root element (something I've never personally seen), we could pass that with `selector`. 
2. My guess is that the answer here is 🤷, but happy for @shakyShane to weigh in with more context. 😄

shakyShane: 'Root' was added to enable tooling, if that's not needed here (or not being used in general) feel free to omit it.

injected/src/features/broker-protection/captcha-services/providers/hcaptcha.js
```diff
@@ -0,0 +1,109 @@
+import { PirError } from '../../types';
```
brianhall, line 1: Let's just null this one out like cloudflareTurnstile and not invoke it in the registry. I'd like for us to implement this as a follow-up to your initial work. Feel free to just roll to a separate PR and link in Asana.

injected/src/features/broker-protection/captcha-services/providers/registry.js
```diff
@@ -0,0 +1,27 @@
...
+
+captchaFactory.registerProvider(
+    new ReCaptchaProvider({
+        type: 'recaptchaEnterprise',
+        providerUrl: 'https://www.google.com/recaptcha/enterprise',
+        responseElementName: 'g-recaptcha-response',
+    }),
+);
+
+captchaFactory.registerProvider(new CloudFlareTurnstileProvider());
```
brianhall, line 24: For this pass, let's not register cloudFlare or hCaptcha until we test explicitly, just to reduce the surface area.

brianhall (dismissed):
@madblex This looks great to me. I left some comments and did some testing (see below) but nothing blocking from me.

I tested this PR directly in XCode in the following scenarios:

1. Using existing Spokeo JSON file (e.g. no `captchaType` in `getCaptchaInfo`) - ✅ Succeeded as expected
2. Using an invalid `captchaType` in Spokeo's JSON (`captchaType: 'bleh'`) - ❌ Failed as expected 
3. Using an explicit `captchaType` in Spokeo's JSON (`captchaType: 'recaptcha2'`) - ✅ Succeded as expected
4. Using a real but mismatched captchaType in Spokeo's JSON (`captchaType: 'cloudflareTurnstile'`) - ✅ Fell back to recaptcha2 and succeeded.

Very excited to see us in a better place captcha-wise. Great job.

injected/src/features/broker-protection/execute.js
```diff
@@ -1,25 +1,21 @@
...
- * @param {"extract" | "fillForm" | "click" | "expectation" | "getCaptchaInfo" | "solveCaptcha" | "navigate"} action.actionType
+ * @param {import('./types.js').PirAction} action
  * @param {Record<string, any>} inputData
  * @param {Document} [root] - optional root element
  * @return {Promise<import('./types.js').ActionResponse>}
  */
 export async function execute(action, inputData, root = document) {
     try {
+        const { navigate, extract, click, expectation, fillForm, getCaptchaInfo, solveCaptcha } = resolveActionHandlers({
+            useNewActionHandlers: true, // TODO: replace with remote config setting
```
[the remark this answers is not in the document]
madblex: correct, it is addressed in #1560

injected/src/features/broker-protection/captcha-services/get-captcha-container.js
```diff
@@ -0,0 +1,21 @@
...
+        return PirError.create('missing selector');
+    }
+
+    const captchaContainer = getElement(root, selector);
+    if (!captchaContainer) {
+        return PirError.create(`could not find captcha container with selector ${selector}`);
+    }
+
+    return captchaContainer;
+}
```
madblex, line 21: @brianhall both `getCaptchaInfo` and `solveCaptcha` use this now, we then pass down `captchaContainer` to the supporting functions.

I updated all code and interfaces + tested locally with the Mac OS browser on multiple scenarios

madblex:
> @madblex This looks great to me. I left some comments and did some testing (see below) but nothing blocking from me.
> 
> I tested this PR directly in XCode in the following scenarios:
> 
> 1. Using existing Spokeo JSON file (e.g. no `captchaType` in `getCaptchaInfo`) - ✅ Succeeded as expected
> 2. Using an invalid `captchaType` in Spokeo's JSON (`captchaType: 'bleh'`) - ❌ Failed as expected
> 3. Using an explicit `captchaType` in Spokeo's JSON (`captchaType: 'recaptcha2'`) - ✅ Succeded as expected
> 4. Using a real but mismatched captchaType in Spokeo's JSON (`captchaType: 'cloudflareTurnstile'`) - ✅ Fell back to recaptcha2 and succeeded.
> 
> Very excited to see us in a better place captcha-wise. Great job.

Thanks Brian, and thanks for testing as well! Did a very similar test matrix and it all worked well with the new changes. Ready for another round of reviews.

brianhall (dismissed):
These updated changes look great to me.

injected/src/features/broker-protection/captcha-services/providers/provider.interface.js
```diff
@@ -0,0 +1,75 @@
...
+        throw new Error('injectToken() missing implementation');
+    }
+
+    /**
+     * Creates a callback function to execute when the captcha is solved
+     * @abstract
+     * @param {string} token - The solved captcha token
+     * @returns {PirError|string|null} Callback function to execute when the captcha is solved
+     */
+    getSolveCallback(token) {
```
brianhall, line 72: @madblex Sorry, I missed this on my last review. Could we inject captchaContainerElement here as well? I need it for grabbing the callback name from the container and wrapping it.

brianhall: Added the change here, feel free to disregard if needed - I imagine we'll need some testing updates as well: https://github.com/duckduckgo/content-scope-scripts/pull/1576

madblex: The alternative is to use a setter/initialiser method and remove the `captchaContainerElement` from all methods. I think we can explore something like that in future iterations if we find ourselves changing common/unrelated code often. For now, I'd say what you did in #1576 is appropriate.

 @brianhall Are you fine with me merging your change directly or you'd like to add it as part of your work?

brianhall: Feel free to merge, thanks!

brianhall (dismissed):

shakyShane (approved):

brianhall (approved):

shakyShane (approved):

PR #1561, opened 2025-03-14
add favicon feature

shakyShane:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1201141132935289/task/1209667199269997?focus=true

## Description

- added a new feature for reporting favicons 
- made the `monitor` part configurable remotely (eg: it can be turned off)
- context: we are replacing the following https://github.com/duckduckgo/apple-browsers/blob/main/macOS/DuckDuckGo/Favicons/Model/FaviconUserScript.swift#L70-L107

## Testing Steps

- docs: https://deploy-preview-1561--content-scope-scripts.netlify.app/documents/favicon_monitor
- Integration tests are added, and further testing required in the macOS PR to be open soon.

## Checklist

*Please tick all that apply:*


shakyShane:
* **#1561** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/1561?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/1561?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

injected/src/features/favicon.js
```diff
@@ -0,0 +1,85 @@
...
+            const delta = currentTime - lastEmitTime;
+            if (delta >= interval) {
+                this.send();
+            } else {
+                trailing = setTimeout(() => {
+                    this.send();
+                }, 50);
+            }
+            lastEmitTime = currentTime;
+        });
```
shakyShane, line 41: this will emit updates at most every 50ms, and it will not lose the last one.

injected/src/features/favicon.js
```diff
@@ -0,0 +1,84 @@
...
+function monitor(changeObservedCallback, target = document.head) {
+    const observer = new MutationObserver((mutations) => {
+        for (const mutation of mutations) {
+            if (mutation.type === 'attributes' && mutation.target instanceof HTMLLinkElement) {
+                changeObservedCallback();
+                break;
+            }
+        }
+    });
+    observer.observe(target, { attributeFilter: ['rel', 'href'], attributes: true, subtree: true });
```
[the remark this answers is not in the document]
shakyShane: Ahh good point, I'll add a new test to cover this case :)

injected/src/features/favicon.js
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+export default Favicon;
+
+/**
+ * @param {()=>void} changeObservedCallback
+ * @param {Element} [target]
+ */
+function monitor(changeObservedCallback, target = document.head) {
```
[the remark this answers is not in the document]
shakyShane: Great point, I'll be more careful here.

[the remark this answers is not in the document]
shakyShane: I think limiting to `head` is smart just because of how noisy. But yeah it might not need to be a param

shakyShane:
@muodov  I think I've resolved your feedback in https://github.com/duckduckgo/content-scope-scripts/pull/1561/commits/483c4727ae7fb6d6638d70d3acdaab363c544bb2

injected/src/features/favicon.js
```diff
@@ -51,18 +53,35 @@ export default Favicon;
...
         for (const mutation of mutations) {
             if (mutation.type === 'attributes' && mutation.target instanceof HTMLLinkElement) {
                 changeObservedCallback();
                 break;
             }
+            if (mutation.type === 'childList') {
+                for (const addedNode of mutation.addedNodes) {
+                    if (addedNode instanceof HTMLLinkElement) {
+                        changeObservedCallback();
+                        break;
```
muodov, line 71: I think you meant to break from the outer loop too?

muodov (approved):
This looks good to me, just a small nit about loop breaking.

injected/src/features/favicon.js
shakyShane: Oh yeah, you're correct! thanks

shakyShane: I'll just return instead

muodov (approved):

injected/src/features/favicon.js
```diff
@@ -0,0 +1,103 @@
+import ContentFeature from '../content-feature.js';
+import { isBeingFramed } from '../utils.js';
+
+export class Favicon extends ContentFeature {
+    init() {
+        if (this.platform.name === 'ios') return;
```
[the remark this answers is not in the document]
shakyShane: yeah we could roll out config at the same time yeah, but since we don't need to for this feature, I was under the impression we could skip it until needed?

If there's still a preference to do via config, then I can of course, no worries either way :)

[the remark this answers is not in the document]
jonathanKingston: Updated the comment to suggest bundle, which we're doing also.

injected/integration-test/test-pages/message-bridge/config/message-bridge-enabled.json
```diff
@@ -5,6 +5,10 @@
       "state": "enabled",
       "exceptions": []
     },
+    "favicon": {
```
jonathanKingston, line 8: Do we need these changes here? It doesn't look like they're being tested are they?

It would be better to validate that "state": "disabled" actually disables the feature somewhere (if you're not already).

injected/src/utils.js
```diff
@@ -703,7 +703,7 @@ export function isGloballyDisabled(args) {
  * @import {FeatureName} from "./features";
  * @type {FeatureName[]}
  */
-export const platformSpecificFeatures = ['windowsPermissionUsage', 'messageBridge'];
+export const platformSpecificFeatures = ['windowsPermissionUsage', 'messageBridge', 'favicon'];
 
 export function isPlatformSpecificFeature(featureName) {
```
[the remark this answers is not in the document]
shakyShane: the 'platformSpecificFeatures' part?

injected/integration-test/test-pages/message-bridge/config/message-bridge-enabled.json
shakyShane: it's a side-effect of our recent change. Without this, the favicon code is now running always - so I can either update the message-bridge tests, to expect additional messages, or disable here.

injected/src/utils.js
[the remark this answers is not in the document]
jonathanKingston: Yeah seems the wrong name now.

injected/integration-test/test-pages/message-bridge/config/message-bridge-enabled.json
jonathanKingston: And if the favicon code runs it breaks the test? Sorry just wasn't clear as a side effect.

shakyShane: it only 'breaks' the test in the sense that I'd have to filter out the message. which feels even more wrong that doing this haha

jonathanKingston (approved):

PR #1564, opened 2025-03-14
allow non-protection features to survive protections toggle

shakyShane:
**Asana Task/Github Issue:** https://app.asana.com/0/1201141132935289/1209713939349679/f

## Description

- [x] apple, android, windows: only call 'load' on features if they are enabled in config
- [x] apple, android, windows: don't return early if protections were toggled
- [x] extension: NO CHANGE: `load` will be called for every feature in the bundle as before

apple, android, windows: 

Untitled (5)

extension:

image

## Testing Steps

- [x] updated tests to show no impact.

## Checklist

*Please tick all that apply:*

### DIFF
--- a/injected/entry-points/integration.js
+++ b/injected/entry-points/integration.js
@@ -26,7 +26,14 @@ function generateConfig() {
             domain: topLevelUrl.hostname,
             isBroken: false,
             allowlisted: false,
-            enabledFeatures: ['fingerprintingCanvas', 'fingerprintingScreenSize', 'navigatorInterface', 'cookie'],
+            enabledFeatures: [
+                'fingerprintingCanvas',
+                'fingerprintingScreenSize',
+                'navigatorInterface',
+                'cookie',
+                'webCompat',
+                'apiManipulation',
+            ],
         },
     };
 }
--- a/injected/entry-points/windows.js
+++ b/injected/entry-points/windows.js
@@ -2,7 +2,7 @@
  * @module Windows integration
  */
 import { load, init } from '../src/content-scope-features.js';
-import { processConfig, isGloballyDisabled, platformSpecificFeatures } from './../src/utils';
+import { processConfig, platformSpecificFeatures } from './../src/utils';
 import { WindowsMessagingConfig } from '../../messaging/index.js';
 
 function initCode() {
@@ -14,9 +14,7 @@ function initCode() {
     const userPreferences = $USER_PREFERENCES$;
 
     const processedConfig = processConfig(config, userUnprotectedDomains, userPreferences, platformSpecificFeatures);
-    if (isGloballyDisabled(processedConfig)) {
-        return;
-    }
+
     processedConfig.messagingConfig = new WindowsMessagingConfig({
         methods: {
             // @ts-expect-error - Type 'unknown' is not assignable to type...
--- a/injected/src/content-scope-features.js
+++ b/injected/src/content-scope-features.js
@@ -43,11 +43,12 @@ export function load(args) {
 
     const bundledFeatureNames = typeof importConfig.injectName === 'string' ? platformSupport[importConfig.injectName] : [];
 
-    // if we're globally disabled, only allow `platformSpecificFeatures`
+    // prettier-ignore
     const featuresToLoad = isGloballyDisabled(args)
+        // if we're globally disabled, only allow `platformSpecificFeatures`
         ? platformSpecificFeatures
-        : // otherwise, use `enabledFeatures` if available, falling back to every feature in the bundle
-          args.site.enabledFeatures || bundledFeatureNames;
+        // otherwise, use `enabledFeatures` if available, falling back to every feature in the bundle
+        : args.site.enabledFeatures || bundledFeatureNames;
 
     for (const featureName of bundledFeatureNames) {
         if (featuresToLoad.includes(featureName)) {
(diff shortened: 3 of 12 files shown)


shakyShane:
* **#1564** <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/1564?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/duckduckgo/content-scope-scripts/1564?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

injected/entry-points/apple-isolated.js
```diff
@@ -0,0 +1,46 @@
...
+    processedConfig.messagingConfig = new WebkitMessagingConfig({
+        webkitMessageHandlerNames: ['contentScopeScriptsIsolated'],
+        secret: '',
+        hasModernWebkitAPI: true,
+    });
+
+    // prettier-ignore
+    const featuresToLoad = isGloballyDisabled(processedConfig)
+        ? featuresExemptFromProtectionToggle
+        : processedConfig.site.enabledFeatures;
```
shakyShane, line 27: just an idea, very much open to others :)

injected/src/content-scope-features.js
```diff
@@ -60,8 +63,7 @@ export async function init(args) {
     }
     registerMessageSecret(args.messageSecret);
     initStringExemptionLists(args);
-    const resolvedFeatures = await Promise.all(features);
```
jonathanKingston, line 63: Was looking at this the other day. This used to be `features.push`'ing a promise I think.

injected/entry-points/apple-isolated.js
dbajpeyi: Shouldn't this be in `processConfig` where `enabledFeatures` are computed? Or do we want to be very explicit about it?

injected/entry-points/integration.js
```diff
@@ -26,7 +26,14 @@ function generateConfig() {
...
             isBroken: false,
             allowlisted: false,
-            enabledFeatures: ['fingerprintingCanvas', 'fingerprintingScreenSize', 'navigatorInterface', 'cookie'],
+            enabledFeatures: [
+                'fingerprintingCanvas',
+                'fingerprintingScreenSize',
+                'navigatorInterface',
+                'cookie',
+                'webCompat',
+                'apiManipulation',
```
shakyShane, line 35: this change is needed to keep the tests passing. In production the extension calls 'load' on every feature.

injected/entry-points/windows.js
```diff
@@ -14,9 +14,7 @@ function initCode() {
     const userPreferences = $USER_PREFERENCES$;
 
     const processedConfig = processConfig(config, userUnprotectedDomains, userPreferences, platformSpecificFeatures);
```
jonathanKingston, line 16: I'm not really sure why we import something from utils to then pass it back in again... but it's also fine not to change this.

In Deeps change, we've tried to move more stuff out of these boilerplate files when it's not needed. I think we can do much more of it and/or almost get rid of them now.

injected/src/utils.js
```diff
@@ -487,6 +487,13 @@ export function isUnprotectedDomain(topLevelHostname, featureList) {
     if (!topLevelHostname) {
         return false;
     }
+
+    // special-case for localhost (eg: testing)
+    if (topLevelHostname === 'localhost') {
```
jonathanKingston, line 492: Is this just a short circuit? Doesn't it work anyway? If not can we make it work rather than have an additional branch for this?

injected/entry-points/windows.js
shakyShane: yeah I can change this tbh

injected/src/content-scope-features.js
```diff
@@ -41,13 +41,22 @@ export function load(args) {
...
-        const ContentFeature = platformFeatures['ddg_feature_' + featureName];
-        const featureInstance = new ContentFeature(featureName, importConfig, args);
-        featureInstance.callLoad();
-        features.push({ featureName, featureInstance });
+    // prettier-ignore
+    const featuresToLoad = isGloballyDisabled(args)
+        // if we're globally disabled, only allow `platformSpecificFeatures`
+        ? platformSpecificFeatures
+        // otherwise, use `enabledFeatures` if available, falling back to every feature in the bundle
+        : args.site.enabledFeatures || bundledFeatureNames;
```
jonathanKingston, line 51: When don't we have `enabledFeatures`?

shakyShane: when the extension calls 'load'

injected/src/utils.js
shakyShane: actually not needed right now, I'll remove to further reduce the change

injected/entry-points/windows.js
shakyShane: Ahh I remember now, it's to allow things like autofill to call `processConfig`. It's not ideal and we can continue to tweak, but as you mentioned let's not alter it as part of this change

injected/src/content-scope-features.js
jonathanKingston: I wonder if we should make it more explicitly commented?

jonathanKingston (approved):
Didn't test but looks as expected, thanks for doing this!

jonathanKingston (approved):

PR #1573, opened 2025-03-17
Fix a zoom issue in emulated desktop mode on Android

muodov:
**Asana Task/Github Issue:** 
https://app.asana.com/1/137249556945/project/1206777341262243/task/1207691440660873?focus=true

## Description

Fix an issue with emulated desktop mode on Adnroid. This affects instagram.com.

## Testing Steps

- 

## Checklist

*Please tick all that apply:*


injected/src/features/web-compat.js
```diff
@@ -719,6 +719,8 @@ export class WebCompat extends ContentFeature {
             // Race condition: depending on the loading state of the page, initial scale may or may not be respected, so the page may look zoomed-in after applying this hack.
             // Usually this is just an annoyance, but it may be a bigger issue if user-scalable=no is set, so we remove it too.
             forcedValues['user-scalable'] = 'yes';
+            // make sure you can zoom out to see the whole page. See https://app.asana.com/1/137249556945/project/1206777341262243/task/1207691440660873
+            forcedValues['minimum-scale'] = 0;
```
jonathanKingston, line 723: Asking not telling, should this apply in the !viewportTag case?

muodov: It doesn't really matter if the viewport tag is absent: the default value is `0.1`, and in all reasonable scenarios it should be less than the `initial-scale` we calculate above.

jonathanKingston: Can we limit this to when there's an initial scale set by the site? and/or optionally make this remotely configurable?

I agree there _should_ be no impact here, however I know we've had surprises here before.

jonathanKingston (approved):

PR #1606, opened 2025-03-31
Expand patching changes

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/0/481882893211075/1209844047592403/f

## Description

Adds support for more conditional matching. See tech design linked for more detailed explanation.

## Testing Steps

- Ensure that existing features aren't broken by this change; namely element-hiding.

## Checklist

*Please tick all that apply:*


injected/src/config-feature.js
```diff
@@ -41,26 +42,72 @@ export default class ConfigFeature {
...
         const domains = this._getFeatureSettings()?.[featureKeyName] || [];
         return domains.filter((rule) => {
-            if (Array.isArray(rule.domain)) {
-                return rule.domain.some((domainRule) => {
-                    return matchHostname(domain, domainRule);
-                });
-            }
-            return matchHostname(domain, rule.domain);
+            // Support shorthand for domain matching for backwards compatibility
+            return this.matchConditionalChanges(rule.condition || { domain: rule.domain });
```
dbajpeyi, line 55: Could these be totally separate functions?

injected/src/config-feature.js
```diff
@@ -41,26 +42,72 @@ export default class ConfigFeature {
...
+
+    /**
+     * Takes a conditional block and returns true if it applies.
+     * All conditions must be met to return true.
+     * @param {ConditionBlock} conditionBlock
+     * @returns {boolean}
+     */
+    matchConditionalChanges(conditionBlock) {
+        for (const key in conditionBlock) {
+            switch (key) {
```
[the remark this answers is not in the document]
jonathanKingston: I restructured to be a for loop checking an array of conditions, similar to this suggestion based on our call.

injected/src/config-feature.js
jonathanKingston: Split out to flatten to support it turning into an array of conditions as we discussed :)

injected/src/utils.js
```diff
@@ -118,31 +118,40 @@ export function hasThirdPartyOrigin(scriptOrigins) {
...
+export function getTabUrl() {
+    let framingURLString = null;
     try {
         // @ts-expect-error - globalThis.top is possibly 'null' here
-        framingOrigin = globalThis.top.location.href;
+        framingURLString = globalThis.top.location.href;
     } catch {
-        framingOrigin = globalThis.document.referrer;
+        framingURLString = globalThis.document.referrer;
     }
```
jonathanKingston, line 131: I think if we don't have a framingURLString we should do the ancestorOrigins check here. But to note; this won't have a path.

jonathanKingston:
Remote config changes here: https://github.com/duckduckgo/privacy-configuration/pull/2992

This validates this new structure even if we're not going to use it immediately.

injected/src/utils.js
dbajpeyi: I guess we have this check already!

injected/src/utils.js
```diff
@@ -126,31 +126,54 @@ export function hasThirdPartyOrigin(scriptOrigins) {
...
 
+    let framingURL;
     try {
-        // @ts-expect-error - framingOrigin is possibly 'null' here
-        framingOrigin = new URL(framingOrigin).hostname;
+        framingURL = new URL(framingURLString);
     } catch {
-        framingOrigin = null;
+        framingURL = null;
+    }
```
dbajpeyi, line 153: Can't we just reduce this whole block to:

```
let framingURL;
    try {
        const url = framingURLString ?? getTopLevelOriginFromFrameAncestors();
        framingURL = new URL(url);
    } catch {
        framingURL = null;
    }
```

Haven't tried it, but should be equivalent?

dbajpeyi: I just think it reads better.

dbajpeyi: And discussed on MM: can we do tests for this part?

jonathanKingston: yeah added some test cases, doing integration ones would be pretty painful would rather avoid.

I simplified the logic but differently.

dbajpeyi (approved):

PR #1610, opened 2025-04-02
Support navigation based conditional matching

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/0/481882893211075/1209844047592403/f

## Description

Part of: https://app.asana.com/0/481882893211075/1209844047592403/f for navigation handling, super PoC.

## Testing Steps

- 

## Checklist

*Please tick all that apply:*


injected/src/features/api-manipulation.js
```diff
@@ -26,6 +28,11 @@ export default class ApiManipulation extends ContentFeature {
         }
     }
 
+    urlChanged(site) {
+        super.urlChanged(site);
+        this.init();
```
jonathanKingston, line 33: This means we'll rerun rules on navigation change.

injected/src/features/autofill-password-import.js
```diff
@@ -490,23 +491,15 @@ export default class AutofillPasswordImport extends ContentFeature {
...
+    }
+
     init() {
         this.setButtonSettings();
 
         const handlePath = this.handlePath.bind(this);
-        const historyMethodProxy = new DDGProxy(this, History.prototype, 'pushState', {
-            async apply(target, thisArg, args) {
-                const path = args[1] === '' ? args[2].split('?')[0] : args[1];
-                await handlePath(path);
```
jonathanKingston, line 500: I think this might have been wrong as the DDGReflect.apply hadn't applied

jonathanKingston: Actually just re-read the code again, I think it's fine :)

injected/src/features/element-hiding.js
```diff
@@ -360,19 +361,14 @@ export default class ElementHiding extends ContentFeature {
         } else {
             applyRules(activeRules);
         }
-        // single page applications don't have a DOMContentLoaded event on navigations, so
-        // we use proxy/reflect on history.pushState to call applyRules on page navigations
-        const historyMethodProxy = new DDGProxy(this, History.prototype, 'pushState', {
-            apply(target, thisArg, args) {
-                applyRules(activeRules);
```
jonathanKingston, line 367: @dharb  This wasn't applying after the URL had changed, which meant it might not have worked correctly in all cases. This change shall make the apply rules code run after the url bar has changed.

injected/src/content-scope-features.js
```diff
@@ -74,6 +75,16 @@ export async function init(args) {
     resolvedFeatures.forEach(({ featureInstance, featureName }) => {
         if (!isFeatureBroken(args, featureName) || alwaysInitExtensionFeatures(args, featureName)) {
             featureInstance.callInit(args);
+            // Either listenForUrlChanges or urlChanged ensures the feature listens.
+            if (featureInstance.listenForUrlChanges || featureInstance.urlChanged) {
+                registerForURLChanges(() => {
+                    // The rationale for the two separate call here is to ensure that
+                    // extensions to the class don't need to call super.urlChanged()
+                    featureInstance.recomputeSiteObject();
```
jonathanKingston, line 83: @dbajpeyi this is now split out so feature authors don't need to call super() as before.

injected/src/url-change.js
```diff
@@ -0,0 +1,51 @@
...
+        listener();
+    }
+}
+
+function listenForURLChanges() {
+    const urlChangedInstance = new ContentFeature('urlChanged', {}, {});
+    if ('navigation' in globalThis && 'addEventListener' in globalThis.navigation) {
+        // if the browser supports the navigation API, we can use that to listen for URL changes
+        // Listening to navigatesuccess instead of navigate to ensure the navigation is committed.
+        globalThis.navigation.addEventListener('navigatesuccess', () => {
```
jonathanKingston, line 27: @dbajpeyi the navigate event actually fired too early, I validated this works as expected. The API has a way to modify and deny navigations before they happen.

injected/src/features/element-hiding.js
jonathanKingston: Also to note I've added support for the navigation api when it's available; this should be less risk on Chromium based browsers as we'll not need to do any wrapping.

injected/src/content-scope-features.js
dbajpeyi: Non-blocking: do we forsee any major issues if we have both `recomputeSiteObject` and `urlChanged` being called? Can there be any inconsistencies?

dbajpeyi (approved):
LGTM!

PR #1612, opened 2025-04-03
Add Image Captcha Solver

brianhall:
**Asana Task/Github Issue:** 

## Description
Asana: https://app.asana.com/0/72649045549333/1209395858706166/f

## Testing Steps

- See Asana

## Checklist

*Please tick all that apply:*


injected/src/features/broker-protection/captcha-services/providers/image.js
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @param {HTMLElement} captchaImageElement - The captcha image element
+     */
+    async getCaptchaIdentifier(captchaImageElement) {
+        if (captchaImageElement.tagName.toLocaleLowerCase() === 'svg') {
+            // Converting to unknown to avoid TypeScript error when going from HTMLElement to SVGElement
+            const unknownElement = /** @type {unknown} */ (captchaImageElement);
+            const svgElement = /** @type {SVGElement} */ (unknownElement);
+            const base64Image = await svgToBase64Jpg(svgElement);
```
brianhall, line 35: I chose not to use `safeCallWithError` here because I’d have to rewrite it to work in an asynchronous context, but let me know if you think that’s a mistake. I could be convinced otherwise.

injected/src/features/broker-protection/captcha-services/utils/token.js
```diff
@@ -6,22 +6,36 @@ import { getElementByTagName } from '../../utils/utils';
...
+    }
+
     if (!element) {
-        return PirError.create(`[injectTokenIntoElement] could not find element with name ${elementName}`);
+        return PirError.create(`[injectTokenIntoElement] could not find element to inject token into`);
     }
 
     return safeCallWithError(
         () => {
+            if ('value' in element) {
```
brianhall, line 32: It occurs to me that we’re injecting these into either inputs or textarea, so I think ‘value’ would be sufficient in either case? I think the innerHTML thing is a carryover from the old code. If this is the case we can simplify here. Let me know if you disagree.

brianhall:
Not sure why the auto respond PR steps are failing…

injected/src/features/broker-protection/captcha-services/providers/image.js
```diff
@@ -0,0 +1,80 @@
...
```
madblex, line 35: nit: To avoid this you can create a type guard:

```js
/**
 * @param {HTMLElement} element
 * @return {element is SVGElement}
 */
function isSVGElement(element) {
  return element.tagName.toLocaleLowerCase() === 'svg';
}
```

and use as follows:
```suggestion
        if (isSVGElement(captchaImageElement)) {
            const base64Image = await svgToBase64Jpg(captchaImageElement);
```

injected/src/features/broker-protection/captcha-services/providers/image.js
madblex: It's ok, once we add a catch-all pixel for actions, we'll have visibility into this

injected/src/features/broker-protection/captcha-services/utils/image.js
```diff
@@ -0,0 +1,77 @@
...
+
+    // Set canvas dimensions to match the image
+    canvas.width = imageElement.width;
+    canvas.height = imageElement.height;
+
+    // Draw the image onto the canvas
+    ctx.drawImage(imageElement, 0, 0, canvas.width, canvas.height);
+
+    // Convert the canvas content to a JPEG base64 string
+    const base64String = canvas.toDataURL('image/jpeg'); // You can change the format if needed
```
madblex, line 74: nit: I think the code is self explanatory and does not need comments

injected/src/features/broker-protection/captcha-services/utils/token.js
madblex: It's a carryover so I'd ask @shakyShane. It won't hurt to check for the input types since there are other HTML elements supporting the `value` attribute (lists, selects, etc)

madblex (approved):
LGTM! No blockers, just some recommendations

injected/src/features/broker-protection/captcha-services/utils/token.js
brianhall: Do you mean add an additional check on the type attribute to ensure we’re only inserting tokens into inputs and textareas?

injected/src/features/broker-protection/captcha-services/providers/image.js
brianhall: Ah, that’s much better. Thank you!

injected/src/features/broker-protection/captcha-services/utils/token.js
alexCanditech: Exactly

madblex: Yes, that's what I meant

brianhall: Ok, thanks. I’ve updated this, please take a look when you get a chance.

injected/src/features/broker-protection/captcha-services/utils/token.js
```diff
@@ -1,27 +1,59 @@
...
     }
 
     return safeCallWithError(
         () => {
-            element.innerHTML = token;
+            if ((isInputElement(element) && ['text', 'hidden'].includes(element.type)) || isTextAreaElement(element)) {
+                element.value = token;
+            } else if (isInputElement(element)) {
+                return PirError.create(`[injectTokenIntoElement] element is not a text input or textarea`);
+            }
```
madblex, line 37: Is it possible that `element` is neither an input / HTML? I guess it's unlikely, but we'd return success in that case, which is confusing

madblex:
> Not sure why the auto respond PR steps are failing…

@brianhall I pushed a commit to fix it, hope that's ok.

brianhall:
> > Not sure why the auto respond PR steps are failing…
> 
> @brianhall I pushed a commit to fix it, hope that's ok.

For sure, thank you. My editor must have stripped the extensions out, thank you!

injected/src/features/broker-protection/captcha-services/utils/token.js
brianhall: Oops, that was a typo. Should be fixed now.

injected/src/features/broker-protection/captcha-services/utils/token.js
```diff
@@ -1,27 +1,59 @@
...
     return safeCallWithError(
         () => {
-            element.innerHTML = token;
+            if ((isInputElement(element) && ['text', 'hidden'].includes(element.type)) || isTextAreaElement(element)) {
+                element.value = token;
+            } else {
+                return PirError.create(`[injectTokenIntoElement] element is neither a text input or textarea`);
+            }
+
             return PirSuccess.create({ injected: true });
```
madblex, line 39: nit: Now that the conditions are clarified, would it make sense to group the code for success and error paths? i.e to move the `return PirSuccess.create({ injected: true });` on line 35 after setting the element value

injected/src/features/broker-protection/captcha-services/utils/token.js
madblex: Looks good!

madblex (approved):
LGTM with one nit

injected/src/features/broker-protection/captcha-services/utils/token.js
brianhall: Yep, absolutely right. Adopted this suggestion.

madblex (approved):

jonathanKingston (approved):
Just for code ownership r+ maybe it's re-required after the file changes.

PR #1613, opened 2025-04-07
Setting of document for tracker analysis

jonathanKingston:
**Asana Task/Github Issue:** https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/3015

## Description

For assist with: https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/3015

Allows setting of the document as used in the cookie testing in the extension when it's used in a node environment with jsdom.

## Testing Steps

- 

## Checklist

*Please tick all that apply:*

### DIFF
--- a/injected/src/trackers.js
+++ b/injected/src/trackers.js
@@ -4,6 +4,9 @@ import { getGlobal } from './utils.js';
  * Check if the current document origin is on the tracker list, using the provided lookup trie.
  * @param {object} trackerLookup Trie lookup of tracker domains
  * @returns {boolean} True iff the origin is a tracker.
+ *
+ * Note: getGlobal() is used in testing to get the global object,
+ * it's a work around for ESM modules are essentially singletons preventing overriding of global variables.
  */
 export function isTrackerOrigin(trackerLookup, originHostname = getGlobal().document.location.hostname) {
     const parts = originHostname.split('.').reverse();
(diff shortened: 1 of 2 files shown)


injected/src/trackers.js
```diff
@@ -1,9 +1,11 @@
+import { getGlobal } from './utils';
+
 /**
  * Check if the current document origin is on the tracker list, using the provided lookup trie.
  * @param {object} trackerLookup Trie lookup of tracker domains
  * @returns {boolean} True iff the origin is a tracker.
  */
-export function isTrackerOrigin(trackerLookup, originHostname = document.location.hostname) {
```
jonathanKingston, line 6: @shakyShane this makes me pretty sad code wise but I don't see a cleaner way to fix: https://github.com/duckduckgo/duckduckgo-privacy-extension/pull/3015/files other than complete re-think which I don't have time for.

dbajpeyi: Are we doing this specifically for the test (if I understand correctly), then maybe a comment could be there?

dbajpeyi (approved):
LGTM, other than the minor comment.

dbajpeyi (dismissed):

injected/src/trackers.js
jonathanKingston: Added a comment to both files to explain the weirdness :).

I think the only way would to be avoid using jsdom shimming but we don't have time for that or moving the tests here.

dbajpeyi (approved):

PR #1620, opened 2025-04-08
feat(pir): add report metric

madblex:
**Asana Task/Github Issue:** https://app.asana.com/0/0/1209781527904255
**Tech Design:** https://app.asana.com/0/0/1209769226280553

## Description

This PR adds the mechanism for reporting metrics from C-S-S through the native platforms.

Right now it's scoped to Broker Protection feature, but it can be used anywhere else.

## Testing Steps

- TODO

## Checklist

*Please tick all that apply:*


injected/src/features/broker-protection/metrics-reporting/types.d.ts
```diff
@@ -0,0 +1,44 @@
+import type { PirAction, ActionResponse } from '../types.js';
+
+/**
+ * All supported PIR metric names.
+ */
+export type PirMetricName = 'pir.action_error';
```
jonathanKingston, line 6: Don't all .d.ts files become globals, we've avoided them for anything we don't want to become globally defined.

Can we put them in the file instead as types with /* ... */ syntax?

injected/src/features/broker-protection/metrics-reporting/pir-metrics-reporting.js
```diff
@@ -0,0 +1,100 @@
+/** @typedef {import('@duckduckgo/messaging').Messaging} Messaging */
+/** @typedef {import("./types").PirMetric} PirMetric */
+/** @typedef {import("./types").ReportActionMetricParams} ReportActionMetricParams */
+/** @typedef {import("../types").PirAction} PirAction */
+
+import { ErrorResponse } from '../types';
+
+const SUPPORTED_ACTIONS_FOR_PIR_METRICS = ['getCaptchaInfo', 'solveCaptcha'];
```
jonathanKingston, line 8: What's the rationale here, to prevent bugs?

madblex: We just haven't audited the other actions' error messages for PII, added a clarifying comment

injected/src/features/broker-protection/metrics-reporting/types.d.ts
madblex: Oh ok, we should add this to a styleguide somewhere or enforce with an eslint rule (if possible)

jonathanKingston: Yeah if there's a way to enforce that'd be great. (I'm not aware of anything easy other than a dumb test that looks for .d.ts files and makes you add it to a list perhaps?)

injected/src/features/broker-protection/metrics-reporting/pir-metrics-reporting.js
jonathanKingston: Ah I see, I wonder if reportActionMetric should be part of the pir feature itself then? That way there's not a strong coupling to this file. I think we can then just make all content-feature.js instances get the this.metric implementation is my thinking.

madblex: @jonathanKingston 

> I wonder if reportActionMetric should be part of the pir feature itself then?

If we end up moving `reportMetric` to `ContentFeature`, then it makes sense. The initial reason I made it a separate class was for possibly being able to reuse in other PIR functions. If it comes to that, I guess we could pass the function as a parameter, even though it's not a great pattern.

> I think we can then just make all content-feature.js instances get the this.metric

Re/ the 2nd part, I wanted to make `reportMetric` a member of the `ContentFeature` class at the beginning, but got pushback [here](https://app.asana.com/1/137249556945/task/1209769226280553/comment/1209765279718711?focus=true).

jonathanKingston:
👋 @madblex — quick housekeeping: this PR hasn't been updated in ~13 months. Is it still active, or can it be closed out? No rush — just doing a sweep of stale PRs to keep the queue tidy. (I won't close it; leaving that to you.)

PR #1662, opened 2025-05-01
Add query param color=white to 

mgurgel:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1207252092703676/task/1210128066798031?focus=true

## Description

YouTube's seemingly rolling out a new embedded player UI which is used in Duck Player. We discovered that adding a URL param (color=white) will force the YouTube player to use the legacy UI instead of the new player.

## Testing Steps

- Run this branch of C-S-S on an Android or iOS build
- Open Duck Player on any video
- Use Safari or Chrome to inspect the Duck Player window
- Confirm that the iframe has a src attribute that includes the query param `color=white` . You can run this on the console to confirm `document.querySelector('iframe').src.includes('color=white');`
- Confirm that the player interface looks like this

image

and not like this

image

## Checklist

*Please tick all that apply:*

### DIFF
--- a/special-pages/pages/duckplayer/app/embed-settings.js
+++ b/special-pages/pages/duckplayer/app/embed-settings.js
@@ -71,6 +71,7 @@ export class EmbedSettings {
 
         url.searchParams.set('rel', '0'); // shows related videos from the same channel as the video
         url.searchParams.set('modestbranding', '1'); // disables showing the YouTube logo in the video control bar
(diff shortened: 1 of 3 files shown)


mgurgel:
@shakyShane I'm checking whether we can roll this change out to desktop too.

shakyShane (approved):

mgurgel:
Tested on desktop. No issues I could see. 
image

PR #1752, opened 2025-06-17
Fix config schema tests

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/72649045549333/task/1210563609609919?focus=true

## Description
Schema validate JSON files where it makes sense to.

## Testing Steps

- 

## Checklist

*Please tick all that apply:*


injected/unit-test/features.js
```diff
@@ -23,3 +34,83 @@ describe('Features definition', () => {
...
+        path.resolve(__dirname, '../integration-test/test-pages/duckplayer/config/overlays-live.json'),
+        path.resolve(__dirname, '../integration-test/test-pages/duckplayer/config/click-interceptions-disabled.json'),
+        path.resolve(__dirname, '../integration-test/test-pages/duckplayer/config/disabled.json'),
+        path.resolve(__dirname, '../integration-test/test-pages/duckplayer/config/overlays-drawer.json'),
+        path.resolve(__dirname, '../integration-test/test-pages/duckplayer/config/overlays.json'),
+        path.resolve(__dirname, '../integration-test/test-pages/duckplayer/config/thumbnail-overlays-disabled.json'),
+        path.resolve(__dirname, '../integration-test/test-pages/duckplayer/config/video-alt-selectors.json'),
+        path.resolve(__dirname, '../integration-test/test-pages/duckplayer/config/video-overlays-disabled.json'),
+        // Message bridge configs
+        path.resolve(__dirname, '../integration-test/test-pages/message-bridge/config/message-bridge-enabled.json'),
```
jonathanKingston, line 98: These ones specifically are actually invalid. For example if we add new bridges then this will become immediately invalid; which it doesn't matter.

For now I think this is good enough but perhaps in future we'd have a way to mark _why_ it's invalid. So the rest of the file gets schema checked still.

PR #1753, opened 2025-06-17
Duck Player: Add support for new YouTube Player UI links

mgurgel:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1207252092703676/task/1210128066798037?focus=true

## Description

Adds a click handler to links to the current video within the embed

## Testing Steps

- Run this branch on a native app (or check [test builds here](https://app.asana.com/1/137249556945/project/1207252092703676/task/1210572366557467?focus=true))
- Set Duck Player to Always On
- Play a video on Duck Player
- Click on the video title or YouTube logo
- Confirm that new tab opens in YouTube and not Duck Player

## Checklist

*Please tick all that apply:*

### DIFF
--- a/special-pages/pages/duckplayer/app/components/Player.jsx
+++ b/special-pages/pages/duckplayer/app/components/Player.jsx
@@ -6,7 +6,10 @@ import { useSettings } from '../providers/SettingsProvider.jsx';
 import { createIframeFeatures } from '../features/iframe.js';
 import { Settings } from '../settings';
 import { useTypedTranslation } from '../types.js';
-import { EmbedSettings } from '../embed-settings';
+
+/**
+ * @import {EmbedSettings} from '../embed-settings.js';
+ */
 
 /**
  * Player component renders an embedded media player.
--- a/special-pages/pages/duckplayer/app/components/Player.jsx
+++ b/special-pages/pages/duckplayer/app/components/Player.jsx
@@ -2,10 +2,11 @@ import { h } from 'preact';
 import cn from 'classnames';
 import styles from './Player.module.css';
 import { useEffect, useRef } from 'preact/hooks';
-import { useSettings } from '../providers/SettingsProvider.jsx';
+import { useSettings, usePlatformName, useOpenOnYoutubeHandler } from '../providers/SettingsProvider.jsx';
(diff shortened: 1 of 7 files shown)


special-pages/pages/duckplayer/app/components/Player.jsx
```diff
@@ -103,8 +112,16 @@ function useIframeEffects(src) {
...
+            features.replaceWatchLinks(),
         ];
 
+        // Handle watch link clicks in iframe
+        window.addEventListener(WATCH_LINK_CLICK_EVENT, () => {
+            if (embed) {
+                openOnYoutube(embed);
+            }
+        });
+
```
shakyShane, line 124: this is better for sure. if you want to go 1 step further just pass a callback instead, like this https://github.com/duckduckgo/content-scope-scripts/commit/19f6d41f5b270870a9bf59c9b57284b3dfac1cc3

^ no need to apply this, just showing

special-pages/pages/duckplayer/app/features/replace-watch-links.js
```diff
@@ -0,0 +1,62 @@
...
+            console.log('could not access contentDocument');
+            return () => {};
+        }
+
+        if (win && doc) {
+            doc.addEventListener(
+                'click',
+                (e) => {
+                    /** @type {HTMLLinkElement|null} */
+                    const closestLink = /** @type {Element} */ (e.target).closest('a[href]');
```
shakyShane, line 36: I wouldn’t cast in this scenario, just because it’s not our environment, add something like this instead to be safer and please Typescript at the same time

```js
if (!(e.target instanceof Element)) return;
```

shakyShane (changes requested):
1 thing to change (the casting) and theres a demo PR for you to look at (but no need to adopt it, just wanted to close the conversation we had before)

special-pages/pages/duckplayer/app/components/Player.jsx
mgurgel: Applied. Thanks for removing the leftovers from previous iterations too.

shakyShane (dismissed):

mgurgel:
Latest version tests on iOS (iPad), Android, Windows and macOS.

PR #1794, opened 2025-07-04
Jkt/getting started

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1201614831475344/task/1210717003881747?focus=true

## Description

Move content from: https://app.asana.com/1/137249556945/project/1201614831475344/task/1203837324583081?focus=true

## Testing Steps

- 

## Checklist

*Please tick all that apply:*

### DIFF
--- a/injected/docs/development-utilities.md
+++ b/injected/docs/development-utilities.md
@@ -4,6 +4,54 @@
 
 To handle the difference in scope injection we expose multiple utilities which behave differently per browser in `src/utils.js` and `ContentFeature` base class. For Firefox the code exposed handles [xrays correctly](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Sharing_objects_with_page_scripts) without needing the features to be authored differently.
 
+## Development Setup
+
+### Local Development Setup
+
+**Important**: Before cloning the repo, to avoid major headaches, make sure you clone it with unix-style line endings:
+
+```shell
+git clone --config core.autocrlf=false https://github.com/duckduckgo/content-scope-scripts
+```
+
+The Content Scope Scripts repo includes the build artifacts, which need to be generated as part of your commit.
+
+### Windows Development
+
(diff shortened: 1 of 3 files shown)


injected/docs/development-utilities.md
```diff
@@ -4,6 +4,77 @@
...
+Setting up the tools on Windows will be a pain (and really impossible without modifying the scripts, as the build-scripts are not OS-agnostic).
+
+**Recommended**: Use Windows Subsystem for Linux - [WSL Installation Guide](https://learn.microsoft.com/en-us/windows/wsl/install)
+
+Once you have WSL running, make sure you have node and make installed:
+
+```shell
+sudo apt update
+sudo apt install make
+```
```
shakyShane, line 53: this is only true if we allow new scripts that break the build. I currently build CSS on windows all the time and it works fine with a regular node install. We shouldn't proliferate the use of bash scripts for build (eg: just keep them for releases etc). I don't think the WSL requirement is a good one, it's a barrier to entry that isn't needed if we write cross-platform node scripts instead of bash scripts. especially important given it's a repo of Javascript etc.

For example, right now I can open any terminal in Windows (that has node installed) and I can build CSS (minus a legacy perl thing).

shakyShane (changes requested):
I don't mind having a 'recommendation' for WSL, but I wouldn't want us to get closer to codebases like the extension where the continued use of bash has made it difficult for FE/JS folks to contribute to. Node is cross-platform by default and we should lean on it IMO - that's how you make contribution on Windows easier. Requiring WSL is the opposite, it's a more complicated contribution path :) Maybe we just change the wording or something to indicate 'you might prefer WSL' and not 'you can only build with WSL'?

jonathanKingston:
@shakyShane OK I've made it a suggestion along with the other bits. I think this sort of highlights orphaned docs like this in Asana and how outdated they get. I think some of the other advice is worth having anyway.

My view is we can continue to iterate here and AI tools are likely to find silliness faster too.

shakyShane (approved):

PR #1799, opened 2025-07-07
Add conditional action type

brianhall:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/72649045549333/task/1210684503159912?focus=true

## Description
Some brokers display a different flow depending on the user's data entered in the opt out form. This change handles these cases and returns the steps back to the native layer in the case of a successful condition.

## Testing Steps

- `cd injected && npx playwright test integration-test/broker-protection-tests/broker-protection.spec.js `

## Checklist

*Please tick all that apply:*


injected/src/features/broker-protection/utils/expectations.js
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/**
+ * Check that the current URL includes a given string
+ *
+ * @param {import("../types").Expectation} expectation
+ * @return {import("../types").BooleanResult}
+ */
+export function urlExpectation(expectation) {
+    const url = window.location.href;
```
jonathanKingston, line 105: We already have URL conditions in C-S-S for "condition". Can we reuse or code-share? It supports any URL pattern condition: https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API

injected/src/features/broker-protection/utils/expectations.js
```diff
@@ -0,0 +1,123 @@
...
+export function elementExpectation(expectation, root) {
+    if (expectation.parent) {
+        const parent = getElement(root, expectation.parent);
+        if (!parent) {
+            return {
+                result: false,
+                error: `parent element not found with selector: ${expectation.parent}`,
+            };
+        }
+        parent.scrollIntoView();
```
jonathanKingston, line 46: Wouldn't keeping conditionals isolated from actions be cleaner? Again we could then reuse these conditions elsewhere in C-S-S more.

brianhall: Happy to discuss further when we work on consolidating with the other related code in C-S-S.

shakyShane (approved):

PR #1806, opened 2025-07-08
Device enum support

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/72649045549333/task/1210512782751384?focus=true

## Description

Adds a simple shim that talks to native if `willPrompt` if it will, we then use the presence of devices to shim the API.
If it won't prompt, we use the default API which shouldn't then cause issues.

## Testing Steps

- Go to example.com
- Type `navigator.mediaDevices.enumerateDevices()` in the console
- Check that the response works correctly.

## Checklist

*Please tick all that apply:*


injected/integration-test/test-pages/webcompat/config/enumerate-devices-api.json
```diff
@@ -0,0 +1,14 @@
+{
```
dharb, line 1: Any reason you have both this and device-enumeration.json?

jonathanKingston: Probably file rename let me delete :)

jonathanKingston:
> Bizarre that this is necessary, but it is what it is.

Totally lol!

PR #1817, opened 2025-07-11
messaging: move docs to Markdown and never throw on calls to .notify

shakyShane:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1201141132935289/task/1210768128803703?focus=true

## Description

Small improvements towards https://app.asana.com/1/137249556945/project/1201141132935289/task/1210680503659356?focus=true

Screenshot 2025-07-11 at 08 02 21

## Testing Steps

- No functional changes, just moving docs around

## Checklist

*Please tick all that apply:*


messaging/index.js
```diff
@@ -87,12 +86,17 @@ export class Messaging {
             method: name,
             params: data,
         });
-        this.transport.notify(message);
+        try {
+            this.transport.notify(message);
+        } catch (e) {
+            // Silently ignoring any transport errors here as per section 4.1 of https://www.jsonrpc.org/specification
+            // Notifications are fire+forget and should be able to be sent without any knowledge of the receiving ends support
+        }
```
shakyShane, line 94: This was already the case on other platforms, just on webkit we were bubbling up an error which we should not have.

jonathanKingston (approved):
Do we need to validate all consumers of notify don't expect the throw? I know for bridge you were using it as a proxy for lack of support right?

jonathanKingston:
@cursor anything to add?

jonathanKingston:
> @cursor anything to add?

Leave me unread then. 😢

shakyShane:
> Do we need to validate all consumers of notify don't expect the throw? I know for bridge you were using it as a proxy for lack of support right?

No, the bridge only delegates to the transports regardless.

PR #1820, opened 2025-07-14
NTP: Add useBlurWorkaround hook to fix WebKit blur issue in Omnibar

noisysocks:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1210784594537119

## Description

Fixes an issue in the Omnibar widget where, in the macOS browser, moving focus to the address bar doesn’t close the suggestions list.

The problem is that Safari/WebKit doesn’t trigger a blur on input elements when focus is moved to the address bar.

I noticed that Safari/WebKit **does** correctly remove `:focus` styling, though, so my “clever” workaround is to detect when `:focus` is removed by setting a CSS variable in the `:focus` style and then polling for changes to this.

## Testing Steps

1. Go to https://deploy-preview-1820--content-scope-scripts.netlify.app/build/pages/new-tab/?omnibar=true in Safari or DuckDuckGo on macOS.
2. Type “pizza” into the Search form so that suggestions list opens.
3. Click on the browser address bar. Suggestions list should be dismissed.

## Checklist

*Please tick all that apply:*

### DIFF
--- a/special-pages/pages/new-tab/app/omnibar/components/useBlurWorkaround.js
+++ b/special-pages/pages/new-tab/app/omnibar/components/useBlurWorkaround.js
@@ -14,7 +14,7 @@ export function useBlurWorkaround() {
         if (!element) return;
 
         let lastFocusState = '0';
-        let rafId = null;
+        let rafId = 0;
 
         const checkFocusState = () => {
             const currentFocusState = getComputedStyle(element).getPropertyValue('--focus-state').trim();


noisysocks:
Oh, weird, the `blur` event fires if it’s added using `addEventListener`. Maybe this crazy workaround is unnecessary.

shakyShane (approved):

noisysocks:
Closing in favour of https://github.com/duckduckgo/content-scope-scripts/pull/1851.

PR #1826, opened 2025-07-15
fix: get ownerDocument before calling element.evaluate(...)

borgateo:
## Context

in [data-broker-protection debugger]( https://github.com/duckduckgo/data-broker-protection-json)  when using `safeQuerySelectorXPath`, a DOMException was thrown with the message:

```
"safeQuerySelectorXPath threw DOMException: Node cannot be used in a document other than the one in which it was created safeQuerySelectorXPath utils.js:113"
```

This occurred because nodes created in one document (e.g., from an iframe) cannot be used in another document context.

## Solution

Added a check to ensure that the node’s ownerDocument matches the current document before proceeding with operations in safeQuerySelectorXPath. This prevents cross-document node usage and avoids the DOMException.

### DIFF
--- a/injected/src/features/broker-protection/utils/utils.js
+++ b/injected/src/features/broker-protection/utils/utils.js
@@ -1,4 +1,14 @@
 /**
+ * Gets the owner document of an element or uses document as fallback
+ *
+ * @param {Node|Element} element
+ * @return {Document} The owner document
+ */
+function getOwnerDocument(element) {
+    return element.ownerDocument || document;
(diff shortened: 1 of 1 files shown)


jonathanKingston:
👋 @borgateo — quick housekeeping: this PR hasn't been updated in ~11 months. Is it still active, or can it be closed out? No rush — just doing a sweep of stale PRs to keep the queue tidy. (I won't close it; leaving that to you.)

PR #1848, opened 2025-07-25
Handle multiple bridge instances within navigator interface

brianhall:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1210888277209380?focus=true

## Description
Lets C-S-S handle message bridge instances.

## Testing Steps
TBD

## Checklist

*Please tick all that apply:*


injected/src/features/navigator-interface.js
```diff
@@ -35,7 +37,17 @@ export default class NavigatorInterface extends ContentFeature {
...
                      */
                     createMessageBridge(featureName) {
-                        return createPageWorldBridge(featureName, args.messageSecret);
+                        const existingBridge = store[featureName];
+                        if (existingBridge) return existingBridge;
+
+                        const bridge = createPageWorldBridge(featureName, args.messageSecret);
+
+                        if (!bridge) {
+                            throw new Error('Unable to create bridge');
```
jonathanKingston, line 46: Looking at `createPageWorldBridge` I don't see it ever not returning, instead it just throws so I don't think will ever get called.

shakyShane: yeah, just drop this bit 👍🏻

brianhall:
Ready for review, gents. Thank you!

injected/src/features/navigator-interface.js
```diff
@@ -35,7 +37,13 @@ export default class NavigatorInterface extends ContentFeature {
...
                      */
                     createMessageBridge(featureName) {
-                        return createPageWorldBridge(featureName, args.messageSecret);
+                        const existingBridge = store[featureName];
+                        if (existingBridge) return existingBridge;
+
+                        const bridge = createPageWorldBridge(featureName, args.messageSecret);
+
+                        store[featureName] = bridge;
+                        return bridge;
```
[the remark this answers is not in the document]
jonathanKingston: Neither of these seem like a concern. There's only going to be one instance per frame and secret per frame too. If we needed to rotate that key, we could find a solution to wire in the updated key.

jonathanKingston (approved):

PR #1858, opened 2025-07-29
NTP: Fix disappearing search icon in WebKit

noisysocks:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1209121419454298/task/1210868202766664?focus=true

## Description

Fixes `.tab` being obscured by `.blob` when the user opens the customize sidebar.

This seems to be a WebKit rendering bug caused by both `.App_mainLayout` having a stacking context (from `will-change: transform`) and `.blob` having a stacking context (from `will-change: translate`).

Removing either `will-change` fixes the issue but causes other issues (see https://github.com/duckduckgo/content-scope-scripts/pull/1822). Instead, I’m forcing a third stacking context onto `.tab` using the `translateZ()` hack. 

## Testing Steps

1. In Safari/DuckDuckGo, navigate to https://deploy-preview-1858--content-scope-scripts.netlify.app/build/pages/new-tab/?omnibar=true.
2. Open the Customize sidebar.
3. The Search tab and icon in the omnibar widget should not disappear.

## Checklist

*Please tick all that apply:*

### DIFF
--- a/special-pages/pages/new-tab/app/omnibar/components/TabSwitcher.module.css
+++ b/special-pages/pages/new-tab/app/omnibar/components/TabSwitcher.module.css
@@ -26,6 +26,7 @@
     justify-content: center;
     padding: 0 var(--sp-3);
     position: relative;
+    transform: translateZ(0); /* Force .tab into its own stacking context. Workaround for WebKit mistakenly layering .blob on top of .tab when .App_mainLayout animates. */
     z-index: 1;
 
     &:hover:not([aria-selected="true"])::before {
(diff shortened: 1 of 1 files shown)


special-pages/pages/new-tab/app/omnibar/components/TabSwitcher.module.css
```diff
@@ -53,6 +54,7 @@
     translate: calc(2px + var(--tab-index) * (100% + 2px));
     width: calc((100% - 4px - (var(--tab-count) - 1) * 2px) / var(--tab-count));
     will-change: translate;
+    z-index: 0;
```
noisysocks, line 57: Not necessary, just adding it to make it explicit that `.blob` is supposed to be under `.tab`.

shakyShane (approved):

PR #1866, opened 2025-07-31
NTP: Make omnibar always overflow

noisysocks:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1209121419454298/task/1210769573066743

## Description

This PR has two key user-visible changes:

1. When the Duck.ai textarea expands for multi-line input, the omnibar will now overflow its container and cover other widgets. This is the intended design.
2. The omnibar will only animate height changes when switching between Search and Duck.ai or when Duck.ai textarea expands for multi-line input. The suggestions list no longer causes height change animations. This better matches the native omnibar behaviour.

To do this I had to change the markup structure so that there’s a new middle container (popup) which has a non-animating height.  `SuggestionsList` then moves there. See 82c14edb40bc87c887e8c28055bb2acebe5ba897.

```html
<div class=“spacer”> // Fixed animated height. Contents will overflow
    <div class=“popup”> // Auto non-animated height. Has border when suggestions list open
        <div class=“field”> // Fixed animated height. Has border when suggestions list closed
            <SearchForm /> or <AiChatForm />
        </div>
        <SuggestionsList />
    </div>
</div>
```

This meant refactoring `SuggestionsList` out of `SearchForm` which I did by turning `SearchForm` into a composite component consisting of `SearchFormProvider`, `SearchForm`, and `SuggestionsList`. See ebf2bd698ba683790a00e0839df346c4f55eb0d3.

I then further refactored things so that `SearchForm` contains all of the view-related concerns and `SearchFormProvider` contains minimal data-related concerns. See 266199e011f45cf209dc965babd372bde9ee65b7.

I also inlined `Container` into `Omnibar` since they’re so tightly coupled. See 0531a00d14cc032a845378dc7c9bbf723a496f06.

A big effort, but I think the concerns are much more neatly separated now, and it enables the desired UI behaviour.

## Testing Steps

https://deploy-preview-1866--content-scope-scripts.netlify.app/build/pages/new-tab/?omnibar=true

1. Test that changes to suggestions list don’t cause animated height changes.
2. Test that switching tab and multi-line input into the Duck.ai textarea do cause animated height changes.
3. Smoke test all omnibar functionality.

## Checklist

*Please tick all that apply:*

### DIFF
--- a/special-pages/pages/new-tab/app/omnibar/components/ResizingContainer.js
+++ b/special-pages/pages/new-tab/app/omnibar/components/ResizingContainer.js
@@ -2,9 +2,9 @@ import { h } from 'preact';
 import { useLayoutEffect, useRef, useState } from 'preact/hooks';
 
 /**
- * @param {import('preact').JSX.HTMLAttributes<HTMLDivElement>} props
+ * @param {Omit<import('preact').JSX.HTMLAttributes<HTMLDivElement>, 'style'>} props
  */
-export function ResizingContainer(props) {
+export function ResizingContainer({ children, ...props }) {
     const contentRef = useRef(/** @type {HTMLDivElement|null} */ (null));
     const [currentHeight, setCurrentHeight] = useState(/** @type {number|null} */ (null));
 
@@ -21,7 +21,7 @@ export function ResizingContainer(props) {
 
     return (
         <div {...props} style={{ height: currentHeight ?? 'auto' }}>
-            <div ref={contentRef}>{props.children}</div>
+            <div ref={contentRef}>{children}</div>
         </div>
     );
 }


special-pages/pages/new-tab/app/omnibar/components/Omnibar.js
```diff
@@ -62,33 +65,35 @@ export function Omnibar({ mode, setMode, enableAi }) {
...
-                        onBlur={() => setFocusRing(false)}
-                        onInput={() => setFocusRing(false)}
-                        onChange={setQuery}
-                        onSubmit={handleSubmitChat}
-                    />
-                )}
-            </Container>
+            <SearchFormProvider
+                // Remove any newlines that come from switching from chat to search
+                term={query.replace(/\n/g, '')}
```
shakyShane, line 73: I'm not suggesting this will be any kind of perf hit, but having this here incurs the 'replace' on every render (key stroke?). Isn't there a way we can perform this change as part of the user-initiated action instead? that way the program would better represent the intention? 

if you tried that and it's too much of a change, feel free to ignore for now. but this is the only thing that stands out in the changeset.

noisysocks: Hm, it’s a good point.

For context, this `replace` was introduced in a previous PR to fix a bug that used to exist when you typed multiline input into the Duck.ai tab, then switched to the Search tab. The `<input>` in Search can’t display line breaks and things get royally weird when they’re encountered.

I still want the newlines to be in the state, though. That is, I want you to be able to type multiline content into Duck.ai, switch to Search, then immediately switch back, and have everything be the same.

That’s why I'm stripping out newlines during render. The "source of truth” data has newlines and we’re simply removing them for display. Similar to how you might call `.toUpperCase()` on a string before displaying that string.

Can you think of a better approach? If concerned about performance we can useMemo but I don’t think `replace` would be particularly slow.

shakyShane (approved):

special-pages/pages/new-tab/app/omnibar/components/ResizingContainer.js
```diff
@@ -0,0 +1,27 @@
...
+        setCurrentHeight(content.scrollHeight);
+
+        const resizeObserver = new ResizeObserver(() => setCurrentHeight(content.scrollHeight));
+        resizeObserver.observe(content);
+        return () => resizeObserver.disconnect();
+    }, []);
+
+    return (
+        <div {...props} style={{ height: currentHeight ?? 'auto' }}>
+            <div ref={contentRef}>{props.children}</div>
```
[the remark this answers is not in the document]
noisysocks: OK have it your way ad62039807154fd5fb6b9703e6d4b426c4ffc771

PR #1906, opened 2025-08-19
Open up canShare some more

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/72649045549333/task/1211078754350004?focus=true

## Description

Fast follow from: https://github.com/duckduckgo/content-scope-scripts/pull/1899
Makes the following changes:
- Allows all fields to be undefined or null
- Ensures that one of url, title or text are still present and a string.
- Allows for `FileList` also that are empty.

## Testing Steps

- 

## Checklist

*Please tick all that apply:*

### DIFF
--- a/injected/src/features/web-compat.js
+++ b/injected/src/features/web-compat.js
@@ -21,9 +21,18 @@ const MSG_DEVICE_ENUMERATION = 'deviceEnumeration';
 
 function canShare(data) {
     if (typeof data !== 'object') return false;
-    if (!('url' in data) && !('title' in data) && !('text' in data)) return false; // At least one of these is required
+    // Make an in-place shallow copy of the data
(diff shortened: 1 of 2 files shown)


injected/src/features/web-compat.js
```diff
@@ -21,9 +21,18 @@ const MSG_DEVICE_ENUMERATION = 'deviceEnumeration';
 
 function canShare(data) {
     if (typeof data !== 'object') return false;
-    if (!('url' in data) && !('title' in data) && !('text' in data)) return false; // At least one of these is required
+    // Make an in-place shallow copy of the data
```
jonathanKingston, line 24: This just avoids having to do all of the checking on every if statement below. Including the "at least one of these" if statement.

Issue #1900, opened 2025-08-18
can't access property "includes", args.site.enabledFeatures is undefined


fei-wtg:
Hi, the latest Firefox extension is stopping my web site, 
the message is `can't access property "includes", args.site.enabledFeatures is undefined`

`Stack Trace:
isFeatureBroken@<anonymous code>:980:143
updateFeaturesInner/<@<anonymous code>:9240:12
updateFeaturesInner@<anonymous code>:9239:22
async*update@<anonymous code>:9232:24
@<anonymous code>:9261:15`

it looks args.site is the limited site object when above functions are called

jonathanKingston:
Looks like it's click to play related, so sites that use the Facebook API.

jonathanKingston:
Took longer than expected as Firefox changed their store policy so we had to block on some internal process. Sorry for the delay. Is now live :)

PR #1907, opened 2025-08-20
Ensure message correctness in extension

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1201614831475344/task/1211099580633727?focus=true

Fixes: #1900
## Description

I think this issue is due to tab restoration in the service worker mv3 handling code in Firefox code only and only when the extension is updated (hence a new bout of reproductions).

I've done the following:
- Short circuited the init code when we get an invalid message, we can't do much in this scenario. The extension side needs to investigate this. These messages are likely more corrupt than just the `enabledFeatures` missing, so we just shouldn't risk breaking the page.
- Added typing to the `isFeatureBroken` method which would have prevented this exact issue.
- Handled the undefined `enabledFeatures` case in this method.
- Split out the logic as it was getting a little unreadable.

As a perf benefit:
- Short circuited the update logic just for the CTL feature as it's not required so it cuts out many function calls with checking `listenForUpdateChanges`

## Testing Steps

- Visit various pages and validate no error is seen.
- Also visit https://privacy-test-pages.site/privacy-protections/click-to-load/ and ensure it's working as expected.
- Ensure https://osmand.net/map/plan/#15/46.1266/5.2599 has no warnings.

## Checklist

*Please tick all that apply:*


dharb (approved):
Looks good 👍

PR #1912, opened 2025-08-21
Add support for maxSupportedVersion

jonathanKingston:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1208598406046969/task/1211114711130445?focus=true

## Description

Adds a matching implementation to minSupportedVersion check that **is inclusive**.
## Testing Steps

- 

## Checklist

*Please tick all that apply:*


injected/src/utils.js
```diff
@@ -712,6 +730,12 @@ export function computeEnabledFeatures(data, topLevelHostname, platformVersion,
                     return false;
                 }
             }
+            // Check that the platform supports maxSupportedVersion checks and that the feature has a maxSupportedVersion
+            if (feature.maxSupportedVersion && platformVersion) {
```
jonathanKingston, line 734: Dunno if we should add this, it's not supported on other platforms and in the config syntax. I think I'll drop it.

shakyShane (dismissed):

shakyShane (approved):

PR #1934, opened 2025-09-02
Notification: Add toString and give permission a set value

laghee:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1201870266890790/task/1210988190527661?focus=true

## Description

- Switched `Notification.permission` to automatic value instead of using a get
- Added `toString`s for Notification shim
- Added integration tests

## Testing Steps

- 

## Checklist

*Please tick all that apply:*


injected/src/features/web-compat.js
```diff
@@ -202,36 +202,71 @@ export class WebCompat extends ContentFeature {
...
 
-        this.defineProperty(window.Notification, 'maxActions', {
+        this.defineProperty(/** @type {any} */ (window.Notification), 'permission', {
+            value: 'denied',
+            writable: false,
+            configurable: true,
+            enumerable: true,
+        });
+
+        this.defineProperty(/** @type {any} */ (window.Notification), 'maxActions', {
```
jonathanKingston, line 251: Do we need these typings since you've dropped out the `defineProperty` changes?

injected/src/features/web-compat.js
```diff
@@ -202,36 +202,71 @@ export class WebCompat extends ContentFeature {
...
             enumerable: true,
         });
+
+        this.defineProperty(/** @type {any} */ (window.Notification), 'requestPermission', {
+            value: Object.assign(
+                function requestPermission() {
+                    return Promise.resolve('denied');
+                },
+                {
+                    toString: () => 'function requestPermission() { [native code] }',
```
jonathanKingston, line 263: Why not use the same pattern as `window.Notification.toString`?

I think this hardcoding of the `toString` should be something we discuss fast follow removing. We have those existing wrapper methods, they're just in a bit of a weird format?

injected/src/features/web-compat.js
```diff
@@ -202,36 +202,71 @@ export class WebCompat extends ContentFeature {
...
+
+        this.defineProperty(/** @type {any} */ (window.Notification), 'toString', {
+            value: toStringFunc,
+            writable: false,
             configurable: true,
-            enumerable: true,
+            enumerable: false,
+        });
+
+        this.defineProperty(/** @type {any} */ (window.Notification), 'prototype', {
```
jonathanKingston, line 230: Is this one needed? `NotificationConstructor` should get a prototype of Object by default iirc.

I ask because if we do need it, this will also get a prototype by default which I don't think would be expected. You'd have `Notification.prototype.prototype `accessible. (This can be avoided with `Object.create(null)` instead of `{}`).

But overall, I don't think this whole `defineProperty` is needed but validate that.

injected/src/features/web-compat.js
```diff
@@ -202,36 +202,71 @@ export class WebCompat extends ContentFeature {
         if (window.Notification) {
             return;
         }
+
         // Expose the API
+        // window.Notification polyfill is intentionally incompatible with DOM lib types
+        const NotificationConstructor = function Notification(title) {
```
jonathanKingston, line 208: Drop `title` in the function signature, lint is complaining about it.

injected/integration-test/web-compat.spec.js
```diff
@@ -130,6 +130,16 @@ test.describe('Ensure Notification interface is injected', () => {
...
+
+        const notificationToString = await page.evaluate(() => {
+            return window.Notification.toString();
+        });
+        expect(notificationToString).toEqual('function Notification() { [native code] }');
+
+        const requestPermissionToString = await page.evaluate(() => {
+            return window.Notification.requestPermission.toString();
+        });
+        expect(requestPermissionToString).toEqual('function requestPermission() { [native code] }');
```
jonathanKingston, line 142: Should we solve the good ol `toString.toString()` problem here? Happy to leave to a follow up though.

jonathanKingston (changes requested):
Dropped some feedback inline, overall this is looking a lot more focused now thanks!

injected/src/features/web-compat.js
laghee: In the end, I think it was switching permission and requestPermission to be values that fixed those sites, so a fast follow on toString makes sense to me. I think hardcoding it like that is causing or contributing to the toString2 naming weirdness (although I haven't seen that actually break anything yet, it seems suboptimal).

laghee: Ok, in the end, the 💡 FINALLY went on and I just switched these over to use `wrapToString`, which seems to both A) work and B) solve the stupid toString2 madness so 🎉

PR #1954, opened 2025-09-09
history: adding sites support

shakyShane:
**Asana Task/Github Issue:** 

## Testing Steps

- 

## Checklist

*Please tick all that apply:*


special-pages/pages/history/app/components/Sidebar.module.css
```diff
@@ -39,6 +39,23 @@
             background: var(--color-white-at-6);
         }
     }
+
+    &:last-child {
```
shakyShane: I don't think it's always last-child, right? it's only this new 'sites' thing.

special-pages/pages/history/integration-tests/history.spec.js
```diff
@@ -263,4 +263,20 @@ test.describe('history', () => {
...
+            // rgba(0, 0, 0, 0.06) is --color-black-at-6 var value
+            await hp.lastItemDividerHasColor({ rgb: 'rgba(0, 0, 0, 0.06)' });
+        });
+        test('divider color - dark', async ({ page }, workerInfo) => {
+            const hp = HistoryTestPage.create(page, workerInfo).withEntries(5);
+            await hp.openPage({});
+            await hp.darkMode();
+            // rgba(255, 255, 255, 0.06) is --color-white-at-6 var value
+            await hp.lastItemDividerHasColor({ rgb: 'rgba(255, 255, 255, 0.06)' });
+        });
```
shakyShane: there are screenshot tests you could use here instead, take a look at the exisiting ones since you might need to update them anyway

special-pages/pages/history/app/components/Sidebar.module.css
moon0326: @shakyShane Thank you! I initially thought 'Sites' would always be the last item, but I see we might add or move other items later. I've updated it in https://github.com/duckduckgo/content-scope-scripts/pull/1954/commits/8facc6254372de53e28c5bcf998a920b9b934165 to target 'Sites' directly.

mgurgel (approved):

PR #1982, opened 2025-09-26
Add timeout to enumerateDevices call

jonathanKingston:
**Asana Task/Github Issue:** 

## Testing Steps

- 

## Checklist

*Please tick all that apply:*

### DIFF
--- a/injected/src/features/web-compat.js
+++ b/injected/src/features/web-compat.js
@@ -883,10 +883,27 @@ export class WebCompat extends ContentFeature {
              * @returns {Promise<MediaDeviceInfo[]>}
              */
             apply: async (target, thisArg, args) => {
+                // Get timeout settings - enabled by default with 5 second timeout
+                const settings = this.getFeatureSetting('enumerateDevices') || {};
+                const timeoutEnabled = settings.timeoutEnabled !== false; // enabled by default
+                const timeoutMs = typeof settings.timeoutMs === 'number' ? settings.timeoutMs : 5000;
+
                 try {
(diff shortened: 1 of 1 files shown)


injected/src/features/web-compat.js
```diff
@@ -867,6 +867,17 @@ export class WebCompat extends ContentFeature {
         return deviceInfo;
     }
 
+    /**
+     * Helper to wrap a promise with timeout
+     * @param {Promise} promise - Promise to wrap
+     * @param {number} timeoutMs - Timeout in milliseconds
+     * @returns {Promise} Promise that rejects on timeout
+     */
+    withTimeout(promise, timeoutMs) {
```
jonathanKingston, line 876: We may wish to move this to the messaging layer for any await'd requests at least for Windows. For now this is the most prevalent message we're triggering in web context Windows.

shakyShane (approved):
looks good, and I agree it might warrant being in the messaging layer. my original implementation allows passing an abort signal as part of an options object on .request() calls, so we should probably go down that route. 

approving for now, thanks

jonathanKingston:
> looks good, and I agree it might warrant being in the messaging layer. my original implementation allows passing an abort signal as part of an options object on .request() calls, so we should probably go down that route.
> 
> approving for now, thanks

As we discussed, we quickly tried this. Some of the wiring is there only for Windows but it's unclear if it's working even if we wire in the opts to the transport.
The immediate concern is fixing this and there were concerns with this design anyway.

PR #2052, opened 2025-11-10
detector service for breakage reporting

jdorweiler:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/72649045549333/task/1211666118504301?focus=true

## Description

Building on #2048 with a more generalize framework for breakage reports.

**New detector service:**
- Registry + caching with configurable TTL
- Auto-registration from remote config during page load
- Optional auto-run with configurable delay (default 100ms)
- Domain gating + custom `shouldRun()` checks (only during auto-run)
- Multi-consumer design (any feature can import and use)

**Changed:**
- Architecture: Service with dynamic registration
- Caching: Built-in with TTL vs re-scan every call
- Configuration: Remote config vs function parameters
- Lifecycle: Auto-registers on page load + optional auto-run vs on-demand only
- Gating: Domain restrictions + custom checks
- Consumers: Any feature can use service

See the [injected/src/detectors/README.md](./injected/src/detectors/README.md) for more details 

## Testing Steps

- 

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> Introduce on-demand bot/fraud detectors and wire them into breakage reporting, enable new feature across platforms, and add integration tests with mock pages.
> 
> - **Detectors**:
>   - Add on-demand utilities `runBotDetection` and `runFraudDetection` with shared helpers in `injected/src/detectors/**` and fallback `default-config`.
>   - New feature wrapper `injected/src/features/web-interference-detection.js` exposing `detectInterference` via messaging.
> - **Breakage Reporting**:
>   - Enhance `injected/src/features/breakage-reporting.js` to optionally include `detectorData` using `webInterferenceDetection` settings.
> - **Feature Enablement**:
>   - Register `webInterferenceDetection` in `injected/src/features.js` and enable on `apple`, `android`, `windows`, `chrome`, and `chrome-mv3`.
> - **Tests & Pages**:
>   - Expand `injected/integration-test/breakage-reporting.spec.js` with cases for clean page, Cloudflare, reCAPTCHA, and PerimeterX.
>   - Add test config enabling detectors and mock pages under `injected/integration-test/test-pages/breakage-reporting/**`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7d8619cdb2e752f07cba2da5f9cae4c1eb81485e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/injected/integration-test/breakage-reporting.spec.js
+++ b/injected/integration-test/breakage-reporting.spec.js
@@ -77,9 +77,7 @@ test.describe('Breakage Reporting Feature', () => {
         expect(result.params?.detectorData).toBeDefined();
         expect(result.params?.detectorData?.botDetection.detected).toBe(true);
 
-        const recaptchaResult = result.params?.detectorData?.botDetection.results.find(
-            r => r.challengeType === 'recaptcha'
-        );
+        const recaptchaResult = result.params?.detectorData?.botDetection.results.find((r) => r.challengeType === 'recaptcha');
         expect(recaptchaResult).toBeDefined();
         expect(recaptchaResult.vendor).toBe('reCAPTCHA');
         expect(recaptchaResult.detected).toBe(true);
--- a/injected/src/detectors/default-config.js
+++ b/injected/src/detectors/default-config.js
@@ -25,12 +25,7 @@ export const DEFAULT_DETECTOR_SETTINGS = Object.freeze({
         hcaptcha: {
             state: 'enabled',
             vendor: 'hcaptcha',
-            selectors: [
-                '.h-captcha',
-                '[data-hcaptcha-widget-id]',
-                'script[src*="hcaptcha.com"]',
-                'script[src*="assets.hcaptcha.com"]',
-            ],
+            selectors: ['.h-captcha', '[data-hcaptcha-widget-id]', 'script[src*="hcaptcha.com"]', 'script[src*="assets.hcaptcha.com"]'],
             windowProperties: ['hcaptcha'],
         },
     },
--- a/injected/src/detectors/detections/bot-detection.js
+++ b/injected/src/detectors/detections/bot-detection.js
@@ -42,4 +42,3 @@ function findStatus(statusSelectors) {
 
     return match?.status ?? null;
 }
-
--- a/injected/src/detectors/detections/fraud-detection.js
+++ b/injected/src/detectors/detections/fraud-detection.js
@@ -9,8 +9,7 @@ export function runFraudDetection(config = {}) {
         .filter(([_, alertConfig]) => alertConfig?.state === 'enabled')
         .map(([alertId, alertConfig]) => {
             const detected =
-                checkSelectorsWithVisibility(alertConfig.selectors) ||
-                checkTextPatterns(alertConfig.textPatterns, alertConfig.textSources);
+                checkSelectorsWithVisibility(alertConfig.selectors) || checkTextPatterns(alertConfig.textPatterns, alertConfig.textSources);
             if (!detected) {
                 return null;
             }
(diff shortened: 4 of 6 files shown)


injected/src/detectors/detections/detection-base.js
madblex: Will we still be needing this?

injected/src/detectors/detector-service.js
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Get data from multiple detectors in a single call
+ * @param {string[]} detectorIds - Array of detector IDs
+ * @param {Object} [options] - Options for data retrieval
+ * @param {number} [options.maxAgeMs] - Maximum age of cached data in milliseconds
+ * @param {boolean} [options._autoRun] - Internal flag indicating auto-run (gates apply)
+ * @returns {Promise<Record<string, any>>} Object mapping detector IDs to their data
+ */
+export async function getDetectorBatch(detectorIds, options = {}) {
```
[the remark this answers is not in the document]
jdorweiler: Yeah good point. I'll go with `getDetectorsData`. I don't think we need web prefix as we're not following that naming scheme elsewhere (e.g., getJsPerformanceMetrics, getExpandedPerformanceMetrics)

madblex:
> > This looks great. I added some comments and a couple more here:
> > ```
> > 1. I'd like to see how this integrates with content features, especially broker protection, i.e. the equivalent of [this](https://github.com/duckduckgo/content-scope-scripts/pull/2048/files#r2513656558)
> > 
> > 2. I see we removed `WebInterferenceDetection` content feature. Is it because you don't think we'll need a standalone feature for this? i.e. the detectors will **only** be called from other features.
> > 
> > 3. I see we're going the functional route (from classes). Not feeling strongly (I prefer functional most of the time), I just believe the trade-off is that interfaces will be more difficult to enforce. I think classes provide a bit of a better structure, but this is opinionated.
> > ```
> 
> 1. Commented here https://github.com/duckduckgo/content-scope-scripts/pull/2048/files#r2514564110
> 2. Yeah, I changed it so detectors auto register via detector-init.js during the load phase, and features import the service directly. So detector initialization is handled for you. Then you just need to add the message handler to broker protection (see my comment linked in 1)
> 3. I'm not strongly attached to this approach. I wanted to see what functional approach looked like vs classes but I'm open to switching to classes if you think the structure is worth it.

1. 👍 
2. Fine for me. We can always add a dedicated feature later if needed
3. No need to do the extra work. This will likely work fine as well. Thanks.

injected/src/content-scope-features.js
```diff
@@ -44,6 +45,14 @@ export function load(args) {
 
     const bundledFeatureNames = typeof importConfig.injectName === 'string' ? platformSupport[importConfig.injectName] : [];
 
+    // Initialize detectors early so they're available when features init
```
jonathanKingston: I don't think we should put it that deep into the framework. Can we not trigger it in 'load' and whomever triggers firsts kicks it off for both?

injected/src/detectors/detector-init.js
```diff
@@ -0,0 +1,133 @@
...
+            }
+
+            // All gates passed, run the detector
+            return await getData(options);
+        },
+        refresh: refresh ? async (options) => {
+            if (!checkGates(options, domains, shouldRun)) {
+                return null;
+            }
+            return await refresh(options);
```
jonathanKingston: I don't think bot-detection or fraud-detection has refresh but this I believe exposes it as having one meaning we will preference refresh right?

In:
```
    const runner = registration.refresh ?? registration.getData;
```
We're preferencing this function which is a noop as I understand it.

jdorweiler:
@jonathanKingston Thanks for the review. I refactored this into a WebInterferenceDetection ContentFeature, got rid of the standalone service/init, and switched to using getFeatureSetting() conditionalChanges for config. Let me know what you think.

PR #2242, opened 2026-02-06
Allow Regex in Separator

brianhall:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1213147669247609

## Description

Privacy Config schema update: https://github.com/duckduckgo/privacy-configuration/pull/4498

## Testing Steps

1. Drag & Drop content-scope-scripts into XCode, build the browser.
2. Open PIR Debugging mode and enter some profile information
3. Copy [this JSON](https://dub.duckduckgo.com/duckduckgo/dbp-api/blob/d5fdd8e1dceceaeae3fcdc65a507ae5b1624d6e2/dbp-json/data/json/privatereports.com.json) into the lower-right textbox of the debugger and hit scan.
4. The profiles returned should have proper city/state pairs (vs. current production which returns them all as a one big string).

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Medium Risk**
> Changes core scraping/extraction parsing for list fields; mis-specified patterns or regex behavior could alter how broker data is split across configs, though added unit tests reduce the likelihood of regressions.
> 
> **Overview**
> **Broker Protection extraction now supports regex-based list splitting via config.** The `separator` field can be provided as a JSON-safe string in the form `"/pattern/"`, which is converted to a `RegExp` and passed to `String#split` (otherwise behavior remains string/default-separator based).
> 
> Unit tests are expanded to cover regex and non-regex separator behavior for `addressCityStateList` and `relativesList`, including a case showing that non-`"/…/"` regex-like strings are treated as plain strings. `CODEOWNERS` is updated to add ownership for `injected/unit-test/broker-protection*`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6e05900612a4b8d0dc81a82109c057b6b3290290. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

---
  - https://app.asana.com/0/0/1213527430578581

### DIFF
--- a/injected/unit-test/broker-protection-extract.js
+++ b/injected/unit-test/broker-protection-extract.js
@@ -366,6 +366,11 @@ describe('create profiles from extracted data', () => {
                     },
                 },
                 elements: [{ innerText: 'Alice Smith, Bob Jones, Carol Lee' }],
+                /*
+                    This result is a bit unintuitive. First, the regex is invalid, so we treat it as a string (and return no matches).
+                    Then, relativesMatcher splits the string by , and returns the first result. This is primarily to work around a few
+                    brokers that listed relatives in the following format: "Relative Name, Age"
+                */
                 expected: {
                     relatives: ['Alice Smith'],
                 },


injected/unit-test/broker-protection-extract.js
```diff
@@ -163,28 +163,71 @@ describe('create profiles from extracted data', () => {
...
-                addresses: [
-                    { city: 'Carrollton', state: 'TX' },
-                    { city: 'Dallas', state: 'TX' },
-                    { city: 'The Colony', state: 'TX' },
-                ],
+            {
+                selectors: {
+                    addressCityStateList: {
+                        selector: 'example',
+                        separator: '/(?<=, [A-Z]{2}), /',
```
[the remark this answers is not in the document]
brianhall: Changing the format of the actual JSON instructions would be too much of a lift (requiring native changes across platforms), but I think we might be able to find a way to do this in a bit of a roundabout way so in order to contribute upstream. Will work on this and let you know.

injected/src/utils.js
```diff
@@ -321,7 +321,7 @@ const functionMap = {
 
 /**
  * @typedef {object} ConfigSetting
- * @property {'undefined' | 'number' | 'string' | 'function' | 'boolean' | 'null' | 'array' | 'object'} type
+ * @property {'undefined' | 'number' | 'string' | 'function' | 'boolean' | 'null' | 'array' | 'object' | 'regexp'} type
```
brianhall: I think I'm missing where we import this from [CSSConfigSettingSingle](https://github.com/duckduckgo/privacy-configuration/blob/c8d3b9f1d90e72023f8d3bd606448b0992441cb5/schema/feature.ts#L96) as I had to make the change here as well in order for it to work. Am I missing some sort of import process @jonathanKingston?

injected/unit-test/broker-protection-extract.js
```diff
@@ -290,6 +333,65 @@ describe('create profiles from extracted data', () => {
...
+        const elementExamples = [
+            {
+                selectors: {
+                    relativesList: {
+                        selector: 'example',
+                    },
+                },
+                elements: [{ innerText: 'Alice Smith • Bob Jones • Carol Lee' }],
+                expected: {
+                    relatives: ['Alice Smith', 'Bob Jones', 'Carol Lee'],
```
brianhall: Default separator is `•`

kdzwinel (approved):
Nice, LGTM! Please make sure to scope Jonathan's improvement ideas as a follow up

PR #2328, opened 2026-02-27
Ensure BrowserUiLock C-S-S is calling native side at least once on every page load.

aibrahim-:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1212608036467427/task/1213415771791498

## Description

Quick navigation between pages can result in race-conditions between the message arriving on the native side, and the actual locking taking place. So we always ensure the first lock state evaluation is sent to native side. 

Also use a configurable set of overflowy types to check and include 'auto' by default, which matches iOS behaviour.
## Testing Steps

same as here: https://github.com/duckduckgo/Android/pull/7776

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Medium Risk**
> Changes the criteria for detecting scrollbars (treating `overflow-y: auto` as non-visible by default) and forces an initial `uiLockChanged` message, which could affect gesture-lock behavior on some pages and native-side state sync.
> 
> **Overview**
> Ensures the native layer always receives an initial `uiLockChanged` notification on page load by initializing `_currentLockState` to `null` (so the first evaluation notifies even when unlocked).
> 
> Adjusts scrollbar detection by renaming `_hasVisibleScrollbar` to `_hasExplicitlyVisibleScrollbar` and making the set of `overflow-y` values treated as *not showing* a visible scrollbar configurable via `overflowTypes` (defaulting to `['hidden','clip','auto']`). Unit tests are updated to reflect the new defaults and first-notification behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 23e2a5a2d01ba448f8b29432f0c4c75682e9c100. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.github/scripts/review-helpers.mjs
+++ b/.github/scripts/review-helpers.mjs
@@ -50,7 +50,10 @@ export async function isTeamMember(github, org, teamSlug, username) {
             username,
         });
         return data.state === 'active';
-    } catch {
+    } catch (error) {
+        if (error.status === 401) {
+            throw error;
+        }
         return false;
     }
 }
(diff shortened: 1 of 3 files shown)


injected/src/features/browser-ui-lock.js
```diff
@@ -54,6 +57,15 @@ export default class BrowserUiLock extends ContentFeature {
      */
     _startObserving() {
         this._scheduleEvaluation();
+        // Cancel any pending delayed evaluation from a previous navigation
+        if (this._delayedEvalTimerId !== null) {
+            clearTimeout(this._delayedEvalTimerId);
+        }
+        // Re-evaluate after a short delay to catch late-rendering content (e.g. back navigation)
```
jonathanKingston: `urlChanged` should catch this if we move the delayed timer check to `_scheduleEvaluation`.

However so should the mutation observer, are we missing something there?

Is the main value here just setting the null state?

PR #2676, opened 2026-04-29
make overflow-y required to lock, and add a setting for locked sites

aibrahim-:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1212608036467427/task/1214187547786011?focus=true

## Description
outlook.com was getting locked incorrectly, we decided to be overly restrictive to lock by also requiriing `overflow-y:hidden`, while also adding a setting for sites that should be locked regardless. 

## Testing Steps
- sites that shouldn't lock (e.g. bbc.com, google searches, any site).
- sites that should lock (e.g. windy.com, maps.google.com, https://wafflegame.net/daily)
- SERP should always be unlocked.

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Medium Risk**
> Changes a user-visible heuristic that controls when browser gestures are disabled, which can regress locking/unlocking behavior across many sites. Risk is mitigated by conservative defaults and a config override to force-lock known cases.
> 
> **Overview**
> Tightens `BrowserUiLock` detection so pages only lock when **no visible scrollbar exists** *and* `html` or `body` explicitly sets a locking `overflow-y` value (configurable via `overflowTypes`, defaulting to `['hidden','clip']`), reducing false-positive locks.
> 
> Adds an `isLockedPage` feature setting that **forces lock** for configured sites/pages (via privacy-config), bypassing all other heuristics, and updates unit tests to reflect the new default `overflowTypes` behavior plus coverage for the new override and `_hasLockingOverflowY`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a024c960b58fa75631d9b0041e37f550d6af2c9e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/injected/src/features/browser-ui-lock.js
+++ b/injected/src/features/browser-ui-lock.js
@@ -130,11 +130,18 @@ export default class BrowserUiLock extends ContentFeature {
 
     /**
      * Determine if UI should be locked based on scrollbar visibility or content type.
-     * Lock if the page is a direct image display or has no visible vertical scrollbar.
+     * Lock if the site's domain is in the `lockedDomains` list; otherwise lock if
+     * the page is a direct image display, or if there is no visible vertical
+     * scrollbar AND `overflow-y: hidden` is explicitly set on html or body.
      * @returns {boolean}
      */
     _detectShouldLock() {
         try {
+            // Sites configured in lockedDomains bypass all other checks
+            if (this._isLockedDomain()) {
+                return true;
+            }
+
             // Image display pages (navigating directly to an image URL) should lock
             if (this.getFeatureSettingEnabled('lockImagePages', 'enabled') && this._isImageDisplayPage()) {
                 return true;
@@ -151,8 +158,9 @@ export default class BrowserUiLock extends ContentFeature {
                 return false;
             }
 
-            // No visible scrollbar - lock the UI
-            return true;
+            // No visible scrollbar — additionally require overflow-y: hidden
+            // to be explicitly set on html or body before locking.
+            return Boolean((html && this._hasOverflowYHidden(html)) || (body && this._hasOverflowYHidden(body)));
         } catch (e) {
             // Fail open - return false (unlocked) on error
             this.log.warn('Failed to detect scroll state:', e);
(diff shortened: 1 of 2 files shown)


injected/src/features/browser-ui-lock.js
```diff
@@ -170,6 +178,41 @@ export default class BrowserUiLock extends ContentFeature {
...
+        if (!Array.isArray(patterns) || patterns.length === 0) {
+            return false;
+        }
+        const siteUrl = this.args?.site?.url;
+        if (typeof siteUrl !== 'string' || siteUrl.length === 0) {
+            return false;
+        }
+        let hostPath;
+        try {
+            const url = new URL(siteUrl);
```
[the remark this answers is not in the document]
aibrahim-: @cursoragent fix it

injected/src/features/browser-ui-lock.js
```diff
@@ -170,6 +180,41 @@ export default class BrowserUiLock extends ContentFeature {
...
+     *
+     * - exact host match: an entry `"example.com"` matches host `example.com` (any
+     *   path) but does NOT match `www.example.com` or `evil.example.com.attacker`.
+     * - path prefix match: an entry `"example.com/foo"` matches any URL on host
+     *   `example.com` whose path begins with `/foo` followed by `/` or end-of-path.
+     *
+     * @returns {boolean}
+     */
+    _isLockedDomain() {
+        const patterns = this.getFeatureSetting('lockedDomains');
```
jonathanKingston: You could significantly reduce this method by making this use: `getFeatureSettingEnabled` and instead use `patchSettings` in the config to apply conditionally based on hostname/path etc.

In fact I'd just change:
```
if (this._isLockedDomain()) {
```

To:
```
if (this.getFeatureSettingEnabled('isLockedPage')) {
```

injected/src/features/browser-ui-lock.js
```diff
@@ -179,10 +224,19 @@ export default class BrowserUiLock extends ContentFeature {
     _hasExplicitlyVisibleScrollbar(el) {
         const style = getComputedStyle(el);
         const overflowY = style.overflowY;
-        const overflowTypes = this.getFeatureSetting('overflowTypes') ?? ['hidden', 'clip', 'auto'];
+        const overflowTypes = this.getFeatureSetting('overflowTypes') ?? ['hidden', 'clip'];
```
jonathanKingston: Didn't we drop this from the config? Do we need to drop it here?

injected/src/features/browser-ui-lock.js
```diff
@@ -151,8 +160,9 @@ export default class BrowserUiLock extends ContentFeature {
                 return false;
             }
 
-            // No visible scrollbar - lock the UI
-            return true;
+            // No visible scrollbar — additionally require overflow-y: hidden
+            // to be explicitly set on html or body before locking.
+            return Boolean((html && this._hasOverflowYHidden(html)) || (body && this._hasOverflowYHidden(body)));
```
jonathanKingston: It seems like it should be conflict-driven, as well. That way, we can respond to changes. Also, we could selectively apply this to different times.

I think Cursor's comment is pretty correct, to be honest.

injected/src/features/browser-ui-lock.js
aibrahim-: what would configs look like?

injected/src/features/browser-ui-lock.js
aibrahim-: ideally the configs should be there only if needed, we can change it here and remove the config.

injected/src/features/browser-ui-lock.js
aibrahim-: what do you mean by conflict-driven?

jonathanKingston: *config driven

injected/src/features/browser-ui-lock.js
jonathanKingston: https://github.com/duckduckgo/privacy-configuration/blob/d7c5b76a520411933b274f658c1570729d8b0182/overrides/android-override.json#L976-L996 this kinda thing, anything can be changed conditionally based on conditions, which there's a few we support.

injected/src/features/browser-ui-lock.js
jonathanKingston: Don't mind, just checking this was intended.

injected/src/features/browser-ui-lock.js
```diff
@@ -151,15 +161,25 @@ export default class BrowserUiLock extends ContentFeature {
                 return false;
             }
 
-            // No visible scrollbar - lock the UI
-            return true;
+            // No visible scrollbar — additionally require a locking overflow-y
+            // value to be explicitly set on html or body before locking.
+            return Boolean((html && this._hasLockingOverflowY(html)) || (body && this._hasLockingOverflowY(body)));
```
jonathanKingston, line 166: Are we explicitly giving up on overflow x then?

jonathanKingston (approved):
I kinda think we should still support overflow x but not too precious over it. I know this is mostly best effort anyway.

injected/src/features/browser-ui-lock.js
aibrahim-: I don't think it was ever really considered. Also none of our current usecases need it.

PR #2729, opened 2026-05-28
Capture message handler refs at construction on WebKit

laghee:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1212684734587032?focus=true

## Description

The WebkitMessagingTransport previously only captured handler references at construction time on legacy WebKit (macOS Catalina, < 11). On modern WebKit it re-read `window.webkit.messageHandlers[handler]` on every send. With this change, the `apiManipulation` feature can be used to install a prototype-side getter that nullifies `messageHandlers` for site JS. Without this change, any page-world C-S-S feature that calls notify/request/subscribe after the nullify would synchronously throw inside wkSend reading undefined[handler] and silently degrade because the surrounding polyfills (e.g., webCompat Notification) catch the throw and return defaults.

After this change:
- captureWebkitHandlers always runs in the constructor
- WebkitMessagingConfig: dropped hasModernWebkitAPI and secret params/fields
- wkSend / wkSendAndWait / captureWebkitHandlers: dropped if (!hasModernWebkitAPI) branches
- Original postMessage is left in place so any other code reading through window.webkit.messageHandlers continues to see the host binding's normal shape.
- Removed helper methods only used by the legacy crypto flow (generateRandomMethod, randomString, createRandMethodName, algoObj, createRandKey, createRandIv, decryptResponse)
- Removed SecureMessagingParams class (internal-only, not re-exported)
- Removed 10 now-unused captured-global imports
- Updated the 3 in-tree callers (apple.js, create-special-page-messaging.js, webkit.example.js) to stop passing the now-removed params
- Removed the legacy-WebKit describe block in the unit tests; modern tests simplified to drop the unused config params

Adds unit tests covering:
- capture + routing through captured ref
- leaves original postMessage in place
- transport survives wholesale replacement of window.webkit.messageHandlers
- throws MissingHandler when a handler was never registered

## Testing Steps

- 

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Medium Risk**
> Changes the native bridge send path for all Apple WebKit injects; behavior is intentional for privacy hardening but any missed handler capture or platform still on legacy WebKit would break messaging.
> 
> **Overview**
> **WebKit messaging** now **captures** `messageHandlers` at transport construction and **sends** only through that cache, so content-scope features keep talking to native after site hardening (e.g. `apiManipulation` nullifying `window.webkit.messageHandlers`).
> 
> **`wkSend`** uses a **null-prototype** handler cache and **captured `ReflectApply`** on stored `{ handler, postMessage }` pairs (no per-send lookup, no `.bind`, host `postMessage` left on `window.webkit`). **Legacy Catalina** encrypted/async reply path, **`SecureMessagingParams`**, and **`hasModernWebkitAPI` / `secret`** on **`WebkitMessagingConfig`** are **removed**; Apple entry points and special-pages wiring only pass handler names.
> 
> **New unit tests** cover capture/routing, survival after `messageHandlers` replacement, missing handlers, prototype pollution, tampered `Object.create`, and correct `this` on dispatch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8f93c796a4a285b519ae2ced1fde5528ee1c9b42. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/messaging/lib/webkit.js
+++ b/messaging/lib/webkit.js
@@ -6,7 +6,13 @@
 import { MessagingTransport, MissingHandler } from '../index.js';
 import { isResponseFor, isSubscriptionEventFor } from '../schema.js';
 import { ensureNavigatorDuckDuckGo } from '../../injected/src/navigator-global.js';
-import { JSONparse, Error as _Error, ReflectDeleteProperty, objectDefineProperty } from '../../injected/src/captured-globals.js';
+import {
+    JSONparse,
+    Error as _Error,
+    ReflectApply,
+    ReflectDeleteProperty,
+    objectDefineProperty,
+} from '../../injected/src/captured-globals.js';
 
 /**
  * @example
@@ -25,9 +31,16 @@ export class WebkitMessagingTransport {
     /**
      * Null-prototype cache so a hostile page that pollutes `Object.prototype`
      * cannot supply a callable from there if `capture` ever misses a handler.
-     * @type {Record<string, any>}
+     *
+     * Uses the `{ __proto__: null }` literal rather than `Object.create(null)`
+     * because the latter is a method dispatch through `globalThis.Object`, which
+     * page JS could replace before this class field runs if transport
+     * construction is deferred (`Messaging` is lazy on `ContentFeature.messaging`).
+     * The `__proto__: null` literal is a syntactic construct, not method
+     * dispatch, so it always yields a true null-prototype object.
+     * @type {Record<string, { handler: any, postMessage: Function }>}
      */
-    capturedWebkitHandlers = Object.create(null);
+    capturedWebkitHandlers = /** @type {any} */ ({ __proto__: null });
 
     /**
      * @param {WebkitMessagingConfig} config
@@ -56,10 +69,12 @@ export class WebkitMessagingTransport {
      */
     wkSend(handler, data = {}) {
         const captured = this.capturedWebkitHandlers[handler];
-        if (typeof captured !== 'function') {
+        if (!captured || typeof captured.postMessage !== 'function') {
             throw new MissingHandler(`Missing webkit handler: '${handler}'`, handler);
         }
-        return captured(data);
+        // Use the captured ReflectApply so the send path doesn't go through
+        // any page-mutable function (`.call`, `.bind`, etc.).
+        return ReflectApply(captured.postMessage, captured.handler, [data]);
     }
 
     /**
@@ -108,20 +123,26 @@ export class WebkitMessagingTransport {
      * `window.webkit.messageHandlers` (e.g. by privacy hardening that nullifies
      * the namespace for site JS to reduce fingerprinting surface).
      *
+     * Stores the handler object and its `postMessage` function as a pair so
+     * `wkSend` can dispatch via the captured `ReflectApply` rather than calling
+     * `.bind()` here. `.bind` is a method on the page-mutable
+     * `Function.prototype` — if transport construction is deferred (`Messaging`
+     * is lazy on `ContentFeature.messaging`) page JS could replace
+     * `Function.prototype.bind` first and have the cache store an attacker-
+     * controlled function. Storing the unbound pair sidesteps that.
+     *
      * @param {string[]} handlerNames
      */
     captureWebkitHandlers(handlerNames) {
         const handlers = window.webkit.messageHandlers;
         if (!handlers) throw new MissingHandler('window.webkit.messageHandlers was absent', 'all');
         for (const webkitMessageHandlerName of handlerNames) {
-            if (typeof handlers[webkitMessageHandlerName]?.postMessage === 'function') {
-                /**
-                 * `bind` is used here to ensure future calls to the captured
-                 * `postMessage` have the correct `this` context
-                 */
-                const original = handlers[webkitMessageHandlerName];
-                const bound = handlers[webkitMessageHandlerName].postMessage?.bind(original);
-                this.capturedWebkitHandlers[webkitMessageHandlerName] = bound;
+            const handler = handlers[webkitMessageHandlerName];
+            if (typeof handler?.postMessage === 'function') {
+                this.capturedWebkitHandlers[webkitMessageHandlerName] = {
+                    handler,
+                    postMessage: handler.postMessage,
+                };
             }
         }
     }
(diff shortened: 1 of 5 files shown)


messaging/lib/webkit.js
```diff
@@ -84,9 +84,14 @@ export class WebkitMessagingTransport {
...
-            this.captureWebkitHandlers(this.config.webkitMessageHandlerNames);
-        }
+        // Capture handler references at construction on both legacy and modern WebKit.
+        // On modern WebKit this previously read `window.webkit.messageHandlers[handler]`
+        // on every send, which means the transport silently breaks if site-level
+        // privacy hardening (e.g. apiManipulation-driven nullification of
+        // `window.webkit.messageHandlers` to reduce fingerprinting surface) replaces
+        // the namespace after init. Capturing once at construction makes the transport
+        // resilient to those changes.
+        this.captureWebkitHandlers(this.config.webkitMessageHandlerNames);
```
jonathanKingston, line 59: Can you double check the config here to make sure that there isn't any APIs that are missing here? The hasModernwebkitAPI is basically been dead for a considerable amount of time, so it could be that that code path was now broken.

messaging/lib/webkit.js
```diff
@@ -282,8 +284,18 @@ export class WebkitMessagingTransport {
...
+     *
+     * Used on both modern and legacy WebKit:
+     * - **Legacy (macOS Catalina, < 11)**: capture is required for the secure
+     *   messaging protocol, and the original `postMessage` is deleted to prevent
+     *   page-script JS from invoking it directly without the encrypted envelope.
+     * - **Modern**: capture makes the transport resilient to later removal or
+     *   replacement of `window.webkit.messageHandlers` by privacy hardening
+     *   (e.g. apiManipulation-driven nullification to reduce fingerprinting
+     *   surface). The original is left in place so callers reading directly
+     *   from the namespace continue to see the host binding's normal shape.
```
jonathanKingston: I'd slim down this comment. I don't think we need to mention legacy anymore.

messaging/lib/webkit.js
```diff
@@ -299,7 +311,9 @@ export class WebkitMessagingTransport {
                 const original = handlers[webkitMessageHandlerName];
                 const bound = handlers[webkitMessageHandlerName].postMessage?.bind(original);
                 this.capturedWebkitHandlers[webkitMessageHandlerName] = bound;
-                delete handlers[webkitMessageHandlerName].postMessage;
+                if (!this.config.hasModernWebkitAPI) {
```
jonathanKingston: I wonder if we should just remove hasModernWebkitAPI. I think we don't support that version of macOS anymore.

laghee: Probably, yeah. I thought it said < or =, but it's just < 11 (and now we're in the process of dropping 11 too).

messaging/lib/webkit.js
```diff
@@ -98,7 +103,8 @@ export class WebkitMessagingTransport {
      * @internal
      */
     wkSend(handler, data = {}) {
-        if (!(handler in window.webkit.messageHandlers)) {
+        const captured = this.capturedWebkitHandlers[handler];
+        if (typeof captured !== 'function') {
```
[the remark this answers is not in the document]
laghee: Switched `capturedWebkitHandlers` to `Object.create(null)` so the cache has no Object.prototype chain to walk, and added a unit test (ignores Object.prototype pollution when an uncaptured handler is looked up) that pollutes Object.prototype.hostilePollutionHandler with a callable and asserts wkSend throws MissingHandler rather than invoking the polluted function with the outgoing native payload.

messaging/lib/webkit.js
laghee: Ran a quick check, and yes, the config should be complete by construction as wkSend is only ever called with msg.context:

    `async notify(msg) {
        await this.wkSend(msg.context, msg);
    }`

request, wkSendAndWait and subscribe all route msg.context through wkSend. There's no caller that constructs an arbitrary handler-name string and passes it.

messaging/lib/webkit.js
```diff
@@ -299,7 +122,6 @@ export class WebkitMessagingTransport {
                 const original = handlers[webkitMessageHandlerName];
                 const bound = handlers[webkitMessageHandlerName].postMessage?.bind(original);
```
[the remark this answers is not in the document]
laghee: Removed .bind() from the capture path entirely. The cache now stores a { handler, postMessage } pair and wkSend dispatches via the already-captured ReflectApply (which is Reflect.apply.bind(Reflect) captured at module-load time in captured-globals.js, before any page JS can run).
Two regression tests:
- stores handler and postMessage as a separate pair, not a single bound function — structural assertion that the cache holds raw references rather than a bound copy (a bound function would not be reference-equal to the original postMessage). Avoided directly tampering Function.prototype.bind in a test because Jasmine's spy/expect machinery uses .bind internally and the recursion blows the call stack — so the structural assertion is the safe equivalent.
- dispatches through the captured handler with the correct \this` even without `.bind` — asserts host-binding semantics are preserved (thisinside the capturedpostMessageis still the handler object), which proves the send path goes throughReflectApply` rather than a bare unbound call.

messaging/lib/webkit.js
```diff
@@ -74,8 +74,12 @@ import {
...
  */
 export class WebkitMessagingTransport {
-    /** @type {Record<string, any>} */
-    capturedWebkitHandlers = {};
+    /**
+     * Null-prototype cache so a hostile page that pollutes `Object.prototype`
+     * cannot supply a callable from there if `capture` ever misses a handler.
+     * @type {Record<string, any>}
+     */
+    capturedWebkitHandlers = Object.create(null);
```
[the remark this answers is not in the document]
laghee: Switched the cache initializer from Object.create(null) to the { __proto__: null } literal. The literal is a syntactic construct rather than a method dispatch, so it always produces a true null-prototype object regardless of any page-side Object.create replacement. Added a regression test (cache is not derived from a tampered Object.create) that replaces Object.create with a tampered factory before constructing the transport and asserts the cache still rejects unknown handler names with MissingHandler.

jonathanKingston (approved):

PR #2768, opened 2026-06-15
Add `attribute` and `source: pageUrl` support to broker-protection extract

noisysocks:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/72649045549333/task/1213630318520581

## Description

Our `extract` action could only read a profile's `profileUrl` from an element's text or an `<a href>`. Some brokers don't play along: Veripages keeps the profile link in a `data-link` attribute on a `<span>` (no anchor at all), and other layouts only expose the profile's identity in the page URL.

So this adds two new value sources for an extracted field:

- `attribute` – read a named attribute (e.g. `"attribute": "data-link"`) instead of the element's text.
- `source: "pageUrl"` – read from the current page URL instead of the DOM.

No behaviour change for existing rules. The new paths only kick in when a broker rule opts into them, and the value still flows through the existing extractor pipeline (`identifier`, `identifierType`, etc.) unchanged.

## Testing Steps

1. Clone `apple-browsers` and point to this branch in `Package.swift`:
  ```swift
  .package(url: "https://github.com/duckduckgo/content-scope-scripts.git", branch: "pr-releases/randerson/broker-extract-attribute-pageurl")
  ```
2. Build and run the macOS browser. (You may need clean or delete derived data to force package update.)
3. Debug → Personal Information Removal → Run Personal Information Removal Debug Mode.
4. Select Veripages and change the `extract` rule step's `profileUrl` to:
  ```json
  "profileUrl": {
    "attribute": "data-link",
    "identifier": "https://veriforia.com/view/${id}",
    "identifierType": "path",
    "selector": ".search-item-title-link"
  }
  ```
5. Run it, and confirm the extracted `profileUrl` comes from the `data-link` attribute.

Automated coverage: `npm run test-unit` and `npm run test-int` (broker-protection specs) exercise both `attribute` and `source: "pageUrl"`, including a trimmed real-Veripages fixture.

## Checklist

*Please tick all that apply:*

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes broker profile extraction and URL/identifier handling for Personal Information Removal; behavior is gated by per-broker extract rules but misconfigured rules could yield wrong profile URLs or identifiers.
> 
> **Overview**
> Broker-protection **extract** can populate profile fields without element text: **`attribute`** reads a named DOM attribute (e.g. `data-link`), and **`source: "pageUrl"`** uses `location.href` for fields like `profileUrl` without a selector.
> 
> **`ProfileUrlExtractor`** now resolves relative URLs against the current page (same as `<a href>`), and **`createProfile`** routes fields through a shared `stringsForProfileField` path so selector-less sources still run the existing extractors. Selector-based rules are unchanged unless they opt into the new options.
> 
> Coverage adds unit tests for attribute/pageUrl behavior and Playwright cases with Veripages-style fixtures and synthetic page-URL pages.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3e3f4c69f9883519a3607496a518e12aad712e3f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


daxtheduck (dismissed):

daxtheduck (dismissed):

injected/src/features/broker-protection/actions/extract.js
```diff
@@ -169,7 +182,9 @@ export function stringValuesFromElements(elements, key, extractField) {
     return elements.map((element) => {
         let elementValue;
 
-        if ('innerText' in element) {
+        if (extractField?.attribute && 'getAttribute' in element) {
+            elementValue = element.getAttribute?.(extractField.attribute) ?? null;
```
brianhall, line 186: Just one thing to think about here - I noticed that the `data-link` attribute on Veripages only returns a partial with no baseURL. Thankfully the opt out page on Veripages accepts this, but we might run across cases where the full url is required.

If prepending the base url is easy enough, let's do it, otherwise I'm happy for us to punt and add an item to the Data Brokers AOR backlog.

brianhall (approved):
I tested in-browser (against Veripages) and confirmed both new methods. I added one note that's optional, and otherwise this LGTM.

injected/src/features/broker-protection/actions/extract.js
noisysocks: Good shout. Made it so that the attribute value now resolves to an absolute URL against the page, the same way the browser resolves an `<a href>`. So Veripages' `/profile/John-Smith/BMFrB9EB` comes through as the full `https://…` URL, for both `profileUrl` and `identifier`. Rules that read from a real `<a href>` are unchanged since those were already absolute.

PR #2777, opened 2026-06-18
Support `/pattern/i` regex in broker-protection extract afterText/beforeText/separator

noisysocks:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1209121419454298/task/1213630318520583

## Description

`separator` already lets a broker rule pass a regex via the `/pattern/` convention (#2242), but `afterText` and `beforeText` were still plain literal matches, and none of the three could match case-insensitively. We want these fields to be as flexible as possible, so that broker rules can adapt to whatever shape a page throws at them without needing a code change.

So this extends the same `/pattern/` convention to `afterText` and `beforeText`, and adds an optional `i` flag across all three:

```json
"age":                  { "selector": ".age", "afterText": "/age:?\\s+/i" },
"alternativeNamesList": { "selector": ".aka", "separator": "/\\s*aka\\s*/i" }
```

Plain strings stay literal, so there's no behaviour change for existing rules – the regex path only kicks in when a value is wrapped in `/.../`. Anything with an unsupported flag (i.e. not `i`) falls back to being treated as a literal string rather than throwing, so a mistyped config fails gently rather than breaking the whole action.

## Testing Steps

There's no broker that needs this yet, so there's no good way to test it by hand. Coverage is via tests instead:

- `npm run test-int` (broker-protection specs) → `extract with regex afterText / beforeText / separator`. Runs against a synthetic people-search fixture (`results-regex.html`) with several same-name results where only one is the right match, and the badge / age-label / alias-separator casing varies between cards. Asserts the matched profile resolves to `John Smith`, age `38`, Chicago, which only holds when the `i` flag is honoured.
- `npm run test-unit` covers the `/pattern/` parser (`/x/i`, unsupported flags staying literal, the `/` and `//` edge cases) and the afterText/beforeText/separator behaviour, including the match-and-slice semantics that keep capture groups out of the result.

## Checklist

*Please tick all that apply:*

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Touches shared profile scraping logic used in production broker actions; behavior is backward compatible but misconfigured regex could change extracted PII fields until rules are fixed.
> 
> **Overview**
> Broker extract rules can now use the same **`/pattern/`** and **`/pattern/i`** JSON convention for **`afterText`** and **`beforeText`**, not only **`separator`**, so labels like age and AKA can be matched case-insensitively without code changes.
> 
> **`parseRegexFromString`** replaces the old separator-only parser and accepts an optional **`i`** flag; unsupported flags (e.g. `/x/g`) stay **literal strings**. **`shapeString`** routes **`afterText`** / **`beforeText`** through new **`splitOnce`**, which keeps literal **`String#split`** behavior for plain strings and uses **match + slice** for regex so capture groups do not pollute results and “after” keeps everything past the first match.
> 
> Plain-string configs are unchanged. Coverage adds unit tests for the parser and regex shaping, plus an integration test on a synthetic results page with mixed casing on badges and labels.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9966bdbb99521d4d29d52955aa93e1a9b7d62461. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


daxtheduck (dismissed):

brianhall (approved):
This looks great, thanks!

PR #2803, opened 2026-06-26
Add aliases for captcha providers

brianhall:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1213642859971038?focus=true

## Description
This PR allows aliases to be defined for captcha services. This allows for captchas which are solved the same (e.g. an image captcha) but need to pass a distinct type to the backend.

## Testing Steps

- `npm run test-unit`
- `npm run test-int`

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Medium Risk**
> Changes the captcha type sent to dbp-api for aliased flows; incorrect reporting logic could misroute backend solving, but scope is limited to broker-protection captcha handling with new tests.
> 
> **Overview**
> **Captcha providers can register optional type aliases** so multiple backend captcha types (e.g. `red-circle`, `basic-math`) reuse the same DOM extraction logic as `image` while still calling dbp-api with the requested type.
> 
> `CaptchaFactory.registerProvider` accepts `{ aliases }`, maps each alias to the same provider instance, and **deduplicates providers** in detection paths so alias keys do not run `isSupportedForElement` multiple times. The image provider is wired with aliases `red-circle` and `basic-math`.
> 
> **`getCaptchaInfo` changes the reported `type` in the success payload**: when the action’s `captchaType` resolves directly to a provider (including via alias), the response echoes that requested type; if the code fell back to element detection, it uses the provider’s canonical type.
> 
> Coverage adds factory unit tests for alias registration and deduplication, plus an integration test asserting an aliased image captcha returns `type: 'red-circle'` with a base64 `siteKey`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eb14a92abb590a4cf88e0b48645675cf27c1f37c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/injected/src/features/broker-protection/captcha-services/factory.js
+++ b/injected/src/features/broker-protection/captcha-services/factory.js
@@ -58,6 +58,8 @@ export class CaptchaFactory {
      * @returns {Array<import('./providers/provider.interface').CaptchaProvider>}
      */
     _getAllProviders() {
-        return Array.from(this.providers.values());
+        // Aliases register the same provider instance under multiple keys, so deduplicate to
+        // avoid running a provider's detection methods once per alias on the detection miss path.
+        return Array.from(new Set(this.providers.values()));
     }
 }
--- a/injected/src/features/broker-protection/captcha-services/factory.js
+++ b/injected/src/features/broker-protection/captcha-services/factory.js
@@ -9,9 +9,19 @@ export class CaptchaFactory {
     /**
      * Register a captcha provider
      * @param {import('./providers/provider.interface').CaptchaProvider} provider - The provider to register
+     * @param {Object} [options] - The optional provider options
+     * @param {Array<string>} [options.aliases] - An optional list of aliases for that provider
      */
-    registerProvider(provider) {
+    registerProvider(provider, options) {
         this.providers.set(provider.getType(), provider);
+
+        // Some captchas use the same DOM mechanisms to get the relevant captcha information, but
+        // need to call dbp-api with a distinct captcha type.
+        if (options?.aliases) {
+            options?.aliases.forEach((alias) => {
+                this.providers.set(alias, provider);
+            });
+        }
     }
 
     /**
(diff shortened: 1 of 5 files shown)


daxtheduck (dismissed):

daxtheduck (dismissed):

daxtheduck (dismissed):

injected/src/features/broker-protection/captcha-services/factory.js
```diff
@@ -9,9 +9,19 @@ export class CaptchaFactory {
...
+     * @param {Array<string>} [options.aliases] - An optional list of aliases for that provider
      */
-    registerProvider(provider) {
+    registerProvider(provider, options) {
         this.providers.set(provider.getType(), provider);
+
+        // Some captchas use the same DOM mechanisms to get the relevant captcha information, but
+        // need to call dbp-api with a distinct captcha type.
+        if (options?.aliases) {
+            options?.aliases.forEach((alias) => {
```
madblex, line 21: not sure I follow why all providers get the aliases; aren't aliases bound to a specific provider?

madblex (approved):
LGTM with one question, but approving as I am sure it'll be handled

injected/src/features/broker-protection/captcha-services/factory.js
brianhall: Hoping I'm understanding your question (please do push back if not) - `registerProvider` happens in the context of a specific provider, and this just makes sure that the alias resolves to the provider. So in the case of the `image` provider registration, the following gets added to the `providers` Map:

Providers:
- `image` (actual provider name) -> `ImageProvider()`
- `red-circle` (alias) -> `ImageProvider()`
- `local-llm`(alias) -> `ImageProvider()`

madblex: Sorry that was me misreading this, I see what's happening; we're using 2 different key types for the providers object; should we be worried about clashes on captcha types?

I see why you did this though as this gives you for free the mapping to the correct captcha provider without extra code; let's see how it evolves, thanks!

madblex (approved):

injected/src/features/broker-protection/captcha-services/factory.js
brianhall: Yeah, it's a good shout, and is one of the reason I left the alias definition in the registry vs. in the provider itself - so that all the aliases would be in one place.

daxtheduck (approved):

PR #2907, opened 2026-07-27
Support extracting extras from profiles and addresses

noisysocks:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1216773212741748

**Tech Design:** https://app.asana.com/1/137249556945/project/1209121419454298/task/1216773212741757

## Description

Adds support for extracting `extras` from profiles and addresses.

This solves the immediate need to extract `.street` and `.zip` to fix PeopleFinders. In the medium term it gives us more flexibility when fixing rules that require filling a form with extra info from an extracted profile — capturing a new field becomes a config change rather than a code change.

Extras are omitted when empty, and their absence is tolerated everywhere. Profile hashing strips extras and collapses addresses back to one per city/state, so capturing them never churns an existing broker identifier.

Support for filling out a fake DOB using `$generated_dob$` has moved to #2929, which is independent of this PR.

## Testing Steps

Integration tests (`npm run test-int`) and unit tests (`npm run test-unit`) cover extraction and form filling for both profile- and address-level extras.

To test manually, clone the `apple-browsers` repo and follow the steps in https://github.com/duckduckgo/apple-browsers/pull/5995.

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Medium Risk**
> Changes core extract/aggregate/fillForm and profile identifier hashing in broker protection; behavior shifts for address lists and form filling but is covered by tests and hash stability is explicitly preserved.
> 
> **Overview**
> Broker protection **extract** now captures **extras** on profiles and addresses: full-address parsing adds **`street`** and **`zip`** under address `extras`, and unknown config profile fields fall back to generic text extraction into profile-level `extras`. Address aggregation uses smarter **deduplication** so same city/state with different streets stay separate while redundant city-only rows drop when a richer address exists; href **slug** addresses get hyphen cleanup where appropriate.
> 
> **Profile hash identifiers** intentionally ignore extras and collapse addresses back to one per city/state via `omitExtrasForHash`, so brokers that start recording street/zip do not change existing identifiers.
> 
> **fillForm** picks an address by matching the user's city/state (or the first address), resolves field values from flat profile keys, address city/state, and merged profile plus address extras (e.g. `street`), and still skips optional empty fields like `middleName`. Integration and unit tests cover href/title/text address shapes, profile extras, and opt-out form filling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9c464861b595397c36d9f6d61eac13f0d1286a7c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


injected/src/features/broker-protection/extractors/extra.js
```diff
@@ -0,0 +1,13 @@
+import { firstString, selectStrings } from '../actions/extract.js';
```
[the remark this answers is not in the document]
noisysocks: Good catch — fixed in 412bf1c. Added `injected/src/features/broker-protection/extractors/extra.js` to `CORE_FILES`; `npm run tsc-strict-core` passes with no changes needed to the extractor itself.

---

injected/src/features/broker-protection/actions/extract.js
```diff
@@ -408,6 +430,57 @@ export function aggregateFields(profile) {
...
+/**
+ * Drop exact duplicates, plus any bare city/state pair that a richer version of the same city/state
+ * already covers.
+ *
+ * e.g. [{Dallas, TX}, {Dallas, TX, 1 Main St}, {Dallas, TX, 2 Oak Ave}] -> [{Dallas, TX, 1 Main St}, {Dallas, TX, 2 Oak Ave}]
+ *
+ * @param {{city: string, state: string|null, extras?: ProfileExtras}[]} addresses
+ * @return {{city: string, state: string|null, extras?: ProfileExtras}[]}
+ */
+function dedupeAddresses(addresses) {
```
brianhall: Can we move this and the below functions to [utils](https://github.com/duckduckgo/content-scope-scripts/blob/main/injected/src/features/broker-protection/utils/utils.js) (where `sortAddressesByStateAndCity` is)? Minor, but I think they'd be better there.

injected/src/features/broker-protection/actions/fill-form.js
```diff
@@ -112,33 +120,62 @@ export function fillMany(root, elements, data) {
...
+/**
+ * Pick the address to fill from: the first profile address whose city/state matches any of the
+ * user's addresses, otherwise the first profile address, otherwise `null` when there are none.
+ *
+ * @template {{city: string; state: string | null}} T
+ * @param {T[]} [profileAddresses]
+ * @param {{city: string; state: string | null}[]} [userAddresses]
+ * @return {T | null}
+ */
+export function selectAddress(profileAddresses, userAddresses) {
```
brianhall, line 189: I had to look this up from the Tech Design as it's a pretty big behavioral shift, but that helped refresh my memory. 🙂 

> **C-S-S picks the address.** Because native forwards every address, C-S-S has to choose which one to fill from. We can match the userProfile city / state, fall back to addresses[0], and error if empty.

brianhall: Just to make sure I'm not forgetting an edge case (please correct me if I get this wrong) - this is something new to let us fill addresses in when `dataSource` in `fillForm` is set to `extractedProfile`.

Today we can only do this when `dataSource` in `fillForm` is set to `userProfile` since we do one search per profile location, so there's only one to choose from when filling forms.

Once this change lands, `fillForm` with the `extractedProfile` dataSource will potentially offer a bunch of addresses from what we extracted, so we've got to decide which we want to use, which is the logic here.

injected/src/features/broker-protection/extractors/address.js
```diff
@@ -93,11 +103,43 @@ export function extractAddressFull(select, root, spec) {
...
+            .join(' '),
+    );
+    if (street) extras.street = street;
+    if (addr.zip) extras.zip = addr.zip;
+    return extras;
+}
+
+/**
+ * Not `capitalize` from build-url-transforms.js: that one splits on spaces only, which would leave
+ * the second half of a hyphenated part lowercase ("po-box" -> "Po-box").
```
brianhall: Should we just update build-url-transforms instead? To be honest I haven't thought deeply through potential side effects, just thought I'd ask since this seems really similar.

injected/src/features/broker-protection/extractors/profile-url.js
```diff
@@ -120,7 +120,32 @@ export class ProfileHashTransformer {
...
         return {
             ...profile,
-            identifier: await hashObject(profile),
+            identifier: await hashObject(omitExtrasForHash(profile)),
         };
     }
 }
+
+/**
+ * Rebuilds a profile in its pre-`extras` shape for hashing, so capturing extras never churns
```
brianhall, line 129: Good catch here. 👍

brianhall (commented):
Left a few comments. Overall looking really good.

injected/src/features/broker-protection/actions/extract.js
noisysocks: Agreed — moved in 5c94dfa. `dedupeAddresses` and `hasExtras` are now exported from `utils/utils.js` next to `sortAddressesByStateAndCity`, with `cityStateKey` and `addressDedupKey` kept module-private there since nothing outside needs them.

---

injected/src/features/broker-protection/actions/fill-form.js
noisysocks: That's exactly right — no edge case missed.

`userProfile` is unchanged: native still does one search per profile location, so there's a single address and nothing to choose between. The new `selectAddress` logic only comes into play for `extractedProfile`, where the extracted profile can now carry several addresses and we have to pick one.

The order is: first profile address whose city/state matches any of the user's addresses → else `profileAddresses[0]` → else `null` when there are none. It also only ever supplies `city`/`state` when the flat profile key is absent — `lookupValue` checks the profile's own keys first, so this can't change what an existing `userProfile` fill puts in a field.

---

injected/src/features/broker-protection/extractors/address.js
noisysocks: Fair question — I looked at that, and I'd rather not, because the blast radius is very different.

`capitalize` in `build-url-transforms.js` isn't just an opt-in transform: it's in `baseTransforms`, so it's applied by default to `firstName`, `lastName` and `city` for every broker URL we build, on top of the explicit `${name|capitalize}` syntax. Teaching it to capitalize after hyphens would change generated URLs for every hyphenated name or city — `winston-salem` → `Winston-Salem` instead of `Winston-salem`, `mary-jane` → `Mary-Jane`. Those URLs have to match what the broker actually serves, so that's a live behavioral change across all brokers using those transforms, and it'd want its own testing rather than riding along here.

The local `titleCase` only formats a street we're about to store in `extras`, so getting `Po-Box` right there is self-contained.

Happy to raise it separately if you think the URL-side casing is worth fixing on its own — it may well be, just with its own blast radius.

---

injected/src/features/broker-protection/actions/fill-form.js
noisysocks: Apologies, I was trying out `/autofix-pr` in Claude Code… Not a fan of how it spits long replies out at unsuspecting reviewers! 😅

Yes that’s right.

When `dataSource` is `userProfile`, native calls `fillForm` with a flattened profile containing only one address. When `dataSource` is `extractedProfile`, native calls `fillForm` with what it received from `extractProfile` verbatim including multiple addresses. 

So we need to select _which address?_ in order to fill an address from `dataSource` is `extractedProfile`.

injected/src/features/broker-protection/extractors/address.js
noisysocks: Yeah this can be tightened. I dropped `titleCase` in favour of existing `capitalize` and made it so that we’re only de-slugging when address comes from `attribute: “href”`.

injected/src/features/broker-protection/extractors/address.js
```diff
@@ -78,26 +84,72 @@ export function cityStatePartToCombo({ city, state }) {
...
+ * Clean up addresses that were parsed from a href attribute by removing hyphens from non-number fields.
+ *
+ * @param {Record<string, string>} parsed
+ * @return {Record<string, string>}
+ */
+function deslugParsedAddress(parsed) {
+    return Object.fromEntries(
+        Object.entries(parsed).map(([key, value]) => [
+            key,
+            key === 'number' || key === 'sec_unit_num' || key === 'zip' ? value : value.replace(/-/g, ' '),
```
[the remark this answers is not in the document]
noisysocks: Aye, we have to choose between it not working with cities like Winston-Salem or cities like New York City. Such is life – it’s only a best effort parse.

brianhall (approved):
LGTM, thanks @noisysocks

PR #2922, opened 2026-08-03
NTP: Don't show suggestions for a term that was already cleared

sashalavron:
Asana: https://app.asana.com/1/137249556945/project/1213083312441631/task/1216559242448599?focus=true

## Summary
Fix suggestions flashing back into view after the user has already cleared or dismissed the omnibar input.

## Changes
- Omnibar suggestions fetches can now be cancelled; clearing the input or pressing ESC orphans any in-flight fetch so its late response can no longer re-open the suggestions list for a term that's gone.

## Steps to test
1. Open `https://rawcdn.githack.com/duckduckgo/content-scope-scripts/05733fb80fc5a8f18bdea091f87f87f2e197bc5f/build/integration/pages/new-tab/index.html?omnibar=true&omnibar.suggestionsDelay=100`
2. Type `pizza` in the omnibar, then immediately clear the input
3. With this fix, they stay hidden. (Originally suggestions pop back open for the cleared term)

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Low Risk**
> Localized omnibar suggestion state handling and test/mock helpers; no auth, security, or data-path changes.
> 
> **Overview**
> Fixes **suggestions flashing back** after the user clears the omnibar or dismisses the list while a fetch is still in flight.
> 
> `useSuggestions` now tracks a **`pendingTerm`** for each in-flight request. **`setSuggestions`** only updates the list when the response term still matches that pending term; **`hideSuggestions`** (empty input or ESC) clears `pendingTerm`, so late responses are ignored. Fetches go through **`requestSuggestions`**, which sets the pending term before calling native.
> 
> Adds Playwright coverage with **`omnibar.suggestionsDelay`** on the mock transport and documents that URL param in the new-tab readme.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 203e3d566077254696413439a86d5e5883cc352a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sashalavron:
bugbot run

special-pages/pages/new-tab/app/omnibar/unit-tests/omnibarService.spec.js
```diff
@@ -38,6 +38,65 @@ const samplePageContext = {
...
+
+        service.getSuggestions('pizza');
+        service.cancelSuggestions();
+        const pending = service.getSuggestions('pizza dough');
+        resolveRequest(sampleSuggestions);
+        await pending;
+
+        deepEqual(received, ['pizza dough']);
+    });
+});
```
shakyShane: I'm not convinced we should continue this pattern. I can't see anything in here that can't be replicated inside the playwright tests. The fact that you're typing the dependencies as "any" gives an indication that we're trying to do too much in these tests. I would just drop this and consider not adding other ones like this when we have a full setup for testing time and API requests and everything like that already within Playwright

special-pages/pages/new-tab/app/omnibar/omnibar.suggestions.service.js
```diff
@@ -32,6 +32,14 @@ export class OmnibarSuggestionsService {
...
     }
 
+    /**
+     * Bumping the id orphans the in-flight fetch, so a response landing after the input
+     * was cleared or dismissed can't re-show suggestions for a term that's no longer there.
+     */
+    cancelFetch() {
+        this.#lastFetchId++;
+    }
+
```
shakyShane: I don't think any of this should involve the service and this obviously isn't really cancelling anything so the mismatched naming kind of proves the point. Your change is a UI concern so I'd just handle it in the reducer instead. Change your state there and don't worry about the service, don't make any changes to this.

This deliberately simple implementation of a staleness guard has served us well here for ages and only when it's an actual service concern should we change this here.

shakyShane (changes requested):
@sashalavron thanks for the PR, but i think the fix could be better suited to a UI concern (eg: handle it in the reducer with updated local state). that way you don’t need the unit tests and nothing in the boundary/service layer needs to change.

special-pages/pages/new-tab/app/omnibar/unit-tests/omnibarService.spec.js
sashalavron: I agree 1000%

special-pages/pages/new-tab/app/omnibar/omnibar.suggestions.service.js
sashalavron: Resolved as agreed, the logic was movied to the reducer

daxtheduck (dismissed):

sashalavron:
Hey @shakyShane , could you check the PR again? It’s been reworked, and all the previous issues should be resolved.

daxtheduck (approved):

shakyShane (approved):

PR #2924, opened 2026-08-04
Onboarding: remove the v3 flow and make v4 the default

noisysocks:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1217018051146861

## Description

While fixing #2925, I noticed `@formkit/auto-animate` is only initialised by a component in the old v3 onboarding flow, which neither Windows nor macOS request anymore ([duckduckgo/apple-browsers#6155](https://github.com/duckduckgo/apple-browsers/pull/6155) removes the feature flag that could otherwise have made Apple fall back to v3). v4 has been the deployed flow for a while now, so it's safe to remove v3 entirely rather than just its `auto-animate` usage.

## Testing Steps

- Open https://rawcdn.githack.com/duckduckgo/content-scope-scripts/b3e46050e5ca693e7ef2c1f205642ef343de700a/build/integration/pages/onboarding/?platform=macos.
- Step through the whole flow and confirm it's the v4 (bubbles) flow throughout, with settings-row icons rendering on the `systemSettings` and `customize` steps.
- Repeat with `?platform=windows` on the same URL.

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Medium Risk**
> Large removal of a former production onboarding path; risk is mitigated because v4 was already the deployed flow, but any native client still depending on v3 or init `order` would silently get v4 only.
> 
> **Overview**
> Removes the legacy **v3** onboarding UI and routing so the special page always runs the **v4** (bubble) experience. The app entry no longer switches between `v3`/`v4` apps, drops the components playground, and **Settings** defaults to v4 step definitions without `order` / `withNamedOrder` overrides from init or URL.
> 
> **Deleted scope:** the full `v3/` tree plus v3-only shared pieces (list/progress/Rive/timeout components and related icons). Init messaging docs no longer mention an `order` field. v4 row icon types move to `RowIcon` in v4 `data-types` instead of the removed `ListItem` export.
> 
> **Tests:** v3 integration tests and screenshot suite are removed; v4 screenshots no longer pass `order: 'v4'` in init mocks.
> 
> **CI:** `semver-label` caps `BUILD_DIFF`, `SOURCE_DIFF`, and `PR_FILES` at 120KB before export so very large PRs avoid Linux `E2BIG` when spawning `node` (aligned with existing 80k truncation inside `semver-analysis.mjs`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3a3604b07c3f4d69465684cc7a68878d7e1b4ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


noisysocks:
Heads up on the second commit (04ff65d) — it touches shared CI, not onboarding, so it deserves an explicit call-out.

`semver_label` failed on the first push with `Argument list too long` (exit 126). The cause is in `.github/workflows/semver-label.yml`:

```sh
export SOURCE_DIFF="$(cat /tmp/source_diff.txt 2>/dev/null || echo '')"
RESULT=$(node base/.github/scripts/semver-analysis.mjs)
```

Linux caps any single env var at `MAX_ARG_STRLEN` (128KiB). This PR's source diff is ~330KB, so `execve` fails with `E2BIG` before node ever starts. Reproduced locally:

```
full diff bytes  : 329722
full             : OSError [Errno 7] Argument list too long: 'node'
head -c 120000   : exit=0, node sees 119998 chars
```

This is a pre-existing limitation rather than anything this PR introduced — any PR with a >128KB text diff hits it. Note it can't be fixed by editing `semver-analysis.mjs` from a PR branch either, since the workflow deliberately runs the copy from the `base` checkout.

The fix caps the diffs with `head -c 120000` before exporting. That's deliberately chosen to be a no-op for the analysis: `semver-analysis.mjs` already calls `truncateDiff(sourceDiff, 80_000)`, so anything past 80,000 chars was being discarded regardless. 120,000 bytes sits above that cap and below the 128KiB exec limit, so the model receives byte-identical input to what it would have gotten had the exec succeeded. `BUILD_DIFF` got the same treatment since it has its own 80,000-char cap and this PR deletes a lot of built artifacts.

Happy to pull this out into its own PR against `main` if you'd rather keep this one purely about onboarding — the tradeoff is that `semver_label` stays red here until that lands.

Everything else on the first push was green: all 8 `injected` integration matrix jobs, all unit jobs on both ubuntu and windows, `snapshots`, `build`, and `production-deps`. The `special-pages/chromium` and `special-pages/webkit` integration jobs were still running when the new push superseded them; I'll follow up on their results from this commit.

`Authorized Review` is pending team approval, and the PR is still a draft — both need a human.

---

shakyShane (approved):

PR #2929, opened 2026-08-05
Support filling a generated date of birth

noisysocks:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1216773212741748

**Tech Design:** https://app.asana.com/1/137249556945/project/1209121419454298/task/1216773212741757

## Description

Adds support for filling out a fake date of birth using `$generated_dob$`.

Some opt-out forms ask for a date of birth we don't hold — we only have the age extracted from the profile. This fills a random date that makes the profile exactly that age as of today, so the form can be submitted without the config having to invent a date.

Split out of #2907, which no longer contains this change. The two are independent.

## Testing Steps

Integration tests (`npm run test-int`) and unit tests (`npm run test-unit`) cover both the generated date of birth and the error when there's no age to generate one from.

To test manually, clone the `apple-browsers` repo and follow the steps in https://github.com/duckduckgo/apple-browsers/pull/5995.

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Low Risk**
> Extends an existing fill-form generator pattern with validation and tests; no auth or persistence changes, and behavior is config-driven for broker opt-out flows.
> 
> **Overview**
> Adds **`$generated_dob$`** to broker-protection **`fillForm`**, alongside existing generated field types. When a form asks for a date of birth but the profile only has **age**, configs can use this type to fill a random **`YYYY-MM-DD`** value that yields that exact age as of today.
> 
> **`generateDateOfBirth`** picks a day within the birth year (with leap-day handling) and **`fillMany`** validates that **`age`** is present and a non-negative number; otherwise the action fails with a clear error like other generated fields.
> 
> Coverage includes unit tests over many random draws, integration success/error scenarios, a **`user_dob`** date input on the test form, and a page-object helper to assert age from the filled value.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ae2a1d0483eacd6570f66c589ce9dbe3d81de16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/injected/src/features/broker-protection/actions/fill-form.js
+++ b/injected/src/features/broker-protection/actions/fill-form.js
@@ -57,11 +57,18 @@ export function fillMany(root, elements, data) {
         } else if (element.type === '$generated_zip_code$') {
             results.push(setValueForInput(inputElem, generateZipCode()));
         } else if (element.type === '$generated_dob$') {
+            if (!Object.prototype.hasOwnProperty.call(data, 'age')) {
+                results.push({
+                    result: false,
+                    error: `element found with selector '${element.selector}', but data didn't contain an 'age' to generate a date of birth from`,
+                });
+                continue;
+            }
             const age = parseInt(data.age, 10);
             if (!Number.isFinite(age) || age < 0) {
                 results.push({
                     result: false,
-                    error: `element found with selector '${element.selector}', but data didn't contain an 'age' to generate a date of birth from`,
+                    error: `element found with selector '${element.selector}', but data contained an 'age' that wasn't a positive number`,
                 });
                 continue;
             }
--- a/injected/src/features/broker-protection/actions/fill-form.js
+++ b/injected/src/features/broker-protection/actions/fill-form.js
@@ -1,6 +1,6 @@
 import { getElement, generateRandomInt } from '../utils/utils.js';
 import { ErrorResponse, SuccessResponse } from '../types.js';
-import { generatePhoneNumber, generateZipCode, generateStreetAddress } from './generators.js';
+import { generatePhoneNumber, generateZipCode, generateStreetAddress, generateDateOfBirth } from './generators.js';
 import { states } from '../comparisons/constants.js';
 
 /**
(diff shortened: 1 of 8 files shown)


daxtheduck (dismissed):

injected/src/features/broker-protection/actions/fill-form.js
```diff
@@ -56,6 +56,16 @@ export function fillMany(root, elements, data) {
             results.push(setValueForInput(inputElem, generatePhoneNumber()));
         } else if (element.type === '$generated_zip_code$') {
             results.push(setValueForInput(inputElem, generateZipCode()));
+        } else if (element.type === '$generated_dob$') {
+            const age = parseInt(data.age, 10);
```
brianhall, line 67: Just noticing that we're checking data presence with `Object.prototype.hasOwnProperty.call(data, 'city')` in [other places](https://github.com/duckduckgo/content-scope-scripts/pull/2929/changes#diff-6bdb382047219e3cfabdff4afc15999eb9562e19346ce2bb979a3db42ec06e2eR94) before attempting to use it. I don't actually think an age can be blank, but perhaps it's worth the guard.

brianhall (approved):
Left one minor comment. Overall I think this makes sense. One thing I do think worth considering is that I imagine some sites will choose to go with separate dropdowns for Y/M/D instead of a date input, which wouldn't work for this.

To handle that case, we could optionally pass in three named selectors for the different date parts. That's a bigger change, so I'm ok if we want to skip and merge as-is, but wanted to get your thoughts.

injected/src/features/broker-protection/actions/fill-form.js
noisysocks: Added in 67fb88b. The `parseInt` check already caught a missing `age` (it comes back `NaN`), but adding the explicit guard means the two cases get distinct messages: absent key vs. present-but-not-a-positive-number. The old message claimed "didn't contain an 'age'" even when the data had an `age` of `"abc"`, so this reads better in a failure report too.

---

daxtheduck (dismissed):

daxtheduck (approved):

noisysocks:
> To handle that case, we could optionally pass in three named selectors for the different date parts. That's a bigger change, so I'm ok if we want to skip and merge as-is, but wanted to get your thoughts.

Let’s do this as a follow-up! 👍

PR #2930, opened 2026-08-05
Add initial xpath support

GuiltyDolphin:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/492600419927320/task/1217111223619806?focus=true

## Description

Adds support for targeting text detection on xpath selectors.

## Testing Steps

Unit tests.

For manual test, load updated extension. In the service worker console paste:

```
await globalThis.components.remoteConfig.modify((config) => {
    const settings = config.features.webDetection.settings;

    settings.detectors.adwalls.generic_en_xpath = {
        match: {
            text: {
                xpath: "//body//text()[not(ancestor::script) and not(ancestor::style) and not(ancestor::noscript)]",
                pattern: [
                    'ad ?block(er)? detected',
                    'adjust(ing)?( your)? ad blocking settings',
                    '(turn(ing)? off|disable|disabling|deactivate)( your| my)? ad ?block(er)?',
                    'your ad ?blocker is on',
                    "you('re| are| may be) using an ad blocker",
                ],
            },
        },
    };

    const before = settings.conditionalChanges.length;
    settings.conditionalChanges = settings.conditionalChanges.filter(
        (change) => ![].concat(change.condition?.domain ?? []).includes('homedepot.com'),
    );
    console.log(`dropped ${before - settings.conditionalChanges.length} homedepot.com conditional change(s)`);

    config.version = Date.now();
    return config;
});
```

Visit a site with the phrase "adblocker detected". Make a breakage report and view request. You should see both `adwalls.generic_en` and `adwalls.generic_en_xpath` in the `webDetection` results.

Now visit homedepot.com and repeat. You should only see `adwalls.generic_en` in the results (xpath variant prevents being detected inside script).

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Medium Risk**
> Changes live-page text matching for configured detectors (false positive/negative and per-tick CPU on large DOMs), though behavior is gated by config and existing selector-based detectors are unchanged unless xpath is added.
> 
> **Overview**
> **Web-detection text conditions** can now target text via **`xpath`** (in addition to CSS **`selector`**), so detectors can match **rendered** page text while ignoring strings that only live in `<script>`, `<style>`, `<template>`, or `<noscript>`. **`selector` and `xpath` are OR’d**; implicit **`body`** scanning applies only when neither source is set.
> 
> XPath evaluation uses **compiled expressions**, **ordered node snapshots**, and **chunked scanning** with optional **`xpathConfig`** (`chunkSize` / `chunkTail`) so large pages are not fully concatenated on each poll. Tail retention walks back to word boundaries so `\b`-style patterns are less likely to false-match at chunk cuts (with documented limits when matches span boundaries longer than the tail).
> 
> The **`@duckduckgo/privacy-configuration`** dependency is bumped to a revision that includes the schema for these fields. **Unit and Playwright integration tests** cover script-only vs rendered text, clean pages, and chunking edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 019a04f3193300900f22095f7e590eb5b6e8f812. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/injected/src/features/web-detection/matching.js
+++ b/injected/src/features/web-detection/matching.js
@@ -145,6 +145,43 @@ function hasContent(element) {
 const ORDERED_NODE_SNAPSHOT_TYPE = 7;
 
 /**
+ * Compiled XPath expressions, keyed by document and then by expression source.
+ *
+ * `document.evaluate()` re-parses the expression string on every call, and
+ * detectors re-evaluate their conditions on every poll tick against a fixed,
+ * config-supplied set of expressions. Compiling once removes that repeated parse.
+ *
+ * Keyed by document because an `XPathExpression` belongs to the document that
+ * created it.
+ *
+ * @type {WeakMap<Document, Map<string, XPathExpression>>}
+ */
+const compiledXPaths = new WeakMap();
+
+/**
+ * Compile an XPath expression, reusing a previously compiled one where possible.
+ *
+ * An invalid expression throws `SyntaxError` here rather than at evaluation time;
+ * either way it propagates to the caller and surfaces as `detected: 'error'`.
+ *
+ * @param {string} expression
+ * @returns {XPathExpression}
+ */
+function compileXPath(expression) {
+    let cache = compiledXPaths.get(document);
+    if (!cache) {
+        cache = new Map();
+        compiledXPaths.set(document, cache);
+    }
+    let compiled = cache.get(expression);
+    if (!compiled) {
+        compiled = document.createExpression(expression, null);
+        cache.set(expression, compiled);
+    }
+    return compiled;
+}
+
+/**
  * Concatenate the text of every node selected by an XPath expression.
  *
  * Nodes are joined without a separator so the result is equivalent to
@@ -160,7 +197,7 @@ const ORDERED_NODE_SNAPSHOT_TYPE = 7;
  * @returns {string}
  */
 function xpathText(expression) {
-    const snapshot = document.evaluate(expression, document, null, ORDERED_NODE_SNAPSHOT_TYPE, null);
+    const snapshot = compileXPath(expression).evaluate(document, ORDERED_NODE_SNAPSHOT_TYPE, null);
     let text = '';
     for (let i = 0; i < snapshot.snapshotLength; i++) {
         text += snapshot.snapshotItem(i)?.textContent || '';
@@ -198,13 +235,10 @@ function evaluateSingleTextCondition(condition) {
 
     const patternComb = new RegExp(patterns.join('|'), 'i');
 
-    // Disjunction: any expression whose selected text matches is success
-    if (xpaths.some((expression) => patternComb.test(xpathText(expression)))) {
-        return true;
-    }
-
-    // Disjunction: any selector having a matching element is success
-    return selectors.some((selector) => {
+    // Disjunction: any selector having a matching element is success.
+    // Checked before xpath because CSS matching avoids the per-call expression
+    // parse and snapshot allocation that `document.evaluate` requires.
+    const selectorMatch = selectors.some((selector) => {
         const elements = document.querySelectorAll(selector);
         for (const element of elements) {
             if (patternComb.test(element.textContent || '')) {
@@ -213,6 +247,12 @@ function evaluateSingleTextCondition(condition) {
         }
         return false;
     });
+    if (selectorMatch) {
+        return true;
+    }
+
+    // Disjunction: any expression whose selected text matches is success
+    return xpaths.some((expression) => patternComb.test(xpathText(expression)));
 }
 
 /**
--- a/injected/src/features/web-detection/matching.js
+++ b/injected/src/features/web-detection/matching.js
@@ -181,28 +181,96 @@ function compileXPath(expression) {
     return compiled;
 }
 
+/** Characters accumulated between pattern tests when scanning XPath text. */
+const DEFAULT_CHUNK_SIZE = 8192;
+
+/** `chunkSize` divisor giving the default `chunkTail`. */
+const CHUNK_TAIL_RATIO = 16;
+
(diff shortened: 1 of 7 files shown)


injected/src/features/web-detection/matching.js
```diff
@@ -132,6 +132,42 @@ function hasContent(element) {
...
+ * selector passed to `querySelectorAll`.
+ *
+ * @param {string} expression
+ * @returns {string}
+ */
+function xpathText(expression) {
+    const snapshot = document.evaluate(expression, document, null, ORDERED_NODE_SNAPSHOT_TYPE, null);
+    let text = '';
+    for (let i = 0; i < snapshot.snapshotLength; i++) {
+        text += snapshot.snapshotItem(i)?.textContent || '';
```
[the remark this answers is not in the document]
jonathanKingston: Hah was just typing this up! See my other comment....

injected/src/features/web-detection/matching.js
```diff
@@ -163,6 +247,12 @@ function evaluateSingleTextCondition(condition) {
         }
         return false;
     });
+    if (selectorMatch) {
+        return true;
+    }
+
+    // Disjunction: any expression whose selected text matches is success
+    return xpaths.some((expression) => patternComb.test(xpathText(expression)));
```
jonathanKingston: This seems overly performance expensive.

```suggestion
    return xpaths.some((expression) => xpathText(patternComb, expression));
```

Where xpathText:

```javascript
function xpathTest(pattern, expression) {
    const snapshot = compileXPath(expression).evaluate(document, ORDERED_NODE_SNAPSHOT_TYPE, null);
    let text = '';
    return [...snapshot].some((match) => pattern.test(match.textContent))
}
```

Hand wrote, so likely full of errors but you get the drift, essentially we'd be short circuiting rather than concatenating large blocks of memory to then text search it.

jonathanKingston (changes requested):

injected/src/features/web-detection/matching.js
GuiltyDolphin: With the per-node matching we miss text that spans multiple text nodes, so we end up with false negatives. `<p>Disable your <b>adblocker</b></p>` would no longer match.

There is an alternative which is to chunk the matching which caps the peak memory consumption. There's still non-zero risk of false negatives but much lower than per-node, and we maintain a consistent memory size regardless of page. I'll look into this further.

jonathanKingston: yeah peak memory was the problem imo. However, don't we get all the ancestor tree with textContent?

GuiltyDolphin: If you have:

`<p>Some text!<b>BOLD</b></p>`

With xpath `//body//text()[not(ancestor::script) and not(ancestor::style) and not(ancestor::template) and not(ancestor::noscript)]` then the output is an iterator of nodes where one has `Some text!` as `.textContent` and one has `BOLD`.

You can traverse the tree like going to the parent node and then getting `textContent`, but then you'd end up including any script text again. xpath 2.0 has more options for getting the text from the tree but not widely supported.

GuiltyDolphin:
@cursoragent rebase & update to latest remote config for `injected/`

injected/src/features/web-detection/matching.js
```diff
@@ -132,6 +132,182 @@ function hasContent(element) {
...
+ * language of the text.
+ *
+ * The equivalence holds for every regex flag except `i` combined with `u`/`v`, where
+ * case folding pulls U+017F and U+212A into `\w`. Using character codes sidesteps that,
+ * along with the `g`/`y` statefulness that made the previous `RegExp.test` form fragile.
+ *
+ * @param {number} code
+ * @returns {boolean}
+ */
+function isWordCode(code) {
```
jonathanKingston, line 219: Why not use a precomputed `^\w$` then (if the aim is parity with \w).

Not worth blocking on though.

injected/src/features/web-detection/matching.js
```diff
@@ -132,6 +132,182 @@ function hasContent(element) {
...
+ */
+function retainTail(buffer, chunkTail) {
+    let cut = buffer.length - chunkTail;
+    if (cut <= 0) return buffer;
+    // Ceiling on the walk, so an unbroken run of word characters cannot grow the buffer
+    // without limit. Reaching it leaves position 0 mid-word, giving up the guarantee below
+    // for this flush - a hard bound is worth more than exact `\b` semantics inside a blob
+    // that a phrase pattern will not match anyway. Never more than `chunkTail`, so a tail
+    // of 0 still retains nothing.
+    const limit = Math.max(0, cut - Math.min(chunkTail, MAX_WORD_LENGTH));
```
jonathanKingston, line 260: This isn't config driven, does it matter?

jonathanKingston (approved):
Added comments, read them before merging :)

injected/src/features/web-detection/matching.js
GuiltyDolphin: This is faster and equiv and scales better with longer tail sizes (e.g., at 512 tail this has a 51x advantage over the regex; if tail is 2048 this grows to a 203x advantage)

injected/src/features/web-detection/matching.js
GuiltyDolphin: It is possible to hit false positives depending on what we're attempting to detect (e.g., if we are interested in detecting longer hex strings or b64 encodings).

For current use cases, the 64 length should be fine as it covers most Eng words & there's an escape hatch where you can disable the buffering entirely if needed. I'll open a followup to think more about other ways to reduce false positives, but for v1 this should be fine as we can gate the config on platform version.

PR #2953, opened 2026-08-12
Add iOS text selection

aataraxiaa:
**Asana Task/Github Issue:** N/A

## Description

Adds iOS-only text-selection handling to the Apple isolated Content Scope Scripts bundle.

The feature asks native whether it is enabled before installing listeners, retains selected text inside the isolated world, and reports only selection presence plus an event timestamp. Native explicitly reads the retained snapshot through `window.__ddgSelectionFrame.readSelection()`.

Selection changes run in every frame. Focus gating prevents unfocused subframes from claiming, a deferred retry handles focus becoming observable after `selectionchange`, and accepted non-empty changes always reassert native ownership. Native ignores events older than its current state, while equal timestamps use arrival order.

Password-field events clear the shared native selection state even when that frame never claimed it. The implementation also preserves delayed clearing with the original event time, pagehide release, and bfcache restoration. Temporary device diagnostics are not included.

## Testing Steps

- Run `npm exec --workspace=injected jasmine -- --config=unit-test/config.json --filter=TextSelection`.
- Run `npm run build`.
- Run `npm run lint`.
- On iOS, alternate main-frame and iframe selections and invoke Ask Duck.ai on the first attempt after each switch; the exact new selection should attach.
- Select ordinary text in one frame, then focus a password field in another frame; the Duck.ai actions should not reuse the earlier selection.

Validated locally with Node 22.23.2:

- `npm run lint-fix`
- `npm run build`
- Focused TextSelection tests: 17 specs, 0 failures
- `npm run lint`, including strict TypeScript
- Device validation of alternating main-frame and iframe selections

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Medium Risk**
> Touches cross-frame selection and native messaging for Duck.ai-style flows; incorrect ownership or stale snapshots could attach wrong context, though text stays in the isolated world and outbound messages omit selected strings.
> 
> **Overview**
> Introduces the **`textSelection`** content feature on Apple-related injected bundles (base features plus **`apple-isolated`**), with wiring in the feature map, utils allowlist, and strict-core checks.
> 
> The feature listens for **`selectionchange`**, **`pagehide`**, and **`pageshow`**, keeps a **snapshot of selected text** in the isolated world, and notifies native via **`selectionFrameChanged`** with only **`hasSelection`** and **`eventTimestamp`**—not the string itself. Native reads text through a frozen **`window.__ddgSelectionFrame.readSelection()`** hook.
> 
> **Frame ownership** is gated so unfocused iframes do not claim selection; a **deferred retry** handles focus arriving after **`selectionchange`**. Non-empty accepted selections **reassert** ownership on repeat events. Clears are **debounced** (~100ms) while preserving the original event time; **`pagehide`** releases the snapshot and **`pageshow`** (bfcache) re-publishes state.
> 
> Adds the **`selectionFrameChanged`** JSON schema, generated **`text-selection`** message types, and **Jasmine unit tests** for reader install, messaging shape, timestamps, iframe claim/cancel paths, and lifecycle edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3a0fa97e5fe0aa755d5d067eef984f7a65a09c5b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


aataraxiaa:
Note: still in draft while related [tech design](https://app.asana.com/1/137249556945/project/481882893211075/task/1217363923358702) is under review

injected/src/features/text-selection.js
```diff
@@ -0,0 +1,147 @@
...
+        this._post(true, true);
+        return true;
+    }
+
+    _canClaim() {
+        return !this._isFramed() || document.hasFocus();
+    }
+
+    _isFramed() {
+        return window.location && 'ancestorOrigins' in window.location ? window.location.ancestorOrigins.length > 0 : window.top !== window;
```
jonathanKingston: We have a helper for this already for sure.

jonathanKingston (changes requested):

injected/src/messages/text-selection/isEnabled.request.json
```diff
@@ -0,0 +1,15 @@
+{
+  "description": "Checks whether native text-selection tracking is enabled.",
```
jonathanKingston: I'd consider removing this and just firing it always. We no longer trigger errors when the handler doesn't exist.

aataraxiaa: Without this C-S-S installs its selection listeners and isolated-world reader even when the native flag is off. So think it’s worth keeping. But not a strong opinion. If you strongly object, let me know

aataraxiaa:
@jonathanKingston opened for review and requested from you.

injected/src/features/text-selection.js
```diff
@@ -0,0 +1,147 @@
...
+     */
+    _publish(text, canClaim = this._canClaim()) {
+        if (text.trim().length === 0) {
+            this._snapshot = '';
+            this._post(false);
+            return true;
+        }
+        if (!canClaim) return false;
+        this._snapshot = text;
+        this._post(true, true);
```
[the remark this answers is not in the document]
aataraxiaa: This is intentional. Each frame’s _lastHasSelection is local, while native ownership is shared.

injected/src/features/text-selection.js
```diff
@@ -0,0 +1,183 @@
...
+    /** @param {Event} event */
+    _secureFieldChanged(event) {
+        if (!this._isPasswordField(event.target)) return;
+        this._releaseImmediately(this._eventTimestamp(event));
+    }
+
+    /** @param {EventTarget | null} element */
+    _isPasswordField(element) {
+        return element instanceof window.HTMLInputElement && element.type === 'password';
+    }
```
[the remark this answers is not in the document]
jonathanKingston: I think this is adding a bunch of complexity and seems like it might not be working fully? Do we need it?

injected/src/features/text-selection.js
```diff
@@ -0,0 +1,183 @@
...
+    _clearTimer = null;
+    /** @type {number | null} */
+    _claimTimer = null;
+    /** @type {string} */
+    _snapshot = '';
+    /** @type {number} */
+    _snapshotTimestamp = 0;
+
+    async init() {
+        if (this.platform?.name !== 'ios') return;
```
jonathanKingston: I'd not gate here, let's just use the config for that. Then we could enable on macos without any js changes :)

injected/integration-test/page-objects/duckplayer-overlays.js
```diff
@@ -219,7 +219,10 @@ export class DuckplayerOverlays {
 
     async userValuesCallIsProxied() {
         const calls = await this.collector.outgoingMessages();
-        const message = calls[0];
+        const message = calls.find(
```
jonathanKingston: How come this file has changed? It seems unrelated?

injected/src/messages/text-selection/isEnabled.request.json
aataraxiaa: Keeping this as the separate native runtime gate. Privacy Configuration controls whether the feature is available on a platform. `textSelection.isEnabled` reflects `aiChatTextActions` and prevents installing the reader and listeners while that flag is off or the handler is absent.

injected/src/features/text-selection.js
aataraxiaa: Done. The hard-coded iOS guard is removed. Platform availability will be controlled through Privacy Configuration, so macOS can be enabled without a JavaScript change.

injected/src/features/text-selection.js
```diff
@@ -0,0 +1,155 @@
...
+        if (text.trim().length === 0) {
+            this._snapshot = '';
+            this._snapshotTimestamp = eventTimestamp;
+            this._post(false, false, eventTimestamp);
+            return true;
+        }
+        if (!canClaim) return false;
+        this._snapshot = text;
+        this._snapshotTimestamp = eventTimestamp;
+        this._post(true, true, eventTimestamp);
```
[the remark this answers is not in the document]
aataraxiaa: Intentional. A frame may need to reassert after another frame becomes current even when its selected text is unchanged, so snapshot-text dedup would restore the alternating-frame bug. `e0548dd6` now tests two identical snapshots producing two notifications. We can revisit coalescing if profiling shows bridge pressure.

injected/src/features/text-selection.js
```diff
@@ -0,0 +1,155 @@
...
+
+        document.addEventListener('selectionchange', (event) => this._selectionChanged(event), true);
+        window.addEventListener('pagehide', (event) => this._pageHidden(event), true);
+        window.addEventListener('pageshow', (event) => this._pageShown(event), true);
+        this._publish(this._selectionText());
+    }
+
+    _selectionText() {
+        const selection = window.getSelection();
+        return selection ? String(selection) : '';
```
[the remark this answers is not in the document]
aataraxiaa: Removal was intentional following reviewer feedback because the DOM heuristic was incomplete, especially for Shadow DOM. This feature only reads `window.getSelection()` and never input values. Secure-field behavior remains a WebKit/native boundary explicitly raised in the security review. We are not adding partial credential-field heuristics here.

injected/integration-test/page-objects/duckplayer-overlays.js
aataraxiaa: Removing the hard-coded iOS guard means the Apple isolated macOS bundle now sends `textSelection.isEnabled` at startup. That exposed the assumption that the Duck Player request was always `calls[0]`. This test-only change finds the Duck Player `getUserValues` request explicitly. It does not change Duck Player behavior.

injected/src/features/text-selection.js
```diff
@@ -0,0 +1,155 @@
...
+    /** @type {number | null} */
+    _claimTimer = null;
+    /** @type {string} */
+    _snapshot = '';
+    /** @type {number} */
+    _snapshotTimestamp = 0;
+
+    async init() {
+        try {
+            const { enabled } = await this.request('isEnabled', {});
```
jonathanKingston: Why do we need this? The feature has a config of it's own? We're adding an unnecessary message to the client for a minor feature.

If we want user control we should pass it through the preferences instead it's statically sent through to the script. I would skip this though.

injected/integration-test/page-objects/duckplayer-overlays.js
jonathanKingston: We should skip sending that message altogether.

injected/src/messages/text-selection/isEnabled.request.json
jonathanKingston: I missed this response, you're adding a blocking round trip message to disable some code that we can already disable statically with the config. If we need the config to disable both toggles that's where we should do that personally.

jonathanKingston (changes requested):
We should skip that init message it's a round trip to the client which we don't need.

injected/src/features/text-selection.js
aataraxiaa: Thx, done, removed the `textSelection.isEnabled` round trip and the corresponding native handler

injected/integration-test/page-objects/duckplayer-overlays.js
aataraxiaa: Thx, the initialization message has been removed, so I’ve also reverted this unrelated Duck Player test change

aataraxiaa:
Thx @jonathanKingston, PR updated based on suggestions

jonathanKingston (approved):

PR #2977, opened 2026-08-20
Element hiding: Improve performance by taking advantage of browser selector indexing

dharb:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/task/1217648927550635?focus=true

## Description
Despite most of our element hiding rules being simple indexable selectors, we're actually not getting any benefit from browser engine indexing because we're wrapping them in an `:is()` pseudo-class to make them [forgiving selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:is#forgiving_selector_parsing). We did this in 2022 to address browser engine compatibility differences—Firefox didn't support the :has() pseudo-class at the time, and we needed it in element hiding to address Sign in with Google pop-ups. By making the selectors forgiving, it made it so the code wouldn't throw an error when we attempted to query the page for unsupported selector types.

This PR adds logic to the element hiding algo so we only wrap selector lists in `:is()`, and we catch errors thrown if malformed selector rules are added to remote config.

## Testing Steps
- Pull branch
- Run integration tests

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Medium Risk**
> Changes how ad-hiding selectors are queried on every page pass. A bug could miss hides or skip remaining rules, but behavior is covered by new containment tests.
> 
> **Overview**
> Stops wrapping every element-hiding `querySelectorAll` selector in `:is()`, so simple indexable selectors can use the browser’s selector index. `:is()` is now applied only when the selector contains a comma (a list).
> 
> Invalid or unsupported selectors are caught per rule so one bad remote-config rule no longer aborts the whole hide/unhide pass. Integration tests cover malformed selectors, unknown pseudo-classes, and mixed valid/invalid lists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e0e14dca0de3789bebd08480f81a209b030d4ef3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/injected/src/features/element-hiding.js
+++ b/injected/src/features/element-hiding.js
@@ -374,10 +374,10 @@ function forgivingSelector(selector) {
 
 /**
  * Resolve a rule's selector for use with querySelectorAll
- * 
+ *
  * :is() makes a selector list forgiving, but has a perf cost since
  * it's a pseudo-class that isn't indexed. Only wrap when the selector
- * might be a list, in which case one malformed selector can lead to 
+ * might be a list, in which case one malformed selector can lead to
  * the rest of the list not applying
  */
 function querySelectorFor(selector) {
--- a/injected/src/features/element-hiding.js
+++ b/injected/src/features/element-hiding.js
@@ -62,6 +62,7 @@ const parser = new DOMParser();
 let hiddenElements = new WeakMap();
 let modifiedElements = new WeakMap();
 let appliedRules = new Set();
+const resolvedSelectors = new Map();
 let shouldInjectStyleTag = false;
 let styleTagInjected = false;
 let mediaAndFormSelectors = 'video,canvas,embed,object,audio,map,form,input,textarea,select,option,button';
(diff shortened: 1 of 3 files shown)


injected/src/features/element-hiding.js
```diff
@@ -331,10 +332,15 @@ function hideAdNodes(rules) {
     const document = globalThis.document;
 
     rules.filter(hasSelector).forEach((rule) => {
-        const selector = forgivingSelector(rule.selector);
-        const matchingElementArray = [...document.querySelectorAll(selector)];
+        let matchingElementArray;
+        try {
+            matchingElementArray = [...document.querySelectorAll(querySelectorFor(rule.selector))];
+        } catch {
```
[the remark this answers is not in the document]
dharb: thanks I guess?

jonathanKingston (approved):

injected/src/features/element-hiding.js
```diff
@@ -363,6 +372,23 @@ function forgivingSelector(selector) {
...
+ * :is() makes a selector list forgiving, but has a perf cost since
+ * it's a pseudo-class that isn't indexed. Only wrap when the selector
+ * might be a list, in which case one malformed selector can lead to
+ * the rest of the list not applying
+ */
+function querySelectorFor(selector) {
+    let resolved = resolvedSelectors.get(selector);
+    if (resolved === undefined) {
+        resolved = selector.includes(',') ? forgivingSelector(selector) : selector;
+        resolvedSelectors.set(selector, resolved);
```
jonathanKingston: Why do we need the memo-isation here? It's just a string transform right?

dharb: We don't, not necessary.

jonathanKingston (approved):

PR #3009, opened 2026-09-01
Fix onboarding crash from a second settings action during an update

aleksey-shubin:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1210369001783739/task/1217188432658912?focus=true

## Description

A settings step could raise an uncaught error, and the browser then ended the onboarding journey as a failure.

`reducer` enters the `executing` status on `update-system-value`, and leaves it on `exec-complete` or `exec-error`. Every other action reaches `throw new Error('unhandled ' + action.kind)`. `dispatch` runs the reducer inside the click handler, so the throw escapes to `window.onerror` instead of the Preact error boundary. Windows telemetry reports these as uncaught exceptions on the system settings step.

The throw stays, so an unreachable action is still loud in development. The fix closes the click path that reached it.

`pendingId` becomes `isBusy`, so every control on the step is inert while an update runs, not only the row that started it. The `Next` button takes `disabled={isBusy}` too. `InlineAction` renders a live recall button for a *skipped* one-time row, and gated it on that row's own `pending` flag. So a skipped row stayed clickable while a different row waited for its native dialog (taskbar pin, import), and the `Next` button had no gate at all.

`pending` was already a plain "this control is inert" flag - `Button` passes it to `disabled`, `Switch` passes it to the input's `disabled`, and neither has a pending-specific style - so the wider flag disables controls that were wrongly live and changes nothing else.

## Testing Steps

The new test in `onboarding.v4.spec.js` covers this. It holds `requestImport` open, as the native import window does, then clicks the taskbar row's button. Without the fix the page reports `[uncaught] unhandled update-system-value`, which is the message behind the reported failures.

`onboarding.page.js` gains `holdRequests(names)` for that. The Windows mock answers a request only when the message carries an `Id`, so the helper removes the `Id` from the named requests. The mock records the call and sends no response, and the page keeps its pending state. It supports the Windows transport only, so the test skips on the macos project.

To check by hand on Windows:

- Open onboarding and go to the system settings step.
- Press **Skip** on the taskbar row. That row keeps a live "Pin to Taskbar" button, and the import row appears.
- Press **Import**. The native import window opens.
- While that window is open, press the taskbar row's button, or press **Next**.
- Before this change the page raised an uncaught error, and onboarding was recorded as a failure. After it, both controls are disabled until the import call returns.

Test runs: `--project windows` 104 passed, `--project macos` 122 passed. `npm run tsc`, ESLint and Prettier are clean.

## Checklist

*Please tick all that apply:*

Local verification also used a Windows debug build.

---

> [!NOTE]
> **Medium Risk**
> User-visible gating on a high-traffic onboarding step during native dialogs; behavior is intentionally stricter (whole step inert) but limited to settings-step execution windows.
> 
> **Overview**
> Prevents onboarding system-settings from accepting a second action while a native settings call is still pending, which could hit an invalid reducer transition and surface as an uncaught error.
> 
> **Settings step UI (`SettingsContent.js`)** replaces per-row `pending` (only the row that started `update-system-value`) with **`isBusy`** whenever global status is `executing`. Every visible row’s controls get the existing `pending` → `disabled` behavior, and the step’s **Next / Start Browsing** button is **`disabled={isBusy}`** as well. That blocks clicks on skipped rows that still show recall buttons (e.g. Pin to Taskbar) while another row waits on a long-lived dialog (e.g. import).
> 
> **Integration tests** add **`holdRequests`** on the onboarding page helper: on Windows it strips `Id` from selected `windowsInteropPostMessage` calls so mocks record the request but never complete it, keeping the app in `executing`. A new v4 spec holds `requestImport` open, starts import, then forces a taskbar click and asserts no exception reports and the taskbar control is disabled.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ae38f88b3e6f7ca1308d1d2faa29ac4370f75fe2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/special-pages/pages/onboarding/app/global.js
+++ b/special-pages/pages/onboarding/app/global.js
@@ -170,10 +170,7 @@ export function reducer(state, action) {
                     };
                 }
                 default:
-                    // A second action while an update is in flight has no valid transition from here.
-                    // Throwing would escape to window.onerror and fail the onboarding journey, so the
-                    // action is dropped instead and the UI keeps its controls inert until 'exec-complete'.
-                    return state;
+                    throw new Error('unhandled ' + action.kind);
             }
         }
     }
--- a/special-pages/pages/onboarding/app/global.js
+++ b/special-pages/pages/onboarding/app/global.js
@@ -170,7 +170,10 @@ export function reducer(state, action) {
                     };
                 }
                 default:
-                    throw new Error('unhandled ' + action.kind);
+                    // A second action while an update is in flight has no valid transition from here.
+                    // Throwing would escape to window.onerror and fail the onboarding journey, so the
+                    // action is dropped instead and the UI keeps its controls inert until 'exec-complete'.
+                    return state;
             }
         }
     }
--- a/special-pages/pages/onboarding/app/v4/components/SettingsContent.js
+++ b/special-pages/pages/onboarding/app/v4/components/SettingsContent.js
@@ -36,7 +36,9 @@ export function SettingsContent({ advance, dismiss, updateSystemValue }) {
     const { step, status, order, activeStep } = globalState;
     const isDone = globalState.activeRow >= step.rows.length;
     const isLastStep = order[order.length - 1] === activeStep;
-    const pendingId = status.kind === 'executing' && status.action.kind === 'update-system-value' ? status.action.id : null;
+    // While an update is in flight every control on the step stays inert, not just the row that
+    // started it: a second action has no valid transition out of the reducer's 'executing' state.
+    const isBusy = status.kind === 'executing';
 
     const [exitingIndex, setExitingIndex] = useState(/** @type {number | null} */ (null));
     const enteringIndex = exitingIndex !== null && exitingIndex + 1 < step.rows.length ? exitingIndex + 1 : null;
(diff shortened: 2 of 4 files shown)


daxtheduck (dismissed):

aleksey-shubin:
@noisysocks could you take a look? I think the disable button change is safe, but I'm unsure if removing the throw altogether is a good idea.

noisysocks (changes requested):
The `SettingsContent` fix makes sense to me. Removing the `throw` sits a little unwell with me, though maybe I'm being too pedantic / too much of a purist. Don't we _want_ to fail noisily here so that we catch exactly bugs like the one you've fixed just now in `SettingsContent`?

I could be persuaded otherwise. 

At any rate, please add and red/green a regression int test 🙏

aleksey-shubin:
@noisysocks thanks! I had the same doubts about removing the `throw`, so I reverted the change.
Added a regression test.

noisysocks (approved):
🚀

PR #3010, opened 2026-09-01
Capture globals

GuiltyDolphin:
**Asana Task/Github Issue:** https://app.asana.com/1/137249556945/project/1200811069326255/task/1217408393396344?focus=true
## Description

Capture more global values used by web detection inc. query selectors.

## Testing Steps

- 

## Checklist

*Please tick all that apply:*

---

> [!NOTE]
> **Medium Risk**
> Large refactor of security-sensitive matcher logic on hostile pages; behavior should be equivalent but any missed uncaptured call would leak detector configuration.
> 
> **Overview**
> **Hardens web-detection matching** so remote-config selectors, XPath expressions, and regex patterns are not observable when a page replaces `querySelectorAll`, `RegExp.prototype.test`, `Array.prototype` methods, or similar slots after injection.
> 
> `captured-globals.js` now snapshots DOM and intrinsic APIs at module load (`document_start`) and exposes receiver-style helpers (`documentQuerySelectorAll`, `regExpTest`, `nodeTextContent`, etc.). `matching.js` and `parse.js` route all config-driven DOM, regex, and collection work through those captures, using **index loops** instead of `for...of`, spread, or `.some`/`join` on config arrays.
> 
> **Enforcement and docs:** ESLint blocks raw globals, risky prototype calls, and replaceable iteration in `injected/src/features/web-detection/**`. Coding guidelines and `BUGBOT.md` describe the disclosure model.
> 
> **Tests:** Shared JSDOM setup (`install-dom-globals.js`) so captures bind before specs load; unit tests that replace prototypes and assert secrets stay hidden; an integration test that proves live reCAPTCHA detection still fires while only a control selector is observed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 910a1f16dbb17396a62ae6d3d9ee549d200054ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/injected/src/features/web-detection/matching.js
+++ b/injected/src/features/web-detection/matching.js
@@ -136,10 +136,8 @@ function hasContent(element) {
         return true;
     }
     // A real (eg cross-origin Turnstile) iframe counts; about:blank does not.
-    return [...elementQuerySelectorAll(parsed, 'iframe')].some((frame) => {
-        const frameElement = /** @type {HTMLIFrameElement} */ (frame);
-        return !frameElement.hidden && frameElement.src !== '' && frameElement.src !== 'about:blank';
-    });
+    const frames = /** @type {NodeListOf<HTMLIFrameElement>} */ (elementQuerySelectorAll(parsed, 'iframe'));
+    return [...frames].some((frame) => !frame.hidden && frame.src !== '' && frame.src !== 'about:blank');
 }
 
 /**
--- a/injected/src/features/web-detection/matching.js
+++ b/injected/src/features/web-detection/matching.js
@@ -1,4 +1,3 @@
- 
 import {
     createXPathExpression,
     documentQuerySelector,
@@ -185,7 +184,7 @@ function compileXPath(expression) {
     }
     let compiled = cache.get(expression);
     if (!compiled) {
-        compiled = createXPathExpression(document, expression);
+        compiled = createXPathExpression(expression, null);
         cache.set(expression, compiled);
     }
     return compiled;
@@ -357,7 +356,7 @@ function evaluateSingleTextCondition(condition) {
     // Checked before xpath because CSS matching avoids the per-call expression
     // parse and snapshot allocation that `document.evaluate` requires.
     const selectorMatch = selectors.some((selector) => {
-        const elements = documentQuerySelectorAll(document, selector);
+        const elements = documentQuerySelectorAll(selector);
         for (const element of elements) {
             if (patternComb.test(element.textContent || '')) {
                 return true;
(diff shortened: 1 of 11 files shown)


injected/src/features/web-detection/matching.js
```diff
@@ -1,5 +1,47 @@
...
+    RegExp,
+    rectHeight,
+    rectWidth,
+    regExpTest,
+    snapshotItem,
+    snapshotLength,
+    stringCharCodeAt,
+    stringSlice,
+    stringTrim,
+} from '../../captured-globals.js';
```
jonathanKingston, line 43: ⚰️

injected/src/features/web-detection/matching.js
```diff
@@ -18,7 +60,52 @@ import { hasOwnProperty, objectKeys } from '../../captured-globals.js';
...
+}
+
+/**
+ * Concatenate `list` with `separator` between entries.
+ *
+ * @param {string[]} list
+ * @param {string} separator
+ * @returns {string}
+ */
+function join(list, separator) {
```
jonathanKingston, line 103: I sort of don't think we should be rewriting these. Can you explain why it's added those?

injected/src/features/web-detection/matching.js
```diff
@@ -18,7 +60,52 @@ import { hasOwnProperty, objectKeys } from '../../captured-globals.js';
  */
 function asArray(value, defaultValue = []) {
```
jonathanKingston, line 61: As mentioned, maybe some of these belong as array utils as its own distinct file. That would remove some of the import chaos.

I still want to question whether we need to rewrite all of the utils rather than just import them.

injected/src/features/web-detection/matching.js
```diff
@@ -381,13 +475,20 @@ function evaluateSingleTextCondition(condition) {
  */
 function evaluateSingleElementCondition(config) {
     const visibility = config.visibility ?? 'any';
+    const selectors = asArray(config.selector);
     // Disjunction: any selector having a matching element is success
-    return asArray(config.selector).some((selector) => {
+    for (let i = 0; i < selectors.length; i++) {
```
jonathanKingston, line 480: Is this because technically iterators can be overridden and they aren't protectable?

jonathanKingston: Overall, my view is that indices for loops are more prone to error, and so if we can avoid them, that would be better.

injected/src/features/web-detection/matching.js
GuiltyDolphin: to avoid a captured `this` in `Array.prototype.join`, but we can do this by capturing via `ReflectApply` (same pattern as `regExpTest`) so I've switched it to use that as it avoids re-implementing.

PR #3021, opened 2026-09-07
Fix iframe sandbox detection to require both allow-scripts and allow-same-origin

jonathanKingston:
**Description**

Improves iframe content extraction in `domToMarkdown` by fixing the sandbox attribute validation logic. Previously, the code only checked for the `allow-scripts` flag when deciding whether to access an iframe's document. This change now requires both `allow-scripts` and `allow-same-origin` flags to be present.

Without `allow-same-origin`, a sandboxed iframe has an opaque origin and its document is inaccessible regardless of whether the src matches the parent's origin. The fix prevents unnecessary access attempts that would trigger sandbox violation errors in WebKit and other browsers.

The implementation also:
- Extracts sandbox flag parsing into a dedicated `getSandboxFlags()` helper function
- Normalizes flags to lowercase for case-insensitive comparison (per HTML spec)
- Reads from the `sandbox` attribute directly rather than the `iframe.sandbox` DOMTokenList (for JSDOM compatibility in tests)
- Updates comments to document both error conditions

**Testing Steps**

- Run `npm run test-unit` to verify all iframe sandbox handling tests pass
- The new test suite covers:
  - Unsandboxed iframes (content included)
  - Empty sandbox attribute (content skipped)
  - Sandbox with only `allow-scripts` (content skipped)
  - Sandbox with only `allow-same-origin` (content skipped)
  - Sandbox with both flags (content included)
  - Case-insensitive flag matching

**Checklist**

https://claude.ai/code/session_013RvaTxYBzRagtcMDrJeef9

---

> [!NOTE]
> **Low Risk**
> Localized change to iframe pre-checks before document access; reduces console noise and skips inaccessible sandboxed frames without touching auth or network paths.
> 
> **Overview**
> **Tightens when `domToMarkdown` reads same-origin iframe bodies** so sandboxed frames are only accessed when both `allow-scripts` and `allow-same-origin` are set. Previously only `allow-scripts` was required, which could still hit opaque-origin sandbox frames and trigger WebKit sandbox console errors.
> 
> Adds **`getSandboxFlags()`** to parse the `sandbox` attribute (lowercased, token-split) instead of `iframe.sandbox`, so checks work under JSDOM and treat flags case-insensitively.
> 
> **Unit tests** cover unsandboxed inclusion, empty/restrictive sandbox skipping, each flag alone, both flags together, and mixed-case sandbox tokens.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d57f37078a8dd1a20c0dfca3570cff076336e8e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/injected/src/features/page-context.js
+++ b/injected/src/features/page-context.js
@@ -41,6 +41,19 @@ function isHtmlElement(node) {
 }
 
 /**
+ * Parse the sandbox attribute of an iframe into its set of flags.
+ * Read from the attribute rather than `iframe.sandbox` so the check works in
+ * environments without a DOMTokenList for it (e.g. JSDOM in unit tests).
+ * Sandbox flags are ASCII case-insensitive, so normalise to lower case.
+ * @param {HTMLIFrameElement} iframe
+ * @returns {Set<string>}
+ */
+function getSandboxFlags(iframe) {
+    const value = iframe.getAttribute('sandbox') ?? '';
+    return new Set(value.toLowerCase().split(/\s+/).filter(Boolean));
+}
+
+/**
  * Check if an iframe is same-origin and return its content document
  * @param {HTMLIFrameElement} iframe
  * @returns {Document | null}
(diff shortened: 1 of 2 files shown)


daxtheduck (dismissed):

daxtheduck (approved):

jonathanKingston:
@shakyShane I'm assigning this to you just for a quick review. I noticed this in a change that we're making, and we're throwing a CSP error, which had no real detriment, but was a little bit annoying and will be for site owners. Let me know if there's any concerns or if you can assign it to someone else.

